PR #6, opened 2017-06-25
Make build run

forki:
I tried to make it run on my machine. Needed to go a bit down with the target framework. Also made it use paket magic mode. But now I'm lost `src\elmish\*` seems to be missing.

image


forki:
tried to link the files via github links. still no touchdown

forki:
yay! now it builds!

2sComplement:
Paket and framework updates look great.

Elmish sources were intended to be pulled in as a submodule. Did you try `git submodule --init --recursive`?

forki:
It's your project but could you please try without submodules?!
Using github deps might be a good alternative here

forki:
(it's all automated in this way)

2sComplement:
Can you pin to a specific commit?

forki:
look at the lockfile. it's pinned to a commit

2sComplement:
Ah cool, I only looked at the dependencies file earlier. Sure we can give this a try.

PR #7, opened 2017-06-25
logging should be done by elmish logger

forki:
we should not advertise that in readme


2sComplement:
By elmish logger do you mean `Program.withConsoleTrace`?

forki:
yes

Issue #8, opened 2017-06-25
Support for Framework < 4.6.2


isaacabraham:
Is there a possibility to support 4.6.1 or lower? This would open up stock VS2015 users etc..

isaacabraham:
Ignore this - closing.

forki:
I already adressed that in my PR

2017-06-25 15:03 GMT+02:00 Isaac Abraham <notifications@github.com>:

isaacabraham:
You can get around this by installing 4.7 on a VS2015 machine but yes, I wasn't sure if the bar really needed to be so high as a minimum.

forki:
it's already 4.5 buddy.

isaacabraham:
Yes, I know. When I raised the issue, the nuget package wasn't :-)

BentTranberg:
@forki, when I create an F# 4.0 (FSharp.Core, 4.4.0.0) project on .NET Framework 4.5.2 in VS 2015, and then try to install Elmish.WPF v1.0.0-beta-2, I get the following error message.

Could not install package Elmish.WPF 1.0.0-beta-2. You are trying to install this package into a project that targets .NETFramework,Version=v4.5.2, but the package does not contain any assembly references or content files that are compatible with that framework. For more information, contact the package author.

I worked around this by upgrading the project to Framework 4.6.1, installing the package, then downgrade to 4.5.2 again. Not sure why that works, but the project runs fine.

forki:
I only changed it here in the code. I'm not a maintainer. So I don't know if it went to nuget

2sComplement:
I hadn't updated the target framework in paket.template. Should be fixed in beta-3.

PR #32, opened 2018-02-06
conversion to elmish 2.0

et1975:
using netstandard2.0 Elmish package dependency.


et1975:
The good - everything builds at this point, the bad - WPF + SDK still don't play nice together and something is still borked - all the samples crash at runtime. I only tried it on small headless VM, so maybe someone else could track down the reason.

2sComplement:
Do the samples still work in Windows?
Also, does this invalidate #8?

et1975:
> Do the samples still work in Windows?
see the above "samples crash" ;)

> does this invalidate #8?
Good point, I think we can still build for 4.5 if needed, but it's not in the current version of the Elmish package (461 is though).

et1975:
Come to think, the whole NetStandard was introduced to address multi-targeting mess, so probably should just use `netstandard1.1` to target all the [compatible platforms](/dotnet/standard/blob/master/docs/versions.md) (including net45) at once.

et1975:
No longer needed

cmeeren:
Thanks! :)

cmeeren:
@et1975 When converting the XAML projects to the new project format, did you manage to get them to compile from the command line? When I do `dotnet build` in a XAML project directory and have

```xml
<LanguageTargets>$(MSBuildExtensionsPath)\$(VisualStudioVersion)\Bin\Microsoft.CSharp.targets</LanguageTargets>
```
In the project file, I get the following error:

> C:\Program Files\dotnet\sdk\2.1.400\Sdks\Microsoft.NET.Sdk\Sdk\Sdk.targets(41,3): error MSB4019: The imported project "C:\Program Files\dotnet\sdk\2.1.400\15.0\Bin\Microsoft.CSharp.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk. [D:\Code\GH\Elmish.WPF\src\Samples\OneWaySeq.Views\OneWaySeq.Views.csproj]

I guess this is because `$(VisualStudioVersion)` doesn't make sense outside VS. If I instead use your version from this PR,

```xml
<LanguageTargets Condition="Exists('$(MSBuildProjectDirectory)\$(AssemblyName).csproj') AND Exists('$(MSBuildExtensionsPath)\$(VisualStudioVersion)\Bin\Microsoft.CSharp.targets')">$(MSBuildExtensionsPath)\$(VisualStudioVersion)\Bin\Microsoft.CSharp.targets</LanguageTargets>
```
then I instead get another error:

> MainWindow.xaml.cs(9,13): error CS0103: The name 'InitializeComponent' does not exist in the current context [D:\Code\GH\Elmish.WPF\src\Samples\OneWaySeq.Views\OneWaySeq.Views.csproj]

Which I suppose is because the targets aren't included at all (since the condition fails).

Did you manage to get the XAML projects to compile from the command line / build script?

et1975:
Yes, it compiled on my windows VM, but this looks familiar, I think you need something like [this cludge](https://github.com/AppCoreNet/DependencyInjection/blob/dev/build/netfx.props) to make sure dotnet finds the framework targets and XAML codegen kicks in.

cmeeren:
I've had a look at that issue, but didn't understand from it how to make it compile. Also I don't know what do to with the file you linked to. I tried saving it in the project folder of one of the view projects, but it still fails with the same errors.

For now, I have excluded the samples from the build target.

PR #38, opened 2018-07-18
[Ready for reviews] Binding.complexModel (ex: Binding.collectionTwoWay & Binding.collectionModel)

JDiLenarda:
There, I think it's ready for review.

`Binding.collectionTwoWay` has been dropped. It was of too narrow use-case, dealing only with list of simple values. _(Don't trust the title of last commit, it is collectionTwoWay that is dropped, not collectionModel)_ There may be some place for such function, but let's see what can be done with the rest before.

`Binding.collectionModel` is renamed `Binding.complexModel`, because it also allows to work with recursive models and views.

`complexModel getter: ('model -> IEnumerable<'_model>) -> viewBindings: (unit -> ViewBindings<'_model,'_msg>) -> toMsg: (('_model * '_msg) -> 'msg) -> name:string -> ViewBinding<'model,'msg>`

`getter` retrieves a list of sub-models from a model
`viewBindings` must be a function returning a `ViewBindings<'_model,'_msg>`. It must be a function rather than a value so we can use recursive ViewBindings.
`toMsg` takes the current sub-model and the sub-message as parameters.

The list `getter` returns can be any type implementing `IEnumerable<'T>`, but the API will expose an `ObservableCollection<ViewModelBase<'_model,'_msg>>` to the control. If `getter` returns a non-`ObservableCollection`, it will be wrapped in a new one _at every update_. If it returns an ObservableCollection, it will be used as such (more or less).

Since the sub-model bindings must be kept from one call to update from the other, they are stored in 2 dictionaries to which keys are the binding names : `submodels` store the raw models as `obj`, `submodelbindings` store the `ModelViewBase`. This is the latter that is exposed to the control source property.

Though I submit this work to review, I consider exposing a ReadOnlyObservableCollection instead of ObservableCollection, but I want some opinions and feedback before.

About test projects :
- The Gridview test project shows it on a long list of sub-models, through an editable DataGrid and a read-only one, mirroring every change on the list. The editable DataGrid has a context menu for deleting and inserting rows.
- The TreeView test project shows it on a nested model. Like the other test project, it has two controls showing the same data, one editable, the other as read only for mirroring the changes.

Toy with them, people !

--------------------
Original message :

- Binding.collectionTwoWay  allows binding  to a collection in two way mode. Fit better with single value collection than object collection. See Elmish.Test.CollectionBinding        
- Binding.collectionModel allows binding on collection of model, included nested models. Elmish.Test.CollectionModel.GridView implements it in a gridview with lots of data. Slow to update. Elmish.Test.CollectionModel.TreeView implements a self-nested two way model. The model is a discriminated union. Here again, update has issues : it works, but since the whole model is updated, the tree view collapse.


2sComplement:
Thanks for submitting this. Is this ready for review, or is it still a work in progress?

JDiLenarda:
It is a work in progress. Later today, I will commit a version with a somewhat satisfying Binding.collectionModel with full descriptions.

2sComplement:
Sounds good. Whenever you're happy with it, please remove the [WIP] label from the title.

JDiLenarda:
Here it is.

### Binding.collectionTwoWay
`let collectionTwoWay (listGetter:'model -> seq<'a>) (keyedGetter:'model -> IndexOrKey -> 'a) (keyedSetter:'a -> 'model -> IndexOrKey -> 'msg) name : ViewBinding<'model,'msg>`
Allows binding to the elements of a collection in two-way mode.
See test project to see implementations.
How it works now : when binding to a collection, a `ListViewModel` is created. It inherits `DynamicObject` to have `TryGetIndex` and `TrySetIndex` implementations, and it implements `IEnumerable` to have a `GetEnumerator` implementation.
- The `GetEnumerator` is called when a list is bound to some, say, `ListView.ItemSource`. For each element in the tied list, it creates a `ItemViewModel`, inheriting `DynamicObject` and implementing `TryGetMember` (calling the indexed getter from the API call) and `TrySetMember` (calling the indexed setter).
- The `TryGetIndex` and `TrySetIndex` are called when there is an indexed binding in the xaml such as `Text="{Binding MySource[2]}"`. The value can be an int or a string. They too use the indexed getter and setter from the call.

What occurs to me later, while working on collectionModel, is that it prevents to bind some more complex collections, notably ObservableCollection. Maybe it should be rewritten to work the same way.

### Binding.collectionModel
`let collectionModel (getter:'model -> IEnumerable<'_model>) (viewBindings: unit -> ViewBindings<'_model,'_msg>)  (toMsg: int -> '_msg -> 'msg) name : ViewBinding<'model,'msg>`
Allows bindings on complex models, such as list of sub-models or nested models.
- The Gridview test project shows it on a long list of sub-models, through an editable DataGrid and a read-only one, mirroring every change on the list. The editable DataGrid has a context menu for deleting and inserting rows. I lately noticed a bug : if you delete a row and run through the end of the grid, it'll raise an exception. More on this later.
- The TreeView test project shows it on a nested model. Like the other test project, it has two controls showing the same data, one editable, the other as read only for mirroring the changes.

How it works : there, at first I tried to do something similar to what I did with collectionTwoWay, but I soon realized it was not usable for working with load of data, neither for binding with some complex control like DataGrid. In fact, when I started to tinker with Elmish.WPF code, my goal was to work efficiently with F# collection classes, but really nothing beat an ObservableCollection. So long for purity. Anyway, It still allows to work with any kind of IEnumerable inheriting class.
- If the model is bound to a non-ObservableCollection list, one will be created on the fly at binding. At every update, it will be cleared and refill.
- If the model is bound to an ObservableCollection, then one other will be created on the fly at binding, but dynamically mapped (I used [this snippet](http://www.fssnip.net/dv/title/Map-for-ObservableCollection) for that). Then there is no need to clear and refill at every update and it is as fast as you can hope.

Unfortunately, I then spotted the bug from the GridView project, and I just hope it doesn't show something wrong with this whole design... I'll work to fix it, then I will ask for your feedback and suggestions.

PR #42, opened 2018-08-03
Use existing application instance if it exists

### LINKED ISSUE #41: Application must be instantiated before call to window constructor due to resources in App.xaml
When using `Program.runWindow`, the window must necessarily be instantiated first. Since `Application` isn't instantiated before calling `runWindow`, I get the following exception on the `InitializeComponent()` call in the window constructor:

`Exception: Cannot find resource named 'MaterialDesignMultiFloatingActionAccentPopupBox'. Resource names are case sensitive.`

This error occurs because the resource in question is defined in my `App.xaml`:

```xaml
<Application.Resources>
  <ResourceDictionary>
    <ResourceDictionary.MergedDictionaries>
      <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Light.xaml" />
      <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Defaults.xaml" />
      <ResourceDictionary Source="pack://application:,,,/MaterialDesignColors;component/Themes/Recommended/Primary/MaterialDesignColor.Blue.xaml" />
      <ResourceDictionary Source="pack://application:,,,/MaterialDesignColors;component/Themes/Recommended/Accent/MaterialDesignColor.Lime.xaml" />
      <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.PopupBox.xaml" />
    </ResourceDictionary.MergedDictionaries>
  </ResourceDictionary>
</Application.Resources>
```

Since this is not instantiated/loaded before the window is initialized, I am unable to initialize my main window to start the app.

What is the best way to solve this? Is there something I can do, or must this be fixed in Elmish.WPF?

cmeeren:
Fixes #41 

Turned out to be a devilishly simple fix. Simply use the existing application instance if it exists. AFAIK this is non-breaking, since there can only be one `Application` instance per AppDomain, so existing users have not been able to instantiate their own `Application` instance until now, ensuring that this is an opt-in feature.

With this simple change, users that need application-wide resources or otherwise need to specify an application instance can simply instantiate it before calling the `Program` functions. For example:

```f#
[<EntryPoint;STAThread>]
let main _argv =
  App() |> ignore
  Program.mkSimple init update view
  |> Program.runWindow (MainWindow())
```

Something like this should probably be added to the documentation and/or a sample.

### DIFF
--- a/src/Elmish.WPF/Program.fs
+++ b/src/Elmish.WPF/Program.fs
@@ -27,7 +27,7 @@ let private _run debug (window:Window) (programRun:Program<'t, 'model, 'msg, Vie
     |> programRun
     
     // Start WPF dispatch loop
-    let app = Application()
+    let app = if isNull Application.Current then Application() else Application.Current
     app.Run(window) //blocking
 
 /// Blocking function.


cmeeren:
Closed by #48

PR #44, opened 2018-08-04
memoize mapper in oneWayMap

### LINKED ISSUE #43: Performance suggestion: Make the mapper in oneWayMap bindings lazy or memoized
As far as I can see, `oneWayMap` represents an optimization over `oneWay` in the sense that `PropertyChanged` will not be triggered unless the return value of `getter` has changed. However, if there are multiple bindings from XAML to a single `oneWayMap` (e.g. one to display text and another to select a color based on the text using an IValueConverter), then the `mapper` will still be called several times.

It seems to me there should be a fairly simple fix wherein Elmish.WPF could use `lazy` to ensure that until the `getter` output has changed (or, if simpler, until the model is updated, though that's a lesser optimization), `mapper` is called at most once.

I haven't been able to figure out how, though. Any thoughts?

cmeeren:
Fixes #43

Briefly tested, seems to work fine. `mapper` now remembers the return value of its previous argument. (I chose to limit it to a single value since that fixes #43 and avoids memory usage concerns from remembering an unlimited number of return values.)

### DIFF
--- a/src/Elmish.WPF/Elmish.WPF.fsproj
+++ b/src/Elmish.WPF/Elmish.WPF.fsproj
@@ -57,6 +57,7 @@
       <Paket>True</Paket>
       <Link>elmish/src/program.fs</Link>
     </Compile>
+    <Compile Include="Utils.fs" />
     <Compile Include="Binding.fs" />
     <Compile Include="ViewModel.fs" />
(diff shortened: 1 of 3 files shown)


cmeeren:
Closed by #48

Issue #46, opened 2018-08-14
Merging of cleanup/re-implementation of Elmish.WPF


cmeeren:
In my [quest for recursive sub-model lists](https://github.com/Prolucid/Elmish.WPF/issues/5#issuecomment-411547035), I ended up rewriting most of Elmish.WPF from scratch, cleaning up some cruft and making names and types clearer (IMHO). I find the new implementation easier to grasp (says the author…), and I think you'll find there's quite a bit of attention to detail. In any case, I’d like to explain why and would like an open discussion on how we should proceed (use all/most of the changes vs. rewriting the features that are actually new into the existing implementation).

[The new implementation is here.](https://github.com/cmeeren/Elmish.WPF.Recursive) I haven’t made a PR because the diff would be pointless, but let me know if you want one anyway.

My implementation solves/supersedes #5, #38, #41, #42, #43, #44, and #45.

*Disclaimer: The views expressed here are subjective. It is not my intention to step on any toes. But I don’t want to waste anyone’s time sugarcoating or beating around the bush, so I’ll call ’em like I see ‘em.*

### Why a re-implementation?

Simply put, I had trouble understanding the existing implementation. Not because there’s anything fundamentally wrong with it – it’s great, and mine is fundamentally similar. But all the type aliases confounded me, and it’s clear that the `ViewModelBase`, `ViewBinding`, `ViewBindings`, `Variable`, and `PropertyAccessor` types weren’t made with sub-model usage in mind, or lazy evaluation (which I also wanted).

So I did the only thing that made sense to me in order to understand the architecture: Using the existing implementation as a guide, I rewrote Elmish.WPF mostly from scratch, only complexifying as needed and in the direction that made most sense to me considering the final features I had in mind.

My rewrite turned out very similar, but types and functions relating to bindings and the view model is refactored quite a bit compared to the existing implementation.

### Significant internal changes

* The binding union types have been rewritten from scratch and IMHO have a clearer function. `BindingSpecData` is generated by the `Binding` functions and contains everything the view model needs to set up a binding. `Binding` contains all data used by active bindings in the view model (including mutable types / ref cells where relevant).
* `ViewModelBase` (now called `ViewModel`) has been rewritten from scratch (but works in fundamentally the same way)
* More samples, and all samples are consistently designed and implemented.
* All projects use the new fsproj/csproj formats.
* All non-public stuff has been moved to an `.Internal` namespace.

### Public breaking changes

[...]

Issue #45, opened 2018-08-07
How to use invalid input in update function?


cmeeren:
Consider a form with a single input field. It must be non-empty to be valid. If it's valid, the Submit button is enabled; if it's invalid, the button is disabled.

This simple (and I'd guess common) use case seems impossible in Elmish.WPF.

The reason: AFAIK, in Elmish.WPF, `Binding.twoWayValidation` is the only way to trigger validation messages. `twoWayValidation` only sends a message when the input is valid. This means that changes to an input that moves the input from valid to invalid are not sent, and thus validation state (or invalid input) cannot be a part of the model.

Unless I'm mistaken, this means that it's impossible to e.g. disable a button whenever the input is invalid, because if the user moves from a valid state (text in the input field) to an invalid state (empty input field), no message is sent and the state keeps the (now outdated) previous valid value, happily ignorant that the input field is now in fact invalid.

In short, there needs to be another way to trigger validation in Elmish.WPF that still allows messages (which can then contain the raw, invalid input) to be sent to Elmish even for invalid input.

et1975:
I must be missing something, because one can validate on every update, the invalid value then becomes part of the state along with the validation results, which you'd bind to some other element.

cmeeren:
I can keep the raw (potentially invalid) state in the model, but in that case I can't use `Binding.twoWayValidation` (because I won't get invalid values back into the model), and I don't know of any other (Elmish.WPF-friendly) ways to trigger validation error messages for input controls. Do you have any? (That preferably don't require a boatload of boilerplate XAML?)

et1975:
I see, you want WPF-specific validation to kick in. I personally haven't touched WPF in a while, maybe @2sComplement can comment when he's back from vacation. But...

Here's my (somewhat dated) take on the validation: https://thoughtspam.wordpress.com/2013/09/02/starting-an-open-source-rewrite-of-a-validation-microframework/

The gist of it is, describing your validation rules in terms of your domain and independently from presentation technology has many advantages and that's the suggested "elmish" way. We haven't implemented anything specific to support this, but others have, here's an example: https://github.com/zaaack/fable-validation

et1975:
If you already have validation results based on the model, what's stopping you from storing that in the state and binding them to the error controls? I'm assuming some glue might be required between the two, but it sounds like you're in good position to figure it out.

cmeeren:
[...]

PR #48, opened 2018-08-25
Rewrite

### LINKED ISSUE #5: Binding Collections
If my model contains a list and I have Add and Remove messages to add, how would I bind the list to a grid or listview in the view? Thanks.

cmeeren:
For a full description, see #46.

Closes #5, #29, #38, #41, #42, #43, #44, #45, #46.

TODO:

* [x] Add unit test project stub
* [x] Set up build script
* [x] Set up paket template file or similar
* [x] Set up AppVeyor
* [x] Edit readme
* [x] Get a few approvals on breaking changes
* [x] Edit release notes


paket.dependencies
```diff
@@ -1,4 +1,6 @@
+framework: auto-detect
+
 source https://api.nuget.org/v3/index.json
 
 nuget Elmish 2.0.0-beta-4 prerelease
-nuget FSharp.Core 4.5.2
+nuget FSharp.Core >= 4.5.2
```
inosik: I'd prefer to depend on the lowest possible FSharp.Core. Also see elmish/elmish#146.

cmeeren: In any case, Elmish core requires >= 4.5.2 so it's kinda a moot point. But I lowered it in case Elmish lowers it. No need to require more than we need, even if our dependencies do.

cmeeren:
### Request for review on public breaking changes

This is almost ready to merge now, I'd just like a few approvals on breaking changes (@rchybicki have already approved, https://github.com/elmish/Elmish.WPF/issues/46#issuecomment-413125814).

For a more complete discussion, see #46.

Pinging people who might want to weigh in (completel optional): @2sComplement, @et1975, @JDiLenarda

If I don't hear any protests in a handful of days, I'll probably go ahead and merge this into master and release as 2.0.0-beta-1 on Nuget. Some potentially breaking changes remains before a final 2.0.0 release (e.g. a different comparison method when deciding whether to call NotifyPropertyChanged - currently we might potentially be structurally comparing large lists of records).

The following breaking changes have been made:

* `twoWayValidation` is called `twoWayIfValid` (because that’s what it is, and it clearly separates it from the new `twoWayValidate`)
* `oneWayMap` is called `oneWayLazy` (its implementation has changed, and the use case has expanded, but is similar)
* `cmd` and `cmdIf` have been renamed `paramCmd` and `paramCmdIf`, because the old names have new signatures/use-cases
* `model` has been renamed `subModel` because it’s more clear, and consistent with the new `subModelOpt` and `subModelSeq`
* `Program.runDebugWindow` has been removed in favour of `Program.runWithConfig`

paket.dependencies
```diff
@@ -1,16 +1,29 @@
...
+  source https://api.nuget.org/v3/index.json
 
-nuget FsXaml.Wpf
\ No newline at end of file
+  nuget Hedgehog
+  nuget Hedgehog.Experimental
+  nuget Microsoft.NET.Test.Sdk
+  nuget Unquote
+  nuget xunit.core
+  nuget xunit.runner.visualstudio version_in_path: true
```
inosik: Does this work? At least in VS 2013 test adapters had to be located in `packages/` directly.

cmeeren: Works great in VS2017.

JDiLenarda:
Since I haven't use Elmish.WPF in production case yet, I personally don't mind breaking change and I'm fine with your naming.

Issue #52, opened 2018-09-13
Propositions for a next version - multi-agent application and dialog/window utilities


JDiLenarda:
As I said in #24, I wrote a version [you can see here](https://github.com/JDiLenarda/Elmish.WPF.RunControl). 

This development came in 2 steps :

- first, trying to write an AutoCAD palette with Elmish.WPF
- then thoughts on dealing with multiple windows (modal and modeless).

For the first step, I wrote a function that can bind a `UserControl` to an Elmish program and modified `ViewModel` so its constructor take a WPF thread dispatcher as a parameter (the AutoCAD thread dispatcher can't be retrieved with `Application.Current.Dispatcher`).

The second step came when #24 was reopen. It made me theorizing about the way to deal with multi-windowed application and realizing that what I did for user control can be used as well with in-app windows. Most of this work was trying to set up best practices to do so, through writing some sample projects, and also led to include some utilities function for dealing with Elmish loop as `DataContext`.

See sample projects ModelessWindows.SharedLoop and ModelessWindows.SeparatedLoop. Get a look at the code, this is where I try to find the best way to work with multi window applications. Note that it may work as well with controls. See also OpenDialog, though this one is more raw-coded. Notice that all window management is made in the bindings/views function, not in the update loop.

To boil it down, what does this version allow ?

- writing "multi-agent" app, where windows and controls can run their own Elmish loop. Useful for keeping models light and focused on an application heavy on data navigation (be it through windows or pages), and working asynchronously on some data (like a submit palette control/window).
- writing controls or windows libraries.
- writing plug-ins for extensible applications.

There are a few thing unfinished where I may need help, notably a function that would be able to apply a XAML binding path-like string to a `ViewModel` (for now, I use an abstract construct to do so).

Also, check for the ModelessWindows.* projects and tell me if the practices I try to set are good or not.

And let me know if it is worthy of a pull-request.

JDiLenarda:
First, I'll admit that I never considered other Elmish works. I'm a recent F# enthusiast and was seeking for a way to make WPF GUI with it and Elmish.WPF look like the coolest thing to do so. I of course took a look at the Elm Architecture page, and I got that Elmish.WPF is the ugly duckling of the family, but I'd be interested to see other resource, notably user's experience.

[...]

PR #54, opened 2018-09-17
Refactoring of Progam.fs

JDiLenarda:
new ViewModelUtilities module
Program module refactored for using new module
threadDispatcher added to ViewModel constructor

### DIFF
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -7,7 +7,6 @@ open System.Collections.ObjectModel
 open System.ComponentModel
 open System.Windows
 open Elmish.WPF
-open System.Windows.Threading
 
 /// Represents all necessary data used in an active binding.
 type Binding<'model, 'msg> =
@@ -51,7 +50,7 @@ and [<AllowNullLiteral>] ViewModel<'model, 'msg>
         dispatch: 'msg -> unit,
         bindingSpecs: BindingSpec<'model, 'msg> list,
         config: ElmConfig,
-        threadDispatcher: Dispatcher)
+        uiDispatch: (unit -> unit) -> unit )
       as this =
   inherit DynamicObject()
 
@@ -75,7 +74,7 @@ and [<AllowNullLiteral>] ViewModel<'model, 'msg>
     propertyChanged.Trigger(this, PropertyChangedEventArgs propName)
 
   let raiseCanExecuteChanged (cmd: Command) =
-    threadDispatcher.Invoke(cmd.RaiseCanExecuteChanged)
+    uiDispatch cmd.RaiseCanExecuteChanged
 
   let setError error propName =
     match errors.TryGetValue propName with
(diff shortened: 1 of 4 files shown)


JDiLenarda:
Pull request on demand of @cmeeren  on #52

src/Elmish.WPF/ViewModel.fs
```diff
@@ -49,7 +50,8 @@ and [<AllowNullLiteral>] ViewModel<'model, 'msg>
       ( initialModel: 'model,
         dispatch: 'msg -> unit,
         bindingSpecs: BindingSpec<'model, 'msg> list,
-        config: ElmConfig )
+        config: ElmConfig,
+        threadDispatcher: Dispatcher)
```
cmeeren: Why is a `Dispatcher` injected here? Is it to make `ViewModel` testable? If so, since we only seem to need the `Invoke` method fro the dispatcher, perhaps it could be made even simpler and more testable by changing the type from `Dispatcher` to `unit -> unit`. What do you think? (I have no idea how simple `Dispatcher` is to create/mock, but I like to keep things as simple as reasonably possible.)

Also, do you think it would be clearer if it was given a name that indicates what it dispatches to? (Such as `uiDispatcher`, `mainDispatcher`, or any better suggestion you might have.)

src/Elmish.WPF/ViewModel.fs
```diff
@@ -7,6 +7,7 @@ open System.Collections.ObjectModel
 open System.ComponentModel
 open System.Windows
 open Elmish.WPF
+open System.Windows.Threading
```
cmeeren: Could you move this up below `System.Windows`?

src/Elmish.WPF/ViewModelUtilities.fs
```diff
@@ -0,0 +1,28 @@
+﻿[<RequireQualifiedAccess>]
+module Elmish.WPF.Utilities.ViewModel
```
cmeeren: Is there any reason this (and the additions in the other coming PR from #52) is placed under `Elmish.WPF.Utilities` instead of `Elmish.WPF`? I'm not saying I disagree, I'm just wondering about the rationale behind it.

src/Elmish.WPF/ViewModel.fs
JDiLenarda: This one is if we ever open the library to writing plug-ins for extensible application. I found out while trying to write an AutoCAD palette that `Application.Current` return `null` in such a context. Passing the thread dispatcher from the control to bind solve the problem, and hopefully will in other contexts.

Anyway you're right about passing the function instead the `Dispatcher` object, i'll make this change and call it `uiDispatcher`.

src/Elmish.WPF/ViewModel.fs
JDiLenarda: OK.

src/Elmish.WPF/ViewModelUtilities.fs
JDiLenarda: Simply, I see my functions from #52 as utilities for advanced uses :) If you think it doesn't fit a good style for module naming, I'll retrieve that.

cmeeren: Yes, I think you're right. Let's keep `Utilities`.

src/Elmish.WPF/ViewModel.fs
JDiLenarda: This one has been a little tougher than expected since `Dispatcher.Invoke` demands an `Action` delegate. I ended up with a `(unit -> unit) -> unit` parameter.

src/Elmish.WPF/ViewModel.fs
JDiLenarda: No need for this one anymore if we don't use `Dispatcher` in the module.

JDiLenarda:
(Free comment : I came for the F# and the XAML but get some good lesson about source control :) )

src/Elmish.WPF/ViewModel.fs
cmeeren: Of course, my bad. `(unit -> unit) -> unit` is correct.

src/Elmish.WPF/ViewModelUtilities.fs
```diff
@@ -0,0 +1,28 @@
...
+    (element: FrameworkElement)
+    (programRun: Program<'t, 'model, 'msg, BindingSpec<'model, 'msg> list> -> unit)
+    (program: Program<'t, 'model, 'msg, BindingSpec<'model, 'msg> list>) =
+  let mutable lastModel = None
+
+  let setState model dispatch =
+    match lastModel with
+    | None ->
+        let mapping = program.view model dispatch
+        let vm = ViewModel<'model,'msg>(model, dispatch, mapping, config, (fun f -> element.Dispatcher.Invoke f))
```
cmeeren: This line is a bit long now. Could you define `let uiDispatch (f: unit -> unit) = element.Dispatcher.Invoke f` and use this in the constructor? I think that's the last change.

cmeeren:
Looks ready to merge, thanks!

PR #61, opened 2018-10-03
Fix Parameter documentation in Bindings.fs

magicmonty:
Some Documentation comments in Bindings.fs were off (i.e. `<paramref="get">` instead of `<paramref="map">` and some paramrefs were missing.


cmeeren:
Thanks, much appreciated!

PR #79, opened 2019-04-15
Make TrySetMember return true for known bindings

### LINKED ISSUE #10: Two-way bindings gives 'Cannot save value from target back to source' in debug output
On a simple two-way binding from a textbox to a string on a model, I see the following output when running: -

```
System.Windows.Data Error: 8 : Cannot save value from target back to source. BindingExpression:Path=Name; DataItem='ViewModelBase`2' (HashCode=64479624); target element is 'TextBox' (Name=''); target property is 'Text' (type 'String') InvalidOperationException:'System.InvalidOperationException: Property path is not valid. 'System.Dynamic.DynamicObject+MetaDynamic' does not have a public property named 'Items'.
   at CallSite.Target(Closure , CallSite , Object , Object )
   at MS.Internal.DynamicPropertyAccessorImpl.SetValue(Object component, Object value)
   at MS.Internal.Data.PropertyPathWorker.SetValue(Object item, Object value)
   at MS.Internal.Data.ClrBindingWorker.UpdateValue(Object value)
   at System.Windows.Data.BindingExpression.UpdateSource(Object value)'
```

This doesn't crash the app, but repeated errors like these have an effort on WPF perf when they mount up.

### LINKED ISSUE #74: TwoBinding updating issue
First of all... Thanx for this very nice library!

I have this issue:
I've cloned SubBindings sample and modified with:
-- Program.fs
- added in the Counter Module
```
    let setConter value c =
        let a = { c with CounterValue = value + 1 }
        { a with StepSize = c.StepSize + 1 }

```
- added to type Msg
`    | SetCounter of CounterId * int`

- added to update function 
`    | SetCounter (cid,value) -> { m with AllCounters = m.AllCounters |> mapCounter cid (Counter.setCounter value) }`

- added to counterBindings
```
      "CounterValue" |> Binding.twoWay
        (fun (m, c) -> string c.CounterValue)
        (fun v (m, c) -> SetCounter (c.Id, int v))

``` 

-- MainWindows.xml
- modified the textblock showing CounterValue in 
`<TextBox Text="{Binding CounterValue}" Width="100" Margin="0,5,10,5" />`

Works all but this:
if i insert a number (eg 20) in the textbox and change focus:

1. the model is update (CounterValue is now 20 and StepSize is now 2)
2. the StepSize Slider is update to 2 
**3. the texbox remain 20 (ISSUE)**

now if i hit the "+" button, correctly:
1. the model is update (CounterValue is now 23 and StepSize remain 2)
2. the StepSize Slider remain 2 
3. the texbox is updated to 23

Why i've th bold issue?
Thanx
PS sorry for my poor english, i'm italian
(i attach the modified files)
[Files.zip](https://github.com/elmish/Elmish.WPF/files/2997467/Files.zip)

### LINKED ISSUE #78: Checkboxes erroneously shown as failing validation
### Description

Using the latest version (2.0.0-beta-9) of Elmish.WPF, if you bind to a `DataGridCheckBoxColumn` in a `DataGrid`, then changing its  value results in an error being displayed by the validation template:

screenshot

#### Repro steps

1. Create a `DataGridCheckBoxColumn` whose `Binding` is set to a 2-way Elmish.WPF binding.

1. Click to change the value of the `CheckBox`

An example script can be found in the repository [BillHally/Elmish.WPF.DataGridCheckBoxColumnIssue](https://github.com/BillHally/Elmish.WPF.DataGridCheckBoxColumnIssue). 

#### Expected behavior

CheckBox is displayed without incorrectly indicating that validation has failed.

#### Actual behavior

CheckBox *is* displayed incorrectly indicating that validation has failed.

#### Known workarounds

You could probably set a custom validation template that ignores errors (I've not tried).

#### Related information 

This is caused by [ViewModel.TrySetValue](https://github.com/elmish/Elmish.WPF/blob/5895b83da23beaec93dbd2f6b89b4e243fb8981c/src/Elmish.WPF/ViewModel.fs#L313) returning `false`. The reason it does so seems to be to avoid an [updating issue](https://github.com/elmish/Elmish.WPF/issues/10#issuecomment-416703094).

However, I've performed some tests with the script I linked to above, and it seems that this issue was fixed by the addition of this changeset ["adding UI thread synchronization option for .NET"](https://github.com/elmish/elmish/commit/07c021c0403188e7c16678e08ba0d9c79f6dbe50) to Elmish.

Consequently, I think that `TrySetValue` [should be changed to return true](https://github.com/BillHally/Elmish.WPF/commit/db90171bb71328ca80bb02afcd54b53298c8ecdc) when passed a binder whose name it recognises.

Making this change would resolve:

* [TwoBinding updating issue #74](https://github.com/elmish/Elmish.WPF/issues/74) (providing that `UpdateSourceTrigger` is set to `PropertyChanged`)
* [The error reported in issue #10](https://github.com/elmish/Elmish.WPF/issues/10#issue-238384835)

I'll send a pull request for the relevant changes.

BillHally:
This resolves #78

(Edit by @cmeeren for GH hook: resolves #74, resolves #10)

### DIFF
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -313,18 +313,22 @@ and [<AllowNullLiteral>] ViewModel<'model, 'msg>
   override __.TrySetMember (binder, value) =
     log "[VM] TrySetMember %s" binder.Name
     match bindings.TryGetValue binder.Name with
-    | false, _ -> log "[VM] TrySetMember FAILED: Property %s doesn't exist" binder.Name
+    | false, _ ->
+      log "[VM] TrySetMember FAILED: Property %s doesn't exist" binder.Name
+      false
     | true, binding ->
         match binding with
         | TwoWay (_, set)
         | TwoWayValidate (_, set, _) ->
             dispatch <| set value currentModel
+            true
         | TwoWayIfValid (_, set) ->
             match set value currentModel with
             | Ok msg ->
                 removeError binder.Name
                 dispatch msg
             | Error err -> setError err binder.Name
+            true
         | OneWay _
         | OneWayLazy _
         | OneWaySeq _
@@ -334,9 +338,7 @@ and [<AllowNullLiteral>] ViewModel<'model, 'msg>
         | SubModel _
         | SubModelSeq _ ->
             log "[VM] TrySetMember FAILED: Binding %s is read-only" binder.Name
-    // This function should always return false, otherwise the UI may execute a
-    // subsequent get which may may return the old value
-    false
+            true
 
   interface INotifyPropertyChanged with
     [<CLIEvent>]
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -314,8 +314,8 @@ and [<AllowNullLiteral>] ViewModel<'model, 'msg>
     log "[VM] TrySetMember %s" binder.Name
     match bindings.TryGetValue binder.Name with
     | false, _ ->
-      log "[VM] TrySetMember FAILED: Property %s doesn't exist" binder.Name
-      false
+        log "[VM] TrySetMember FAILED: Property %s doesn't exist" binder.Name
+        false
     | true, binding ->
         match binding with
         | TwoWay (_, set)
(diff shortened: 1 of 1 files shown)


src/Elmish.WPF/ViewModel.fs
```diff
@@ -313,18 +313,22 @@ and [<AllowNullLiteral>] ViewModel<'model, 'msg>
   override __.TrySetMember (binder, value) =
     log "[VM] TrySetMember %s" binder.Name
     match bindings.TryGetValue binder.Name with
-    | false, _ -> log "[VM] TrySetMember FAILED: Property %s doesn't exist" binder.Name
+    | false, _ ->
+      log "[VM] TrySetMember FAILED: Property %s doesn't exist" binder.Name
```
cmeeren: Please indent these two lines one more level

Issue #95, opened 2019-08-20
Obsolete TODO comment?


bender2k14:
There is a [TODO comment](https://github.com/elmish/Elmish.WPF/blob/992daf7e2dfa679a34bceffe31745fd87bed90eb/src/Samples/SubModelSeq/Program.fs#L165-L167) in the SubModelSeq example:
```
// TODO: moving up/down must be done at the correct level of hierarchy;
// currently if the hierarchy is A (B, C), D (where B and C are children of A)
// then D must be moved up three times before it's placed before A
```

I played around with this example, but I didn't find any bugs.  Everything seems to me to be working as expected.

Should this comment be removed?  If so, I can make a PR for that.

cmeeren:
Thanks for the heads-up! I tested and it seems to work fine. No idea why, but the bug was uninteresting enough in the first place (in the context of it only being a simple demo to show unrelated stuff), so I won't spend more time thinking about it.

I removed the comment while I was at it.

bender2k14:
Oh, you are too quick.  I just reproduce the problem.

The list of all counters needs to be `[A, B, C, D]` in that order with `B` and `C` children of `A`.  Actually, just `[A, B, C]` with 'B' a child of `A' suffices.

More specifically,
1. Run the application
2. Let the one exiting counter be `A`
2. Add a child counter `B` to `A`
3. Add a counter `C` without a parent via the button at the top
4. Click the up arrow button on counter `C`

Expected behavior:
`C` is the top counter

Actual behavior:
View is unchanged.  The model changed though from `[A, B, C]` to `[A, C, B]`.

cmeeren:
Ah, the order in which you add them is important.

I reverted the commit now.

Again, since the bug isn't really relevant to what's being demoed, I don't really care that much about it. If you want the challenge of finding a good solution to it (that doesn't needlessly complicate the demo), feel free to have a go at it, but do it for yourself, not for me. :)

PR #99, opened 2019-08-21
Fix reordering bug in tree view sample

bender2k14:
Resolves issue #95.

I think this is one way to fix the bug.  I switched the model from representing the graph (abstractly) as a list vertices and a list of edges to representing it directly.  Specifically, I created
```
module Tree =
  type Node<'a> =
    { Data: 'a
      Children: Node<'a> list }
```
which seems to be what Haskell calls a [Rose Tree](https://en.wikipedia.org/wiki/Rose_tree).

I just got it working, so my code might be a bit messy.  I can clean it up still in this PR or in a future one.

### DIFF
--- a/src/Samples/SubModelSeq.Views/MainWindow.xaml
+++ b/src/Samples/SubModelSeq.Views/MainWindow.xaml
@@ -5,9 +5,7 @@
     Title="SubModelSeq" Height="800" Width="1100"
     WindowStartupLocation="CenterScreen">
   <StackPanel Margin="0,20,0,10">
-    <StackPanel
-        Orientation="Horizontal"
-        HorizontalAlignment="Center">
+    <StackPanel Orientation="Horizontal" HorizontalAlignment="Center">
       <Button Command="{Binding AddCounter}" Content="Add counter" Width="150" Margin="10,0,10,20" />
       <Button Command="{Binding ToggleGlobalState}" Content="Toggle global state" Width="150" Margin="10,0,10,20" />
     </StackPanel>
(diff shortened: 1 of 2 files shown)


bender2k14:
Oh, I meant to say that I changed the functionality lightly.  The behavior used to be that new children were added at the end.  I switched this so that new children are added at the beginning.

Adding at the beginning makes the code (ever so slightly) simpler and (I think) is a bit more (functionally) idiomatic since adding a new head to an immutable list is the ideal case (because it only takes constant time).  Without a business reason to otherwise, I decided to take this approach.

For the purposes of this PR, it makes verifying that the bug is gone a bit tricker.  The sequence of button clicks used to prepare the state that would exhibit now produces a different view.

I think it is easier to see that the bug is gone by analyzing the code.  Moving up or down only modifies the list of children for a single node (or a single node in the list of root nodes...one code clean up is to store those nodes as the children in a dummy root node).  WPF is still given the lists of vertices and edges as before.  WPF doesn't care about the list of edges.  Crucially, the order of the vertices is computed from the tree via a [pre-order traversal](https://en.wikipedia.org/wiki/Tree_traversal#Pre-order_(NLR)).

cmeeren:
Thanks, looks good! Would be great if you could remove the unused `Utils` module. Please go ahead with whatever else cleanup you want to do, and let me know when you're ready to merge.

bender2k14:
Certainly...I removed the `Utils` module and (I think) all other unused code.  That is all the clean up I had planned to do.

I am ready to merge.

cmeeren:
Fantastic work, thank you so much! 👍 🎉

PR #103, opened 2019-08-22
Bug fix 102 start using elm config.measure limit ms

### LINKED ISSUE #102: ElmConfig.MeasureLimitMs not read
[`ElmConfig.MeasureLimitMs`](https://github.com/elmish/Elmish.WPF/blob/master/src/Elmish.WPF/Config.fs#L11) is only written to when defining [`ElmConfig.Default`](https://github.com/elmish/Elmish.WPF/blob/master/src/Elmish.WPF/Config.fs#L13-L17) (where it is set to `1`) and is _never_ read.

Currently [`measure`](https://github.com/elmish/Elmish.WPF/blob/master/src/Elmish.WPF/ViewModel.fs#L117-L126) and [`measure2`](https://github.com/elmish/Elmish.WPF/blob/master/src/Elmish.WPF/ViewModel.fs#L128-L137) contain the line
```
if sw.ElapsedMilliseconds > 0L then
```

Are these lines suppose to be
```
if sw.ElapsedMilliseconds > int64 config.MeasureLimitMs then
```

If so, preserving the default behavior would include initializing `ElmConfig.Default.MeasureLimitMs` to `0`.

Furthermore, it would be convenient if the type of `MeasureLimitMs` were already `int64` in order to avoid the conversion.

bender2k14:
Fixes #102

### DIFF
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -121,7 +121,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
         let sw = System.Diagnostics.Stopwatch.StartNew ()
         let r = f x
         sw.Stop ()
-        if sw.ElapsedMilliseconds > 0L then
+        if sw.ElapsedMilliseconds >= int64 config.MeasureLimitMs then
           log "[%s] %s (%ims): %s" propNameChain callName sw.ElapsedMilliseconds name
         r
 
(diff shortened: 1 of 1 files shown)


cmeeren:
Actually, the default should be `1` and the comparison should be `>=`, not `>`. Then it fits with the property documentation. Could you do that?

bender2k14:
Oh, sure.  Of course.  That is definitely the right comparison.  On it.

bender2k14:
I force pushed the change

cmeeren:
Perfect, thank you so much!

cmeeren:
Will release shortly as v3.2.1

bender2k14:
I am I no hurry

Issue #33, opened 2018-02-07
Binding to events?


dsyme:
Hi all

I'm seeing plenty of examples of Xaml where binding to events via code behind is needed, e.g.  see below for a snippet of Xamarin code.

My question is - what's the right way to go about this in Elmish.WPF?  

From what I've seen there doesn't seem much alternative to writing the code-behind event handler (apart from massive things like [this](https://github.com/xamarin/xamarin-forms-samples/tree/master/Behaviors/EventToCommandBehavior) with very intrusive Xaml) - but how would the event handler capture the Elmish dispatch routine?

Xaml:

            <ListView x:Name="ItemsListView" ItemsSource="{Binding [Items]}" ItemSelected="OnItemSelected" ....  >          

code behind:

        async void OnItemSelected(object sender, SelectedItemChangedEventArgs args)
        {
            ...
        }

2sComplement:
I haven't yet needed to do this in Elmish.WPF, but in vanilla MVVM you would probably get a reference to the Control's `DataContext` (its ViewModel) in code-behind and do something with it there. In Elmish.WPF with the ViewModel being a `DynamicObject`, this should still be possible even if not the cleanest solution (isn't code-behind-to-view-model always a little dirty?).

Stepping back for a second though, in your example you could just hook up a two-way binding to the `ListView`'s `SelectedItem` no?

dsyme:
@2sComplement Thanks!

Yes, it looks like I could use SelectedItem, but I'm trying to work out whether Elmish.XamarinForms is complete enough to do everything necessary.  There are a lot of events in XamarinForms that don't have a corresponding 2-way binding.

2sComplement:
@dsyme Ideally I try to keep code-behind limited to view-only types of interactions. I'd be curious to know what other events would be applicable here, where they would need some way of binding to the ViewModel.

dsyme:
@2sComplement Some of the samples may be sloppy Xaml.  Another example:

		<ToolbarItem Text="Save" Clicked="Save_Clicked" />

should be:

		<ToolbarItem Text=Save Command= SaveCommand />

But I expect there is a lot of this kind of thing around in samples, including educational ones.  Also, in the samples I'm translating I also see event hookup for ``ItemAppearing``, which AFAIK is not accessible via command. More generally, to take an example, the list of events on Xamarin [ListView](https://developer.xamarin.com/api/type/Xamarin.Forms.ListView/#Public_Events) is:

[...]

PR #108, opened 2019-09-07
Hide/Show focus gaining text box

bender2k14:
In the sample about behaviors, added the ability to hide/show the text box that gains focus each time it appears.  This addition was suggested in #33 [here](https://github.com/elmish/Elmish.WPF/issues/33#issuecomment-528830890).


cmeeren:
Wonderful, thanks!

By the way, if you paste a link to a GitHub comment such as `https://github.com/elmish/Elmish.WPF/issues/33#issuecomment-528830890` it will look like this:

https://github.com/elmish/Elmish.WPF/issues/33#issuecomment-528830890

bender2k14:
Oh, thanks for the suggestion about comment linking.  I will do that.

Issue #114, opened 2019-09-12
Throttling individual bindings


bender2k14:
I want to throttle an individual binding.  I started the discussion in https://github.com/elmish/elmish/issues/189#issuecomment-530824226.  I will briefly summarize the relevant parts.

My definition of throttling comes from [this SO answer](https://stackoverflow.com/questions/25991367/difference-between-throttling-and-debouncing-a-function/25991510#25991510).  That answer contrasts throttling with debouncing.  Both of those concepts involve swallowing/dropping some amount of data.  A related idea could be called limiting (as in [rate limiting](https://en.wikipedia.org/wiki/Rate_limiting)), which limits the rate at which data is allowed to flow by but doesn't drop any data.  So although I am primarily motivated by throttling, debouncing and limiting are so similar that we can probably make all of these possible in one go.

My specific use case is a [Slider](https://docs.microsoft.com/en-us/dotnet/api/system.windows.controls.slider?view=netframework-4.8) with discrete interval of length 8000.  As the user drags the thumb (as it is called) of the slider, each integral value encountered dispatches a message.  My `update` function is trivial: it just saves the new value of the slider in the model.  It is not computationally expensive.  The part that is computationally expensive is `update` being called ~4000 in a single flick of the user's wrist.

@cmeeren (as stated in https://github.com/elmish/elmish/issues/189#issuecomment-530688459) and I like this idea.  I am currenting having fun struggling with the details to make it happen :)

cmeeren:
What I imagine, partly from the top of my head, is this:

```diff
and internal TwoWayData<'model, 'msg> = {
  Get: 'model -> obj
  Set: obj -> 'model -> 'msg
+ ModifyDispatch: Dispatch<'msg> -> Dispatch<'msg>
}
```

```diff
  static member twoWay
      (get: 'model -> 'a,
-      set: 'a -> 'model -> 'msg)
+      set: 'a -> 'model -> 'msg,
+      ?modifyDispatch: Dispatch<'msg> -> Dispatch<'msg>)
      : string -> Binding<'model, 'msg> =
    TwoWayData {
      Get = get >> box
      Set = unbox<'a> >> set
+     ModifyDispatch = defaultArg modifyDispatch id
    } |> createBinding
```

[...]

PR #115, opened 2019-09-12
Code clean up while working on #114

bender2k14:
While working on #114, I made many small code improvements.  All changes are meant to be behavior preserving.  I find it helpful to not keep changes to myself for too long, so I thought I would contribute this PR now.

### DIFF
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -48,7 +48,7 @@ type internal VmBinding<'model, 'msg> =
       * toMsg: (obj -> 'msg)
       * getWindow: (unit -> Window)
       * isModal: bool
-      * onCloseRequested: 'msg voption
+      * onCloseRequested: (unit -> unit)
       * preventClose: bool ref
       * windowRef: WeakReference<Window>
   | SubModelSeq of
@@ -127,24 +127,16 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
 
   let measure2 name callName f =
     if not config.Measure then f
-    else
-      fun x y ->
-        let sw = System.Diagnostics.Stopwatch.StartNew ()
-        let r = f x y
-        sw.Stop ()
-        if sw.ElapsedMilliseconds >= int64 config.MeasureLimitMs then
-          log "[%s] %s (%ims): %s" propNameChain callName sw.ElapsedMilliseconds name
-        r
+    else fun x -> measure name callName (f x)
 
   let showNewWindow
       (winRef: WeakReference<Window>)
       (getWindow: unit -> Window)
       dataContext
       isDialog
-      (onCloseRequested: 'msg voption)
+      (onCloseRequested: unit -> unit)
       (preventClose: bool ref)
-      initialVisibility
-      dispatch =
+      initialVisibility =
     Application.Current.Dispatcher.Invoke(fun () ->
       let guiCtx = System.Threading.SynchronizationContext.Current
       async {
@@ -155,13 +147,13 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
           ev.Cancel <- !preventClose
           async {
             do! Async.SwitchToThreadPool()
-            onCloseRequested |> ValueOption.iter dispatch
+            onCloseRequested ()
           } |> Async.StartImmediate
         )
         do! Async.SwitchToContext guiCtx
-        if isDialog then win.ShowDialog () |> ignore
-        else
-          win.Visibility <- initialVisibility
+        if isDialog
+        then win.ShowDialog () |> ignore
+        else win.Visibility <- initialVisibility
       } |> Async.StartImmediate
     )
 
@@ -210,7 +202,8 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
         let getBindings = measure name "bindings" d.GetBindings
         let toMsg = measure name "toMsg" d.ToMsg
         match getModel initialModel with
-        | ValueNone -> SubModel (ref ValueNone, getModel, getBindings, toMsg, d.Sticky)
+        | ValueNone ->
+            SubModel (ref ValueNone, getModel, getBindings, toMsg, d.Sticky)
         | ValueSome m ->
             let chain = getPropChainFor name
             let vm = ViewModel(m, toMsg >> dispatch, getBindings (), config, chain)
@@ -219,6 +212,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
         let getState = measure name "getState" d.GetState
         let getBindings = measure name "bindings" d.GetBindings
         let toMsg = measure name "toMsg" d.ToMsg
+        let onCloseRequested = fun () -> d.OnCloseRequested |> ValueOption.iter dispatch
         match getState initialModel with
         | WindowState.Closed ->
             SubModelWin (
@@ -228,9 +222,10 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
               toMsg,
               d.GetWindow,
               d.IsModal,
-              d.OnCloseRequested,
+              onCloseRequested,
               ref true,
-              WeakReference<_>(null))
+              WeakReference<_>(null)
+            )
         | WindowState.Hidden m ->
             let chain = getPropChainFor name
             let vm = ViewModel(m, toMsg >> dispatch, getBindings (), config, chain)
@@ -238,8 +233,8 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
             let preventClose = ref true
             log "[%s] Creating hidden window" chain
             showNewWindow
-              winRef d.GetWindow vm d.IsModal d.OnCloseRequested
-              preventClose Visibility.Hidden dispatch
+              winRef d.GetWindow vm d.IsModal onCloseRequested
+              preventClose Visibility.Hidden
             SubModelWin (
               ref <| WindowState.Hidden vm,
               getState,
@@ -247,7 +242,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
               toMsg,
               d.GetWindow,
               d.IsModal,
-              d.OnCloseRequested,
+              onCloseRequested,
               preventClose,
               winRef
             )
@@ -258,8 +253,8 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
             let preventClose = ref true
             log "[%s] Creating and opening window" chain
             showNewWindow
-              winRef d.GetWindow vm d.IsModal d.OnCloseRequested
-              preventClose Visibility.Visible dispatch
+              winRef d.GetWindow vm d.IsModal onCloseRequested
+              preventClose Visibility.Visible
             SubModelWin (
               ref <| WindowState.Visible vm,
               getState,
@@ -267,7 +262,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
               toMsg,
               d.GetWindow,
               d.IsModal,
-              d.OnCloseRequested,
+              onCloseRequested,
               preventClose,
               winRef
             )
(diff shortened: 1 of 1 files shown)


src/Elmish.WPF/Binding.fs
```diff
@@ -143,20 +143,20 @@ module internal BindingData =
       }
     | TwoWayData d -> TwoWayData {
         Get = unbox >> d.Get
-        Set = fun v m -> d.Set v (unbox m) |> box
```
cmeeren: To be honest I like the existing forms better for all changes in this file. All the boxing and unboxing is difficult enough to grasp without also using point-free style to remove some parameters of multi-parameter functions. I'd like to keep these multi-parameter functions explicitly as multi-parameter functions. Also see later comment on multi-parameter functions where the parameters are of "equal importance", as it were.

src/Elmish.WPF/ViewModel.fs
```diff
@@ -126,25 +126,16 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
...
-    if not config.Measure then f
-    else
-      fun x y ->
-        let sw = System.Diagnostics.Stopwatch.StartNew ()
-        let r = f x y
-        sw.Stop ()
-        if sw.ElapsedMilliseconds >= int64 config.MeasureLimitMs then
-          log "[%s] %s (%ims): %s" propNameChain callName sw.ElapsedMilliseconds name
-        r
+    fun x -> measure name callName (f x)
```
cmeeren: Are you 100% sure this is behaviour preserving? Multi-parameter functions and partial application can bite you in some cases (see e.g. https://github.com/elmish/Elmish.WPF/issues/33#issuecomment-530536977)

src/Elmish.WPF/ViewModel.fs
```diff
@@ -196,21 +187,22 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
...
         let canExec = measure name "canExec" d.CanExec
-        let execute _ = exec currentModel |> ValueOption.iter dispatch
-        let canExecute _ = canExec currentModel
+        let execute _ = currentModel |> exec |> ValueOption.iter dispatch
+        let canExecute _ = currentModel |> canExec
         Cmd (Command(execute, canExecute, false), canExec)
     | CmdParamData d ->
         let exec = measure2 name "exec" d.Exec
         let canExec = measure2 name "canExec" d.CanExec
-        let execute param = exec param currentModel |> ValueOption.iter dispatch
```
cmeeren: TBH I like this better as it is. I don't see a need to use pipelines for multi-parameter functions unless it actually makes things more clear, and in this case, `param` and `currentModel` are both equally "important" parameters.

For consistency, I'd then also like to keep `(can)exec currentModel` above.

src/Elmish.WPF/ViewModel.fs
```diff
@@ -196,21 +187,22 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
...
-        let canExecute param = canExec param currentModel
+        let execute param = currentModel |> exec param |> ValueOption.iter dispatch
+        let canExecute param = currentModel |> canExec param
         CmdParam <| Command(execute, canExecute, d.AutoRequery)
     | SubModelData d ->
         let getModel = measure name "getSubModel" d.GetModel
         let getBindings = measure name "bindings" d.GetBindings
         let toMsg = measure name "toMsg" d.ToMsg
         match getModel initialModel with
-        | ValueNone -> SubModel (ref ValueNone, getModel, getBindings, toMsg, d.Sticky)
```
cmeeren: I like this better as it currently is. I generally put short lines on the same line. (I do make some exceptions, for example if there are many cases and most of them have their own lines, or sometimes if I match against multiple cases.)

src/Elmish.WPF/ViewModel.fs
```diff
@@ -238,39 +231,37 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
...
               getBindings,
               toMsg,
               d.GetWindow,
               d.IsModal,
-              d.OnCloseRequested,
+              onCloseRequested,
               preventClose,
-              winRef
-            )
+              winRef)
```
cmeeren: For symmetry I generally like to keep the closing parenthesis at the same level of indentation as the line of the start parenthesis, so I'd like to keep this as it is.

src/Elmish.WPF/ViewModel.fs
```diff
@@ -238,39 +231,37 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
...
```
cmeeren: Same as above

src/Elmish.WPF/ViewModel.fs
```diff
@@ -357,25 +348,22 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
     | Cmd _
     | CmdParam _ ->
         false
-    | SubModel ((vm: ViewModel<obj, obj> voption ref), (getModel: 'model -> obj voption), getBindings, toMsg, sticky) ->
```
cmeeren: Can't remember why I put these annotations there in the first place, so please just make sure that this actually is behaviour preserving (not everything is with weakly typed stuff like this, even if it compiles).

src/Elmish.WPF/ViewModel.fs
```diff
@@ -357,25 +348,22 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
...
+    | SubModel (vm, getModel, getBindings, toMsg, sticky) ->
       match !vm, getModel newModel with
       | ValueNone, ValueNone -> false
       | ValueSome _, ValueNone ->
-          if sticky then false
```
cmeeren: Please keep this as it is

src/Elmish.WPF/ViewModel.fs
```diff
@@ -357,25 +348,22 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
...
+          vm := ValueSome <| ViewModel(m, toMsg >> dispatch, getBindings (), config, getPropChainFor bindingName)
           true
       | ValueSome vm, ValueSome m ->
           vm.UpdateModel m
           false
-    | SubModelWin
-        ( vmWinState, getState, getBindings, toMsg,
-          getWindow, isModal, onCloseRequested, preventClose, winRef) ->
-
+    | SubModelWin (vmWinState, getState, getBindings, toMsg, getWindow, isModal, onCloseRequested, preventClose, winRef) ->
```
cmeeren: Line's getting a bit long, but if you think it's better, I can accept it

src/Elmish.WPF/ViewModel.fs
```diff
@@ -502,8 +489,11 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
     | SubModelSelectedItem _ ->
         None
     | Cmd (cmd, canExec) ->
-        if canExec newModel = canExec currentModel then None else Some cmd
-    | CmdParam cmd -> Some cmd
+        if canExec newModel = canExec currentModel
```
cmeeren: I can accept this, but in general I see nothing wrong having a short if-then-else on a single line.

src/Elmish.WPF/ViewModel.fs
```diff
@@ -582,7 +572,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
         match binding with
         | TwoWay (_, set)
         | TwoWayValidate (_, set, _) ->
-            dispatch <| set value currentModel
+            currentModel |> set value |> dispatch
```
cmeeren: Please change to `set value currentModel |> dispatch` or leave as it was (cf. previous comment on multi-parameter functions and pipelines)

src/Elmish.WPF/ViewModel.fs
```diff
@@ -591,7 +581,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
                   (value :?> ViewModel<obj, obj>)
                   |> ValueOption.ofObj
                   |> ValueOption.map (fun vm -> getSubModelId vm.CurrentModel)
-                set value currentModel |> dispatch
+                currentModel |> set value |> dispatch
```
cmeeren: Same as above

src/Elmish.WPF/Binding.fs
bender2k14: Sure.  Sometimes point-free goes too far.  This is probably one of those times.

At the same time, I think putting `unbox` on the left helps to verify correctness.  That is the first hint I had that #116 might be a bug.  Getters should look like `unbox >> get` and setters should look like `unbox >> set >> box`.

In any case, this seems complicated enough that I have reverted all the changes in this file for this PR.  If I refactor this code to put it in a standard form (like above), then it will be in a separate PR related to #116.

src/Elmish.WPF/ViewModel.fs
bender2k14: Ok, reverted.

src/Elmish.WPF/ViewModel.fs
bender2k14: I prefer consistency in parallel structure over fewer lines.  If you insist though, I will change it back.

src/Elmish.WPF/ViewModel.fs
bender2k14: Ok.  I had changed these two to make them consistent with `WeakReference<_>(null))` above.  I just changed these two back and also changed the first case to match.

P.S.
When I don't make the exact change you request (like I did here), I will leave the conversation unresolved so I don't sneak anything by you ;)

src/Elmish.WPF/ViewModel.fs
bender2k14: Addressed in the same commit

src/Elmish.WPF/ViewModel.fs
bender2k14: Sure, reverted.

I "can't" even argue for consistency since I put `then` on its own line just as `else` is on its own line but then didn't also put `false` on its on line like the body of `else` is :P

And I definitely don't want `then \newline false`.  That is way to verbose reguarless of parallel structure.

src/Elmish.WPF/ViewModel.fs
bender2k14: I prefer to think that this line is long not because we didn't include a rather arbitrarily placed new line but because `SubModelWin` has so many arguments.  When I am in the area, I will keep my eyes open for ways that we can reduce the number of arguments.

src/Elmish.WPF/Binding.fs
cmeeren: Thanks! I get your point regarding `unbox >> set >> box`, though I think I still find that the current form is more easily understood.

src/Elmish.WPF/ViewModel.fs
cmeeren: Thanks!

src/Elmish.WPF/ViewModel.fs
cmeeren: That someone would be me, and you're probably right. But there are cases in this file of type annotations being required even when it doesn't seem it should. So it might be that it was needed earlier on, but not after some refactor.

src/Elmish.WPF/ViewModel.fs
cmeeren: Sure, though there's nothing wrong with splitting arguments across lines. I sometimes do, as you can see here and there.

src/Elmish.WPF/ViewModel.fs
bender2k14: Of course the correctness is the same, but I admit that I am not 100% sure that the performance is the same.  I will look into this later.  I am aware of the gotcha one can fall into memoizing a partially applied function.

In the meantime, how about this alternative instead?

src/Elmish.WPF/ViewModel.fs
bender2k14: Sure, I did your first change suggestion.

src/Elmish.WPF/ViewModel.fs
bender2k14: Addressed in same commit

src/Elmish.WPF/ViewModel.fs
bender2k14: I agree.  I found it easier to read at the time.  Since you will accept this, I will leave it as is.

cmeeren:
Okay, seems all comments are resolved. Are you ready for this to be merged?

bender2k14:
I squashed the commits.  Ready now.

Thanks :)

cmeeren:
Thanks for the work you're putting into Elmish.WPF! :)

PR #118, opened 2019-09-15
#114 wrap dispatch in bindings

bender2k14:
Exposes optional dispatch wrapping for `TwoWay`, `TwoWayValidate`, `SubModelSelectedItem`, `Cmd` and `CmdParam`, which is enough to enable users to throttle, debounce, or rate limit each message source #114.  In a future PR, we can make it easy for users to do those three things.

### DIFF
--- a/src/Elmish.WPF/Binding.fs
+++ b/src/Elmish.WPF/Binding.fs
@@ -2,6 +2,7 @@
 
 open System.Windows
 
+open Elmish
 
 [<RequireQualifiedAccess>]
 type WindowState<'model> =
@@ -52,23 +53,27 @@ and internal OneWaySeqLazyData<'model> = {
 and internal TwoWayData<'model, 'msg> = {
   Get: 'model -> obj
   Set: obj -> 'model -> 'msg
+  WrapDispatch: Dispatch<'msg> -> Dispatch<'msg>
 }
 
 and internal TwoWayValidateData<'model, 'msg> = {
   Get: 'model -> obj
   Set: obj -> 'model -> 'msg
   Validate: 'model -> string voption
+  WrapDispatch: Dispatch<'msg> -> Dispatch<'msg>
 }
 
 and internal CmdData<'model, 'msg> = {
   Exec: 'model -> 'msg voption
   CanExec: 'model -> bool
+  WrapDispatch: Dispatch<'msg> -> Dispatch<'msg>
 }
 
 and internal CmdParamData<'model, 'msg> = {
   Exec: obj -> 'model -> 'msg voption
   CanExec: obj -> 'model -> bool
   AutoRequery: bool
+  WrapDispatch: Dispatch<'msg> -> Dispatch<'msg>
 }
 
 and internal SubModelData<'model, 'msg> = {
@@ -98,6 +103,7 @@ and internal SubModelSelectedItemData<'model, 'msg> = {
   Get: 'model -> obj voption
   Set: obj voption -> 'model -> 'msg
   SubModelSeqBindingName: string
+  WrapDispatch: Dispatch<'msg> -> Dispatch<'msg>
 }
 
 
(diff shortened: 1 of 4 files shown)


src/Elmish.WPF/Binding.fs
```diff
@@ -391,13 +412,19 @@ type Binding private () =
   /// <summary>Creates a two-way binding.</summary>
   /// <param name="get">Gets the value from the model.</param>
   /// <param name="set">Returns the message to dispatch.</param>
+  /// <param name="wrapDispatch">
+  ///   Wraps of the dispatch function with additional behavior,
```
cmeeren: Typo, change `Wraps of the` to `Wraps the`, here and other similar places.

While you're at it, I usually aim for a line length of 80-90 characters (though I don't care that much if you'd rather keep it this way).

I skimmed the rest of the file; if you know you did anything unique further below, please let me know (everything looked equal).

cmeeren:
By the way, just to mention it - let's discuss the "future PR" in #114 or another issue first to get design decisions down on making it easy for users. For example, since RX combinators like throttling/debouncing etc. are inherently composable (e.g. `throttle 250 >> distinctUntilChanged`), I'm thinking it makes more sense to define such combinators than to add e.g. optional `throttleMs` and `debounceMs` parameters to all relevant overloads. And depending on how Elmish.WPF-specific these combinators end up, it may be better to implement them at the Elmish level or in a separate library (or is that already done?).

src/Elmish.WPF/Binding.fs
bender2k14: Fixed the documentation typo (via find replace) in a force pushed commit.

I added this documentation as a suggestion.  I don't mind replacing it with anything that you would like instead.

Yes, everything else was straightforward.  For example, I fixed all compile errors in the tests by passing in `id`.

bender2k14: I am indifferent to merging with these separate commits or squashing first.  Whatever you would prefer.

cmeeren:
Would be great it you could squash these, then I can merge. It'll make it easier during the next release to see what to add to the changelog.

bender2k14:
Squashed and force pushed

cmeeren:
Thanks a lot!

bender2k14:
Do you have a plan on when you expect to create a release with this new feature?  (No rush.  Just wondering so I can incorporate it into my work schedule.)

cmeeren:
I can release whenever you want. :) Would you like to add yourself to the author list?

https://github.com/elmish/Elmish.WPF/blob/0f3c5121cda4e9aa3074ebba43d5f873579555e7/src/Elmish.WPF/paket.template#L3

I think the point is to use your nuget username so that your nuget packages can be correlated, but I'm not sure, and there's no validation (in other projects I have used my full name). If you don't have one you can use your GitHub username.

bender2k14:
Yea, if it is no trouble for you, now would be great.  Then I can demo to my team the performance with throttling enabled.

Thank you so much for the offer!  I would love to be listed as an author.  I just registered "bender2k14" on NuGet.org...the same as here on GitHub.  I will use that.

cmeeren:
3.3.0 is in the pipeline and will be released shortly.

src/Elmish.WPF/Binding.fs
bender2k14: > Yes, everything else was straightforward. For example, I fixed all compile errors in the tests by passing in `id`.

Oh, I realized later that maybe I could have provided more insight here.  One of the compile errors in the tests was because the compiler was unable to determine which method to resolve.

Does that change your mind about the new version being 3.3.0 instead of 4.0.0?

Even if you are (attempting to) follow [SemVer](https://semver.org/), I think that it is vague enough that there is ample room for disagreement about whether or not an upgrade such as this, one that can introduce a compiler error, must increment the major version.  I am unsure if SemVer considers this to be a "backwards incompatible change".  I think merely incrementing the minor version (as you did) is fine.

cmeeren: Thanks for bringing this to my attention. I think that in this particular it's not a problem; it's just that the `fail` test helper is completely generic, and thus it can't distinguish between a call using no `canExecute` and only `wrapDispatch`, and a call using `canExecute` but not `wrapDispatch`. In real-world usage, I expect `canExecute` to be sufficiently typed that this won't cause a problem.

For the record: Overload resolution problems are IMHO the main argument against using overloads. The ergonomics are good enough that I think it's a good tradeoff, though. I'm inclined to consider any introduction of overload resolution errors a breaking change, but they can be hard to catch unless you're prescient and know exactly what to look for. Having tests for all overloads and parameter combinations certainly help.

Issue #107, opened 2019-09-07
Advantage of lazily creating the binding list 


bender2k14:
The type of `bindings` passed to Elmish.WPF is `unit -> Binding<'model, 'msg> list`.  In particular, it is a function; the list of bindings is only created after the function is called (and everytime the function is called unless there is some optimization going on in F# of which I am unaware).  Regardless of any efficiency concerns, the syntax is slightly more verbose when working with a function type than just that function's return type.

What are the advantages to having the list of bindings created lazily through a function call?  What would be lost if the type of `bindings` was simply `Binding<'model, 'msg> list`?

cmeeren:
From the top of my head, at the very least it has to be lazy for recursive `subModel` bindings, otherwise the bindings would be initialized as part of their own definition, leading to an infinite loop and stack overflow (though I think you'd get a compiler error, or at least a warning, before that).

bender2k14:
That makes sense.  That example is definitely recursive; no way around it.

I really like the approach F# takes with its syntax in these kinds of cases.  The simpler and more terse syntax is used more often for the more common and simpler cases (like immutable my defaubt or non-recursive by default).  To be mutable or recursive, additional syntax is required (`mutable` and `rec` respectively).

In the same way, I plan to change all of my current `bindings` definitions to be a list instead of a function to a list.  Same goes for `init`; I will remove those and replace them with record definitions named `model`.  At the top level, I will create the required functions for passing to Elmish.WPF.

bender2k14:
Maybe it would be appropriate to add more overloads of the static `Binding` member methods that take a list of bindings instead of a function to such a list.  Of course, this is not critical though.

cmeeren:
Given that the additional syntax is simply `let bindings () =` instead of `let bindings =` (and both the `Binding` members as well as the `mkProgramWpf` functions work with this), I'd rather keep it as-is, both to be consistent and to avoid confusion for users who need recursive sub-modules.

bender2k14:
Another benefit of having it be a function is improve type inference.  I prefer to drop the type annotations, and if I also drop the input `()` to make `bindings` a value, then the there is this compiler error:
> FS0030	Value restriction. The value `bindings` has been inferred to have generic type `val bindings : Binding<Model,'_a> list`. Either define `bindings` as a simple data term, make it a function with explicit arguments or, if you do not intend for it to be generic, add a type annotation.

PR #119, opened 2019-09-15
Extract counter in sub model sample

bender2k14:


src/Samples/SingleCounter/Program.fs
```diff
@@ -18,6 +18,8 @@ type Msg =
   | SetStepSize of int
   | Reset
 
+let canReset = (<>) (init ())
```
cmeeren: I don't like point-free style for this free-standing function (I mostly use point-free style for inline functions). Could you write this as `let canReset m = m <> init ()` or keep inline?

src/Samples/SubModel.Views/SubModel.Views.csproj
```diff
@@ -26,6 +26,12 @@
     <Reference Include="WindowsBase" />
   </ItemGroup>
 
+  <ItemGroup>
```
cmeeren: Please revert this file. VS does this automatically, but it's not needed. (.NET Core 3.0 will bring and end to the csproj workarounds.)

src/Samples/SubModel/Program.fs
```diff
@@ -4,6 +4,41 @@ open System
...
+    { Count = 0
+      StepSize = 1 }
+  
+  type Msg =
+    | Increment
+    | Decrement
+    | SetStepSize of int
+    | Reset
+    
+  let canReset = (<>) (init ())
```
cmeeren: Same as previous comment: Inline or write as not point-free.

src/Samples/SubModel/Program.fs
```diff
@@ -4,6 +4,41 @@ open System
...
+    | Reset -> init ()
+  
+  let bindings () : Binding<Model, Msg> list = [
+    "CounterValue" |> Binding.oneWay (fun m -> m.Count)
+    "Increment" |> Binding.cmd Increment
+    "Decrement" |> Binding.cmd Decrement
+    "StepSize" |> Binding.twoWay(
+      (fun m -> float m.StepSize),
+      int >> SetStepSize)
+    "Reset" |> Binding.cmdIf(Reset, (<>) (init ()))
```
cmeeren: Shouldn't this use `canReset`? (Unless inlined, of course)

src/Samples/SubModel/Program.fs
```diff
@@ -36,39 +71,24 @@ module Clock =
...
   type Model =
-    { Count: int
-      StepSize: int
+    { Counter: Counter.Model
       Clock: Clock.Model }
 
   let init =
-    { Count = 0
-      StepSize = 1
+    { Counter = Counter.init ()
```
cmeeren: Come to think of it, I'm thinking that since `init` is pure and does not depend on anything, perhaps it should simply be a value, not a function. What do you think? Perhaps this should be modified in all samples? (Could probably be a separate PR)

bender2k14:
> By the way, would have been great to have this spelled out in the PR description - I had to spend a few minutes trying to figure out the point of the changes before starting my review. :) (The motivation seems obvious in hindsight, as things generally do...)

Oh, I am sorry.  I thought the title said enough.  Next time I will make sure to elaborate in the description.

> Also, it would be great in general if you could just leave a quick note on how confident you are that the changes are correct (e.g. "sample tested and works" or "untested, but compiles and seems obvious it will work"). The samples in particular are not covered by automated tests (they're not even compiled in CI).

Sure thing.  In this case, I tested the sample and it (still) works :)

cmeeren:
> Oh, I am sorry. I thought the title said enough. Next time I will make sure to elaborate in the description.

No problem. The title described what, but not why (not in itself a problem). Without a written motivation, I just had to convince myself of why this was a good idea. It's been a while since I've looked at various parts of the code/samples. :)

src/Samples/SubModel/Program.fs
bender2k14: Yes, I prefer making it a value.  In fact, this just like by question about `bindings` #107.  I also agree with making this change in a separate PR.

Do you prefer still calling it `init`?  In my project at work, I have been calling this `initModel` (which is short for "initial model" instead of "initialize (model)", but I am not completely sold on the name.

src/Samples/SubModel.Views/SubModel.Views.csproj
bender2k14: Oh, shoot.  I thought I had.  Nice catch.

src/Samples/SubModel/Program.fs
cmeeren: Let's stick to `init`. 👍

src/Samples/SingleCounter/Program.fs
bender2k14: Since the plan is to change `init` from a function to a value, I definitely prefer the resulting inline expression.  I reverted this change.

Also rebased onto newest `master`.

src/Samples/SubModel/Program.fs
bender2k14: Reverted to inline

src/Samples/SubModel/Program.fs
bender2k14: Inlined expression

cmeeren:
Thanks! Do you mind squashing?

bender2k14:
Gladly :)

Done.

PR #122, opened 2019-09-17
reduced circular references among BindingData

bender2k14:
The DU `BindingData` has several cases, just three of which are recursive (by referencing `Binding` which references `BindingData`).  The non-recursive cases don't need to be defined using `and`.

Given these types, I removed all unnecessary uses of `and`.  All remaining uses of 'and' for these binding data types are necessary.


cmeeren:
Thanks! I like the consistency and symmetry of `and` for all these definitions and have seen similar patterns elsewhere, but I have no problems with this change.

Issue #116, opened 2019-09-12
Is d.ToMsg >> unbox correct?


bender2k14:
Consider this line:
https://github.com/elmish/Elmish.WPF/blob/5df8514ce4452883a1e417c9f6f479f545339f9e/src/Elmish.WPF/Binding.fs#L165

Of course the code compiles (since it is in `master`), but I am very surprised by the use of `unbox` there.  I don't think the types make sense.  I was expecting to see `box` so that the whole line would be
```
ToMsg = d.ToMsg >> box
```

Can someone double check if this line is correct?

bender2k14:
> Perhaps try with a value-typed/struct message?

Yes, I think this is a necessary condition to exhibit a bug since I think `box` and `unbox` are no-ops for references types.

> Could you test both alternatives? If box works (and you're sure you've hit that code path), feel free to make a PR.

Definitely.

bender2k14:
I understand the (un)boxing so much better now.

To start, the reason for the (un)boxing is that we want to expose fully/strongly typed binding methods to the user, but we can't* keep all the type parameters around.  When we throw away some of that information by converting from some type parameter to `obj`, we have treat value types and references types differently.

*At least, I think we can't.  The main issue I see is that `BindingData<'model, 'msg>` is recursive.  Specifically, it is recursive via the cases `SubModelData`, `SubModelWinData`, and `SubModelSeqData` (but not via `SubModelSelectedItemData`, so maybe its name should not start with `SubModel`?).  Anyway, the most troublesome type parameters that I see are `'bindingModel`, and `'bindingMsg` because they are the ones uses in the recursively.  If we try to hold onto them, then either we are just delaying the inevitable (i.e. delaying the mapping to `obj` for the type parameters of the children to the type parameters of the grandchildren), or we run into a situation that I actually think is impossible (which is only being able to hold onto a finite number of type parameters but having types for everything in a heterogeneous tree of arbitrary depth).

With that out of the way, here are some improvements that I see related to (un)boxing.

First, we might be able to hold onto more type parameters.  Holding onto more type parameters means that we can remove some calls to `box` and `unbox`.  This would improve performance slightly, but of course the primary advantage would be stronger types within `Elmish.WPF` to help ensure correctness.

[...]

PR #123, opened 2019-09-21
Code clean up 2019-09-20

bender2k14:
I made the following small improvements while reading the code today.

The code compiles, all tests pass, and I tried all the features in all the samples.

The most important change was adding additional type parameters to the binding data records, which is related to #116.  This is helpful for two specific reasons.
1. It is now possible to understand how the functions in each binding data record are related by looking at the definition of the record.  Without this change, I constantly had to go look at the arguments of the various overloaded methods (where the types were still as strong as possible) to see how the functions are related.  (Of course, that is also how I figured out what the type parameters should be in order to make this change.)
2. We are closer to my goal of separating (1) the lifting from simpler to more complex cases (such as lifting from [`oneWaySeq`](https://github.com/elmish/Elmish.WPF/blob/0f3887480b77358fc50a15c7ab5a1fe1c51603b6/src/Elmish.WPF/Binding.fs#L362-L373) to [`oneWaySeqLazy`](https://github.com/elmish/Elmish.WPF/blob/0f3887480b77358fc50a15c7ab5a1fe1c51603b6/src/Elmish.WPF/Binding.fs#L396-L409)) from (2) the boxing of those (most complex) cases.  Intuitively, I would like the flow to look like
```
   simpleBindingData     // expressed as arguments to an overloaded method
=> complexBindingData<'model, 'msg, 'a, ..., 'z>   // lifiting
=> complexBindingData<'model, 'msg, obj, ..., obj> // partial boxing
=> complexBindingData<obj, obj, obj, ..., obj>     // remaining boxing
```

### DIFF
--- a/src/Elmish.WPF/Binding.fs
+++ b/src/Elmish.WPF/Binding.fs
@@ -32,33 +32,33 @@ module WindowState =
     | ValueNone -> WindowState.Closed
 
 
-type internal OneWayData<'model> = {
-  Get: 'model -> obj
+type internal OneWayData<'model, 'a> = {
+  Get: 'model -> 'a
 }
 
-type internal OneWayLazyData<'model> = {
-  Get: 'model -> obj
-  Map: obj -> obj
-  Equals: obj -> obj -> bool
+type internal OneWayLazyData<'model, 'a, 'b> = {
+  Get: 'model -> 'a
+  Map: 'a -> 'b
+  Equals: 'a -> 'a -> bool
 }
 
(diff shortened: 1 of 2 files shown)


cmeeren:
Thanks a lot!

The two last commits aren't "code cleanup", but I can accept this PR as-is. 👍

bender2k14:
Thanks. Time I should call it ` improvements_yyyy-mm-dd` ;)

cmeeren:
Yep ;) Though they could easily be PRs of their own, too. :)

PR #124, opened 2019-09-21
Improve samples

bender2k14:
In some issue (which I was unable to find last time I looked), we discussed simplifying `init` from a function to a value in the Counter samples.  I did this for the SimpleCounter and SubModel samples.

Then I tried to make a similar change to the SubModelSeq sample.  I tried to extract the Counter code so that it looks essentially identical to the other two samples.  That didn't work.  It seems difficult to have a `SubModel` binding within a `SubModelSeq` binding.  Included some code clean up for the SubModelSeq sample anyway.  There was an unused binding.

I noticed that the NewWindow sample used `mkProgramWpf` even though it always returned `Cmd.None` for the command, so I simplified that to `mkSimpleWpf`.

Lastly is minor code clean up in all other samples.

I didn't test anything yet.  I can test tonight.

### DIFF
--- a/src/Samples/EventBindingsAndBehaviors/Program.fs
+++ b/src/Samples/EventBindingsAndBehaviors/Program.fs
@@ -54,9 +54,9 @@ let bindings () : Binding<Model, Msg> list = [
 
 
 [<EntryPoint; STAThread>]
-let main argv =
+let main _ =
   Program.mkSimpleWpf init update bindings
   |> Program.withConsoleTrace
   |> Program.runWindowWithConfig
-      { ElmConfig.Default with LogConsole = true; Measure = true }
-      (MainWindow())
+     { ElmConfig.Default with LogConsole = true; Measure = true }
+     (MainWindow())
--- a/src/Samples/NewWindow/Program.fs
+++ b/src/Samples/NewWindow/Program.fs
@@ -26,8 +26,7 @@ module App =
   let init () =
     { Win1State = WindowState.Closed
       Win1Input = ""
-      Win2 = None },
-    Cmd.none
+      Win2 = None }
 
   type Msg =
     | ShowWin1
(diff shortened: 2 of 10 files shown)


cmeeren:
Thanks a lot! :)

> Then I tried to make a similar change to the SubModelSeq sample. I tried to extract the Counter code so that it looks essentially identical to the other two samples. That didn't work. It seems difficult to have a `SubModel` binding within a `SubModelSeq` binding.

I don't get it; as far as I can see, the reason you can't have a static `init` in the `SubModelSeq` sample is that it's nondeterministic (via `Counter.create`, which uses `Guid.NewGuid ()`).

In any case, I'm ready to merge this once you have tested it and otherwise are happy with it. :)

bender2k14:
I tested NewWindows and SubModelSeq; everything still works.  I am happy with this.

> I don't get it; as far as I can see, the reason you can't have a static `init` value in the `SubModelSeq` sample is that `init ()` is nondeterministic (via `Counter.create ()`, which uses `Guid.NewGuid ()`).

Certainly.  So I tried to split that nondeterminism from the counter code we know so well from the SimpleCounter and SubModel samples.  Much like how I recently extracted the counter code in the SubModel sample.

After I make that split so that some parent model (called `CounterWithId`) has the ID and the counter has a whole, I can't figure out how to create the mapping `ToMsg: 'Counter.Msg-> CounterWithId.Msg` since `CounterWithId.Msg = CounterMsg of Id * Counter.Msg`.  If the `Id` is removed from that message, then it just kicks the problem up one level.  If some sub-model doesn't know the ID of the `SubModelSeq` item to which it belongs, then I see no way to properly create that message.  That is why I said
> It seems difficult to have a `SubModel` binding within a `SubModelSeq` binding.

So difficult that it seems impossible to me.

cmeeren:
Yeah, `subModelSeq` (which is very much an Elmish.WPF construct, not an Elmish one) requires each item/sub-model to be uniquely identified. But there should be no problems having `subModel` inside `subModelSeq`. If you have a separate message type for the `subModel`, it's straightforward. If you want to use the same message type as the parent `subModelSeq` bindings, you just have to pass along the ID in the `toBindingModel` parameter of `subModel` so that it can be used in messages.

PR #126, opened 2019-09-25
Make model/dispatch available to getWindow in Binding.subModelWin

### LINKED ISSUE #33: Binding to events?

cmeeren:
Closes #33

@bender2k14 does this PR look good to you from a boxing/unboxing/type perspective?

By the way, the diff "boundaries" are a bit weird around the new `Binding` overloads. I have simply changed the `getWindow` signature for each of the three `subModelWin` overloads (with slight edits to the docs), and made copies of each of them with the old signature, delegating to the new. The diff makes it seem like I have made extensive edits to the docs because it has identified the "wrong boundaries", so to speak.

### DIFF
--- a/src/Elmish.WPF/Binding.fs
+++ b/src/Elmish.WPF/Binding.fs
@@ -189,8 +189,7 @@ module internal BindingData =
         GetBindings = d.GetBindings
         ToMsg = d.ToMsg >> box
         GetWindow =
-          let getWin = d.GetWindow  // No idea why this is needed
-          fun (m: obj) (d: Dispatch<obj>) -> getWin (unbox m) (unboxDispatch d)
+          fun (m: obj) (disp: Dispatch<obj>) -> d.GetWindow (unbox m) (unboxDispatch disp)
         IsModal = d.IsModal
         OnCloseRequested = d.OnCloseRequested |> ValueOption.map box
       }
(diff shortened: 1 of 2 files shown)


src/Elmish.WPF/Binding.fs
```diff
@@ -188,7 +188,9 @@ module internal BindingData =
         GetState = unbox >> d.GetState
         GetBindings = d.GetBindings
         ToMsg = d.ToMsg >> box
-        GetWindow = d.GetWindow
+        GetWindow =
+          let getWin = d.GetWindow  // No idea why this is needed
```
cmeeren: I'm particularly puzzled by why this temporary copy is necessary. Won't compile otherwise.

bender2k14: It is because you shadowed `d`.  Don't worry if you think that was a particularly dumb mistake.  It still took me ~10 minutes to figure that out :(

bender2k14:
> @bender2k14 does this PR look good to you from a boxing/unboxing/type perspective?

Yep, look perfect :)

> The diff makes it seem like I have made extensive edits to the docs because it has identified the "wrong boundaries", so to speak.

I hate when that happens :(

src/Elmish.WPF/Binding.fs
cmeeren: Thanks! 🤣 Totally did not see that.

PR #129, opened 2019-09-26
Remove paket and FAKE; change release process; add SourceLink

### LINKED ISSUE #127: Remove Paket and FAKE and simplify release?
Originally commented in https://github.com/elmish/Elmish.WPF/issues/106#issuecomment-535350442

### Paket and FAKE

Paket and FAKE are additional tools in the toolchain that may not be needed. I'm usually in favour of using the simplest tooling that gets the job adequately done. I'm not convinced Paket or FAKE adds much to Elmish.WPF development/maintenance, particularly since everyone using WPF is likely to be using Visual Studio anyway and since the built script is trivial. In fact, I think that using just NuGet and VS (or plain `dotnet` CLI if needed) may be simpler for contributors than using FAKE and Paket (see e.g. https://github.com/elmish/Elmish.WPF/issues/106#issuecomment-535339702). Therefore I am considering removing all Paket and FAKE stuff and just use NuGet references in the project files.

### Releases and `stable-*` brances vs. `master`

I'm also considering removing the `stable-*` branches and just triggering releases from `master` using tags. I have never used the `stable-*` branches to release anything on an old major version, and I can't imagine doing it in the foreseeable future. If it is ever needed, we can just branch out from the latest release commit for a certain major release and backport fixes to that.

---

Check out the [remove-paket-fake](https://github.com/elmish/Elmish.WPF/tree/remove-paket-fake) branch to see how I envision it to be after simplifying it.

cmeeren:
Closes #127


src/Elmish.WPF/Elmish.WPF.fsproj
```diff
@@ -15,12 +27,24 @@
     <Compile Include="Utils.fs" />
     <Compile Include="Cmd.fs" />
     <Compile Include="Program.fs" />
-    <None Include="paket.references" />
   </ItemGroup>
+  
+  <ItemGroup>
+    <PackageReference Include="Elmish" Version="3.0.3" />
```
inosik: Would be good to keep the version constraint from `paket.dependencies` to avoid unintentional updates to a potentionally breaking Elmish v4.

cmeeren: Good idea. Do you know the exact syntax?

inosik: This should work:

```suggestion
    <PackageReference Include="Elmish" Version="[3.0.3, 4)" />
```

<https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges-and-wildcards>

cmeeren: Perfect, thanks!

cmeeren:
I notice that the NuGet package now depends on System.ValueTuple. Any ideas why, or if it should be removed, and if so, how?

inosik:
_System.ValueTuple_ is included automatically by the SDK if you target .NET Framework. `Elmish.WPF` currently targets `net461`. However, it isn't included if you target `netstandard2.0`. That should be done anyways to improve support for .NET Core.

I guess you can disable it, if the package doesn't use struct tuples. `Paket.Restore.targets` does that with this MSBuild property:

``` xml
<DisableImplicitSystemValueTupleReference>true</DisableImplicitSystemValueTupleReference>
```

cmeeren:
However, it isn't included if you target `netstandard2.0`. That should be done anyways to improve support for .NET Core.

We will soon multi-target `net461` and `netcoreapp3.0`. We can't target `netstandard2.0` due to the WPF dependencies.

> I guess you can disable it, if the package doesn't use struct tuples. `Paket.Restore.targets` does that with this MSBuild property:
> 
> ```
> <DisableImplicitSystemValueTupleReference>true</DisableImplicitSystemValueTupleReference>
> ```

Thanks, I've added that now.

.github/CONTRIBUTING.md
```diff
@@ -63,6 +63,8 @@ The  `stable-` prefixed branches are what’s actually deployed. When releasing
 
 For maintainers.
 
-* Make changes as necessary and merge into master. **The changes must include updated release notes with a new version number**, as this is automatically used as the Nuget version when deploying.
-* Create a PR from `master` to the latest `stable-` branch
-* Merge if the tests pass.
+* Make necessary changes to the code
+* Update the changelog
+* Update the version and release notes in the fsproj file
+* Commit and tag the commit (this is what triggers deployment from  AppVeyor). For consistency, the tag should be identical to the version  (e.g. `1.2.3`).
```
bender2k14: I recommend having the tag be slightly different from the version.  If the version is is `1.2.3`, then the tag would be something like `v/1.2.3`.  That way, tags related to versioning can be grouped together and visually collapsed in some GUI.

As an example, here are some of the lexicographically-first tags in the repository for my project at work (as seen via [SourceTree](https://www.sourcetreeapp.com/)).
Screenshot 2019-09-26 08 32 41

All commits into `develop` are merge commits, and every one of these commits is automatically built, tagged, and includes published artifacts.  Of course, we would use all of these tags to quickly go from running software back to the corresponding commit when something is found that we want to change (like a bug).  Since our project is still so new, we haven't done much of this yet.  Similarly, these are the only type of tags so far (i.e. no manually created tags and no other automatically created tags).

Screenshot 2019-09-26 08 44 36

In the past, my experience is that the product owner wants to tag commits corresponding to any version they presented to others.  It can be a demo to higher ups within their own company or people outside their company.  Those tags might look like `demo/2019-09-26_CEO` if the project owner demoed the application to the CEO of their company on 2019-09-26.  Since it starts with `demo/`, these too are grouped together and collapsable.

Of course these specific use cases probably don't apply to Elmish.WPF, but it is very difficult to rename tags later, so I wanted to share this idea now.

Elmish.WPF.sln
```diff
@@ -3,56 +3,51 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 16
 VisualStudioVersion = 16.0.28803.352
 MinimumVisualStudioVersion = 10.0.40219.1
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".paket", ".paket", "{021C7D6F-30F1-41CC-B8D6-3460EFF300C3}"
-	ProjectSection(SolutionItems) = preProject
-		..\paket.dependencies = ..\paket.dependencies
-	EndProjectSection
+Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "Elmish.WPF", "src\Elmish.WPF\Elmish.WPF.fsproj", "{8C6E8D34-7205-4C57-9722-87E30E4FC5CE}"
```
bender2k14: My personal preference is to have the solution file in the `src` directory.

Is there an advantage to moving it above `src` and into the root of the repository?

src/Elmish.WPF/Elmish.WPF.fsproj
```diff
@@ -15,12 +31,24 @@
     <Compile Include="Utils.fs" />
     <Compile Include="Cmd.fs" />
     <Compile Include="Program.fs" />
-    <None Include="paket.references" />
   </ItemGroup>
+  
+  <ItemGroup>
+    <PackageReference Include="Elmish" Version="[3.0.3, 4)" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta2-19367-01">
```
bender2k14: What is `Microsoft.SourceLink.GitHub` for?  I could find any reference to it in `master`.  Did you mean to add this?

src/Elmish.WPF/Elmish.WPF.fsproj
```diff
@@ -1,9 +1,25 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
bender2k14: I was wondering about this line while working on the multi-targeting.  I suppose Paket uses this to help parse the file.

Elmish.WPF.sln
cmeeren: The advantage is that `dotnet build` works from the repo root without having to specify a sub-directory. So it somewhat simplifies CI and CLI build. Other than that I'm comfortable with having it in `src`. I have no problems moving it back if you prefer that.

src/Elmish.WPF/Elmish.WPF.fsproj
cmeeren: Yes, that was intended. It is strictly speaking a different change, but I don't mind batching them together in this case.

It's for SourceLink support, meaning that when debugging client apps, it's possible to step into Elmish.WPF code.

src/Elmish.WPF/Elmish.WPF.fsproj
cmeeren: I guess Paket adds that just due to the XML parsing implementation it uses – same reason it removes spaces in the project files, I suppose. Those Paket edits always annoy me.

src/Elmish.WPF/Elmish.WPF.fsproj
cmeeren: Re. SourceLink, see [dotnet/sourcelink](https://github.com/dotnet/sourcelink). I've never been able to get it working with symbol packages no matter what I try, so I just consistently embed PDBs in the main package.

Elmish.WPF.sln
bender2k14: The reason I like it in `src` is because I think of it has part of the code in contrast to "DevOps-y" things (like YAML files) that I put in the repository root or in other directories.  In my experience, the technical advantages (such as being able to omit certain build parameters as you mentioned) are small compared to what I think is a somewhat larger conceptual advantage.

So, yea, I would prefer it to remain in `src` if you don't mind.

src/Elmish.WPF/Elmish.WPF.fsproj
bender2k14: Oh, that sounds great!  It is on my list of "technical debt" for my current project at work to get that kind of debugging into our NuGet package dependencies working.

.github/CONTRIBUTING.md
cmeeren: Thanks! I have been very happy with just the version for my other projects, and it's also what's used by other large projects such as [FAKE](https://github.com/fsharp/FAKE/releases) and [Paket](https://github.com/fsprojects/Paket/releases).

That said, I don't have a problem with your suggestion. Consistency is the most important thing.

I've never seen the `v/` prefix though. I have seen `v`, but no slash. Seems tooling-specific whether that will help or not.

I'm slightly inclined to just to with `1.2.3`, but if you want to insist on `v1.2.3` or `v/1.2.3` or something else, we can decide on that instead.

bender2k14: Not insisting.  Whatever you think is best is fine with me.  I just wanted to share my experience in case that would change your mind.

cmeeren: Thanks for your input! Will leave this unresolved and think/research a bit more.

PR #130, opened 2019-09-26
corrected type parameter of getId in OneWaySeqLazyData

bender2k14:
I made a small mistake in #123.  As clearly seen here...
https://github.com/elmish/Elmish.WPF/blob/10b4deb160cbe1808e4d2f68ebd9c772e48d0ecc/src/Elmish.WPF/Binding.fs#L396-L401

...the type of `getId` in `OneWaySeqLazyData` should be `'b -> 'id`, not `'a -> 'id`.


cmeeren:
Thanks!

Issue #131, opened 2019-09-26
How to display a list with duplicate strings


bender2k14:
I want to populate a [`ListBox`](https://docs.microsoft.com/en-us/dotnet/api/system.windows.controls.listbox?view=netframework-4.8) with a list of strings that are _not necessarily distinct_ (as strings).

Here is the binding I used (after I added C#-style name arguments for clarity here even though it doesn't compile)...
```
Binding.oneWaySeq(
  get: (fun m -> m.MyListOfStrings),
  itemEquals: refEq,
  getId: id)
```

...which calls this method:
https://github.com/elmish/Elmish.WPF/blob/10b4deb160cbe1808e4d2f68ebd9c772e48d0ecc/src/Elmish.WPF/Binding.fs#L362-L365

Even though I will have structurally equal strings, I expected this to work because
1. the list of string instances are stored in my model,
2. I used the reference equality function `refeq` to check for the equality of two string instances, and
3. the strings are all constructed at runtime, so they are not reference equal via [string interning](https://docs.microsoft.com/en-us/dotnet/api/system.string.intern?view=netframework-4.8).

Now that my use case has brought my attention here, I find this a bit problematic.  The issue is how instances of type `'a` are checked for equality.  The obvious choice is `itemEquals`, but it is also possible to check for equality after mapping to instances of type `'id`.  The documentation says that `getId` should return a unique ID for each item...
https://github.com/elmish/Elmish.WPF/blob/10b4deb160cbe1808e4d2f68ebd9c772e48d0ecc/src/Elmish.WPF/Binding.fs#L349-L361

...but it doesn't say how those instances of type `'id` are checked for equality.

This come to a head in this code.
https://github.com/elmish/Elmish.WPF/blob/10b4deb160cbe1808e4d2f68ebd9c772e48d0ecc/src/Elmish.WPF/ViewModel.fs#L332-L339

[...]

PR #133, opened 2019-09-28
improved identifier names in the OneWaySeq case of updateValue

bender2k14:
I improved the names in the `OneWaySeq` case of the function `updateValue` that is the focus of #131.  All names were changed using `Ctrl+R, R` in Visual Studio.  No names were shadowing before, and no names are shadowing now.  This is a behavior preserving transformation.

### DIFF
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -329,32 +329,32 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
         else
           currentVal := lazy (newModel |> get |> map)
           true
-    | OneWaySeq (vals, get', map, equals', getId, itemEquals) ->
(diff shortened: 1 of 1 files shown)


cmeeren:
Thanks! IIRC there are one or two other bindings with similar logic. Should those be renamed similarly, or are those different?

bender2k14:
Sure, I can improve those names as well later today.

bender2k14:
I just forced pushed.  The commit now includes some similar name improvements in the `SubModelSeq` case.  I decided to keep using the same PR and thus the same branch name even though its name is specific to the `OneWaySeq` case.  I can create a new branch and PR if you would prefer a better branch name to match the new contents.  I think it is fine as is though.

cmeeren:
Thanks a lot! 👍

Issue #135, opened 2019-09-30
Improve failwith calls in ViewModel by moving them to initialization of bindings


bender2k14:
There are four places in [ViewModel](https://github.com/elmish/Elmish.WPF/blob/master/src/Elmish.WPF/ViewModel.fs) where an exception is thrown.  There of those are for the same reason:

1. In `updateValue`,
https://github.com/elmish/Elmish.WPF/blob/12ad75bd2d82398b11f5b4b7d0d60eb2909921be/src/Elmish.WPF/ViewModel.fs#L486

2. In `TryGetMember`, and
https://github.com/elmish/Elmish.WPF/blob/12ad75bd2d82398b11f5b4b7d0d60eb2909921be/src/Elmish.WPF/ViewModel.fs#L574

3. In `TrySetMember`
https://github.com/elmish/Elmish.WPF/blob/12ad75bd2d82398b11f5b4b7d0d60eb2909921be/src/Elmish.WPF/ViewModel.fs#L599

I think we can improve this by moving these three failure points to one failure point when the [bindings are initialized](https://github.com/elmish/Elmish.WPF/blob/12ad75bd2d82398b11f5b4b7d0d60eb2909921be/src/Elmish.WPF/ViewModel.fs#L303-L311).  When we create the binding for a `SubModelSelectedItem` case, we case locate and store the corresponding `SubModelSeq` binding.  If we can't locate the the binding, then we raise that same exception here.  We would also sort the bindings here to make sure all the `SubModelSeq` bindings are initialized before all the `SubModelSelectedItem` bindings.

...Except it would be more readable if the the type of each binding case were a separately defined record type instead of a tuple.

Is there any downside to refactoring the cases of [`VmBinding<,>`](https://github.com/elmish/Elmish.WPF/blob/12ad75bd2d82398b11f5b4b7d0d60eb2909921be/src/Elmish.WPF/ViewModel.fs#L13-L68) to be like [`BindingData<,>`](https://github.com/elmish/Elmish.WPF/blob/12ad75bd2d82398b11f5b4b7d0d60eb2909921be/src/Elmish.WPF/Binding.fs#L111-L122) where the type of each case is a previously defined record type?

I think the behavior would be identical, and I would find the resulting code more readable.

cmeeren:
I agree. It has been on my mind, but I haven't prioritized it. Feel free to send a PR with the changes you propose! 👍

PR #136, opened 2019-10-01
Records for vm binding cases

bender2k14:
First of two PRs for #135.  This PR is intended to be a behavior preserving transformation, and I think it is.  The code compiles, and all the tests pass for me.  I tried all the samples, and everything worked.

I can squash the commits into one if you would prefer.  I am indifferent this time.

### DIFF
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -48,8 +48,16 @@ type internal Cmd<'model, 'msg> = {
   CanExec: 'model -> bool
 }
 
+type internal SubModel<'model, 'msg, 'bindingModel, 'bindingMsg> = {
+  GetModel: 'model -> 'bindingModel voption
+  GetBindings: unit -> Binding<'bindingModel, 'bindingMsg> list
+  ToMsg: 'bindingMsg -> 'msg
+  Sticky: bool
+  Vm: ViewModel<'bindingModel, 'bindingMsg> voption ref
+}
+
 /// Represents all necessary data used in an active binding.
-type internal VmBinding<'model, 'msg> =
+and internal VmBinding<'model, 'msg> =
   | OneWay of OneWay<'model, obj>
   | OneWayLazy of OneWayLazy<'model, obj, obj>
   | OneWaySeq of OneWaySeq<'model, obj, obj, obj>
@@ -57,12 +65,7 @@ type internal VmBinding<'model, 'msg> =
   | TwoWayValidate of TwoWayValidate<'model, 'msg, obj>
   | Cmd of Cmd<'model, 'msg>
   | CmdParam of cmd: Command
-  | SubModel of
-      vm: ViewModel<obj, obj> voption ref
-      * getModel: ('model -> obj voption)
-      * getBindings: (unit -> Binding<obj, obj> list)
-      * toMsg: (obj -> 'msg)
-      * sticky: bool
+  | SubModel of SubModel<'model, 'msg, obj, obj>
   | SubModelWin of
       vmWinState: WindowState<ViewModel<obj, obj>> ref
       * getState: ('model -> WindowState<obj>)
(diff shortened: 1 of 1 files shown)


src/Elmish.WPF/ViewModel.fs
```diff
@@ -582,13 +638,13 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
...
     | true, binding ->
         match binding with
-        | TwoWay (_, set, dispatch)
-        | TwoWayValidate (_, set, _, dispatch) ->
+        | TwoWay { Set = set; Dispatch = dispatch }
+        | TwoWayValidate { Set = set; Dispatch = dispatch } ->
             set value currentModel |> dispatch
             true
-        | SubModelSelectedItem (_, _, set, name, dispatch) ->
+        | SubModelSelectedItem { Set = set; SubModelSeqBindingName = name; Dispatch = dispatch } ->
```
cmeeren: I'm thinking it might be less verbose/noisy to not destructure like this, but instead consistently (as long as it makes sense) bind the record to a value, throughout the file. What do you think?

src/Elmish.WPF/ViewModel.fs
```diff
@@ -9,63 +9,95 @@ open System.Windows
 
 open Elmish
 
+
+type internal OneWay<'model, 'a> = {
```
cmeeren: We could suffix the type names with `Data`, just like we do in `Binding.fs`. What do you think?

src/Elmish.WPF/ViewModel.fs
bender2k14: Yea, I know what you mean.  Normally I would have done that in this case, but `SubModelSelectedItem` is going to change a lot in the next PR, so I decided to go with the deconstruction here for now since it is easier to see that behavior was preserved.

I think deconstruction is required when cases are combined.  I also deconstructed if only one or two fields were used.  I think of those as the special cases.  In general, I agree with binding the record to an identifier.

Are there any specific deconstructions (other than for `SubModelSelectedItem`) that would like me to remove?

src/Elmish.WPF/ViewModel.fs
bender2k14: I think adding a suffix is a good idea.  I would prefer to keep the names distinct, so I don't want to also suffix these ones with `Data`.  Those records in `Bindings.fs` are the types for cases in the DU called `BindingData`:
https://github.com/elmish/Elmish.WPF/blob/12ad75bd2d82398b11f5b4b7d0d60eb2909921be/src/Elmish.WPF/Binding.fs#L111-L122

How about we suffix the new records with `Binding` since they are the types for cases in the DU called `VmBinding`?
https://github.com/elmish/Elmish.WPF/blob/12ad75bd2d82398b11f5b4b7d0d60eb2909921be/src/Elmish.WPF/ViewModel.fs#L13

src/Elmish.WPF/ViewModel.fs
cmeeren: Personally I'd like all of these record deconstructions to just use e.g. `b` (because `Binding` suffix) and dot notation, but I won't force the issue if you don't like that.

src/Elmish.WPF/ViewModel.fs
cmeeren: 👍

src/Elmish.WPF/ViewModel.fs
bender2k14: One nice thing about deconstructing is that searching on the DU case name in Visual Studio gives more information.  In that penultimate result, I know that only the `Values` field is used.  That helps me in my static analysis.

Screenshot 2019-10-01 08 09 24

cmeeren: Suit yourself :) 👍

src/Elmish.WPF/ViewModel.fs
bender2k14: Done

cmeeren:
Ready to merge if you are.

bender2k14:
Yep, ready :)

PR #138, opened 2019-10-01
SubModelSeq into SubModelSelectedItem

### LINKED ISSUE #135: Improve failwith calls in ViewModel by moving them to initialization of bindings

bender2k14:
Second of two PRs to resolve #135.  It "moves" three failure points "ahead" in the computation.  Instead of _eventually_ throwing an exception in one of three methods, the exception is now thrown "much sooner" in `initializeBindings`.

I removed a test that ensured that an exception is thrown in those three methods.  I did _not_ add a "replacement" test that ensures `initializeBinding` throws an exception in a similar case.

I created an explicit comparison function for sorting binding data.  I considering deriving this ordering from the order in which the cases are defined in the DU.  For example, in Haskell, this would [look like](http://learnyouahaskell.com/making-our-own-types-and-typeclasses) `data Bool = False | True deriving (Ord)`.  However, I wasn't able to determine if such a concept exists in F#.  Maybe DUs in F# automatically have an ordering; I didn't check.  Even so, It doesn't feel like good or idiomatic F# to do something implicit like that.  I like the explicitness of what I did, but I am certainly open to suggestions for improvements.

I added a test to exercise the explicitly defined comparison function.  I tried to use `GenX.auto`, but it doesn't know how to create function types.  I tried to figure out how to help Hedgehog create functions, but I found nothing.  I left and commented out the code to help anyone that wants to try and make that approach work.

I tested the `SubModelSelectedItem` sample.  It works when the `SubModelSelectedItem` binding is defined both before and after the `SubModelSeq` binding.  It also throws an exception if the binding name doesn't match, which goes unhandled and crashes the program.

### DIFF
--- a/src/Elmish.WPF.Tests/BindingTests.fs
+++ b/src/Elmish.WPF.Tests/BindingTests.fs
@@ -3763,3 +3763,33 @@ module subModelSelectedItem =
         let d = Binding.subModelSelectedItem("", (fail: _ -> _ option), set) |> getSubModelSelectedItemData
         test <@ d.Set (p |> Option.map box |> ValueOption.ofOption) m = set p @>
       }
+
+
+
+module sorting =
+
+  [<Fact>]
+    let ``SubModelSelectedItemData sorted last`` () =
+      Property.check <| property {
+        let! s = GenX.auto<string>
+        let! b = GenX.auto<bool>
+        let! vo = GenX.auto<obj voption>
+        let data =
+          [ SubModelSelectedItemData { Get = fail; Set = fail2; SubModelSeqBindingName = s; WrapDispatch = fail }
+            OneWayData { Get = fail }
+            OneWayLazyData { Get = fail; Map = fail; Equals = fail2 }
+            OneWaySeqLazyData { Get = fail; Map = fail; Equals = fail2; GetId = fail; ItemEquals = fail2 }
+            TwoWayData { Get = fail; Set = fail2; WrapDispatch = fail }
+            TwoWayValidateData { Get = fail; Set = fail2; Validate = fail; WrapDispatch = fail }
+            CmdData { Exec = fail; CanExec = fail; WrapDispatch = fail }
+            CmdParamData { Exec = fail2; CanExec = fail2; AutoRequery = b; WrapDispatch = fail }
+            SubModelData { GetModel = fail; GetBindings = fail; ToMsg = fail; Sticky = b }
+            SubModelWinData { GetState = fail; GetBindings = fail; ToMsg = fail; GetWindow = fail2; IsModal = b; OnCloseRequested = vo }
+            SubModelSeqData { GetModels = fail; GetId = fail; GetBindings = fail; ToMsg = fail }
+            SubModelSelectedItemData { Get = fail; Set = fail2; SubModelSeqBindingName = s; WrapDispatch = fail }
+          ]
+        let sorted = data |> List.sortWith BindingData.subModelSelectedItemDataLast
+        match sorted with
+        | [_; _; _; _; _; _; _; _; _; _; SubModelSelectedItemData _; SubModelSelectedItemData _] -> ()
+        | _ -> failwith "SubModelSelectedItemData was not sorted last"
+      }
--- a/src/Elmish.WPF/Binding.fs
+++ b/src/Elmish.WPF/Binding.fs
@@ -18,7 +18,8 @@ module WindowState =
     | WindowState.Hidden m -> WindowState.Hidden (f m)
     | WindowState.Visible m -> WindowState.Visible (f m)
 
-  /// Converts None to WindowState.Closed, and Some(x) to WindowState.Visible(x).
+  /// Converts None to WindowState.Closed, and Some(x) to
+  /// WindowState.Visible(x).
   let ofOption (model: 'model option) =
     match model with
     | Some x -> WindowState.Visible x
@@ -248,8 +249,9 @@ type Binding private () =
 
 
   /// <summary>
-  ///   Creates a one-way binding to an optional value. The binding automatically
-  ///   converts between the optional source value and an unwrapped (possibly
+  ///   Creates a one-way binding to an optional value. The binding
+  ///   automatically converts between the optional source value and an
+  ///   unwrapped (possibly
   ///   <c>null</c>) value on the view side.
   /// </summary>
   /// <param name="get">Gets the value from the model.</param>
@@ -262,8 +264,9 @@ type Binding private () =
 
 
   /// <summary>
-  ///   Creates a one-way binding to an optional value. The binding automatically
-  ///   converts between the optional source value and an unwrapped (possibly
+  ///   Creates a one-way binding to an optional value. The binding
+  ///   automatically converts between the optional source value and an
+  ///   unwrapped (possibly
   ///   <c>null</c>) value on the view side.
   /// </summary>
   /// <param name="get">Gets the value from the model.</param>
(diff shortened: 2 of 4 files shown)


src/Elmish.WPF/ViewModel.fs
```diff
@@ -358,10 +362,11 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
 
   let bindings =
     log "[%s] Initializing bindings" propNameChain
-    let dict = Dictionary<string, VmBinding<'model, 'msg>>()
-    for b in bindings do
+    let dict = Dictionary<string, VmBinding<'model, 'msg>>(bindings.Length)
```
bender2k14: I initialized the dictionary using the [constructor that accepts a capacity](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.-ctor?view=netframework-4.8#System_Collections_Generic_Dictionary_2__ctor_System_Int32_).  This is (slightly) more efficient since we know that number is exactly how many elements will be in the dictionary when we are done adding.  This is unrelated to everything else in this PR.

cmeeren: 👍

src/Elmish.WPF/Binding.fs
```diff
@@ -1686,9 +1698,10 @@ type Binding private () =
   ///
   ///   Only use this if you are unable to use some kind of <c>SelectedValue</c> or
   ///   <c>SelectedIndex</c> property with a normal <see cref="twoWay" /> binding.
-  ///   This binding is less type-safe and will throw at runtime if
-  ///   <paramref name="subModelSeqBindingName" /> does not correspond to a
-  ///   <see cref="subModelSeq" /> binding, or if the inferred <c>'id</c> type
+  ///   This binding is less type-safe.
+  ///   It will throw when initializing the bindings if <paramref name="subModelSeqBindingName" />
+  ///   does not correspond to a <see cref="subModelSeq" /> binding,
+  ///   and it will throw at runtime if if the inferred <c>'id</c> type
```
cmeeren: Would be great to have these lines normalized to ~80 characters (i.e., no line break after "type-safe.") Similar for the other overloads.

src/Elmish.WPF/Binding.fs
```diff
@@ -131,6 +131,18 @@ and Binding<'model, 'msg> =
 
 module internal BindingData =
 
+
+  let subModelSelectedItemDataLast a b =
+    match a, b with
+    | SubModelSelectedItemData _, SubModelSelectedItemData _ -> 0
+    | SubModelSelectedItemData _,                          _ -> 1
+    |                          _, SubModelSelectedItemData _ -> -1
+    |                          _,                          _ -> 0
```
cmeeren: I see why you'd align it like this, but personally I never do because it's unstable if things are renamed. I'd just do this:

```f#
    match a, b with
    | SubModelSelectedItemData _, SubModelSelectedItemData _ -> 0
    | SubModelSelectedItemData _, _ -> 1
    | _, SubModelSelectedItemData _ -> -1
    | _, _ -> 0
```

But if you prefer it aligned, that's fine with me.

src/Elmish.WPF.Tests/BindingTests.fs
```diff
@@ -3763,3 +3763,69 @@ module subModelSelectedItem =
...
+      //      SubModelData subModelData
+      //      SubModelWinData subModelWinData
+      //      SubModelSeqData subModelSeqData
+      //      SubModelSelectedItemData subModelSelectedItemData
+      //    ]
+      //  let sorted = data |> List.sortWith BindingData.subModelSelectedItemDataLast
+      //  match sorted with
+      //  | [_; _; _; _; _; _; _; _; _; _; SubModelSelectedItemData _; SubModelSelectedItemData _] -> ()
+      //  | _ -> failwith "SubModelSelectedItemData was not sorted last"
+      //}
```
cmeeren: Just leaving an unresolved conversation here as a reminder.

bender2k14:
> > Maybe DUs in F# automatically have an ordering
> 
> Indeed they do.

Not in this case.  What follows is the compiler error that results when trying to compare two.

> Error FS0001 The type `Binding<'model,'msg>` does not support the `comparison` constraint because it is a record, union or struct with one or more structural element types which do not support the `comparison` constraint.

It seems like the functions to compare are automatically implemented if they can be.  That is a neat feature.

src/Elmish.WPF/Binding.fs
bender2k14: Sure, removed the white space.  I only have a mild preference for having things aligned.

src/Elmish.WPF/Binding.fs
bender2k14: Sure.  Auto formatted the comments in that whole file.

src/Elmish.WPF.Tests/BindingTests.fs
bender2k14: I removed the commented out code, used `fail` and `fail2` in place of the functions I had, and used `GenX.atuo` to create `bool`, `string`, and `obj voption` instances.

bender2k14:
Before we merge, I would prefer to squash some of these commits

src/Elmish.WPF/Binding.fs
cmeeren: Brilliant. Are you using the Rewrap VS extension?

bender2k14: I used the extension for VS Code.  That is the first one I found.  I got the impression that one also exists for VS, but I didn't look into it.

cmeeren:
> Before we merge, I would prefer to squash some of these commits

Sure, let me know when you're ready. 👍

bender2k14:
Ready

cmeeren:
Thanks a lot!

PR #144, opened 2019-10-06
Unused values as warnings

bender2k14:
First commit modifies `fsproj` files to turn unused values into build warnings.  The second commit deletes one unused line of test code.  The third commit renames all unused values to `_`.

### DIFF
--- a/src/Elmish.WPF.Tests/BindingTests.fs
+++ b/src/Elmish.WPF.Tests/BindingTests.fs
@@ -120,7 +120,7 @@ module oneWayOpt =
       Property.check <| property {
         let! x = GenX.auto<int>
 
-        let get m = None
+        let get _ = None
         let d = Binding.oneWayOpt(get) |> getOneWayData
 
         test <@ isNull (d.Get x) @>
@@ -601,7 +601,7 @@ module twoWayOpt =
       Property.check <| property {
         let! x = GenX.auto<int>
 
-        let get m = None
+        let get _ = None
         let d = Binding.twoWayOpt(get, fail2) |> getTwoWayData
 
         test <@ isNull (d.Get x) @>
@@ -663,7 +663,7 @@ module twoWayOpt =
       Property.check <| property {
         let! x = GenX.auto<int>
 
-        let get m = ValueNone
+        let get _ = ValueNone
         let d = Binding.twoWayOpt(get, fail2) |> getTwoWayData
 
         test <@ isNull (d.Get x) @>
@@ -725,7 +725,7 @@ module twoWayOpt =
       Property.check <| property {
         let! x = GenX.auto<int>
 
-        let get m = None
+        let get _ = None
         let d = Binding.twoWayOpt(get, (fail: _ option -> int)) |> getTwoWayData
 
         test <@ isNull (d.Get x) @>
@@ -787,7 +787,7 @@ module twoWayOpt =
       Property.check <| property {
         let! x = GenX.auto<int>
 
-        let get m = ValueNone
+        let get _ = ValueNone
         let d = Binding.twoWayOpt(get, (fail: _ voption -> int)) |> getTwoWayData
 
         test <@ isNull (d.Get x) @>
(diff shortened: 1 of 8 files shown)


bender2k14:
> ...I have often found it equally (if not more) nice to know at a glance what a parameter actually is, even if unused, and names help with this. For example, it becomes easier to make use of a parameter later when needed.

I think code should be optimized for reading since that is the typical case.  If someone is going to write new code in that area, then they can look at the type to quickly determine "what a parameter actually is".

In this case of `CmdParam`, the type of the parameter is just `obj`.  For lambda function, the convention I see (both in this project and beyond) is to have the identify be the first letter of the corresponding type.  So for this case, I agree that replacing `p` with `_` makes things significantly less clear when _not_ reading.  However, I think the fix is to give a proper type to the parameter.  See #146.

bender2k14:
> In any case, given that I don't think it's generally a given which option to prefer, I have never turned on `--warnon:1182`. (This is also combined with the fact that I usually always turn on `<WarningsAsErrors />`, forcing me to fix warnings.)

I agree that `<WarningsAsErrors />` is a good idea, but sadly it is currently broken :(
https://github.com/dotnet/core-sdk/issues/1708

bender2k14:
> For example, in the samples, I think I generally prefer names...

A simple compromise would be to only add `--warnon:1182` in non-sample projects.

src/Samples/UiBoundCmdParam/Program.fs
```diff
@@ -26,7 +26,7 @@ let bindings () : Binding<Model, Msg> list = [
   "Numbers" |> Binding.oneWay(fun m -> m.Numbers)
   "Limit" |> Binding.twoWay((fun m -> float m.EnabledMaxLimit), int >> SetLimit)
   "Command" |> Binding.cmdParamIf(
-    (fun p m -> Command),
+    (fun _ _ -> Command),
```
cmeeren: I'd like this reverted.

My reasoning is that the names of even unused parameters in bindings can make signatures/intent more clear to newcomers (who are the primary target audience for the samples).

src/Samples/UiBoundCmdParam/UiBoundCmdParam.fsproj
```diff
@@ -4,6 +4,8 @@
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp3.0</TargetFramework>
     <UseWpf>true</UseWpf>
+    <!--Turn on warnings for unused values (arguments and let bindings) -->
+    <OtherFlags>$(OtherFlags) --warnon:1182</OtherFlags>
```
cmeeren: I'd like all of these reverted in samples

cmeeren:
You make good sense. I can agree that code should be optimized for reading. I can accept the changes to the main project.

I can also accept most of the changes to the samples. Apart from the `--warnon:1182` you mentioned,   there are actually just a single change I'd really like reverted. I've left a comment.

Also, instead of `<OtherFlags>$(OtherFlags) --warnon:1182</OtherFlags>` I *think* there may be a `WarnOn` element of sorts that can be used to add codes more directly (though I can't find it right now, so I may be mistaken). Could you have a quick look? (Don't spend your entire evening on it.)

src/Samples/UiBoundCmdParam/UiBoundCmdParam.fsproj
bender2k14: Done

src/Samples/UiBoundCmdParam/Program.fs
bender2k14: Sure.  Done.

bender2k14:
I would prefer to squash some commits before merging

cmeeren:
Let me know when you're ready!

bender2k14:
> Also, instead of `<OtherFlags>$(OtherFlags) --warnon:1182</OtherFlags>` I think there may be a `WarnOn` element of sorts that can be used to add codes more directly (though I can't find it right now, so I may be mistaken). Could you have a quick look? (Don't spend your entire evening on it.)

This is the only way that I know how to do it in the project file.

I have found this part of devops rather confusing.  On the one had, there is very clear information about what [options `fsc`, the F# compiler, accepts](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/compiler-options).  On the other hand, us developers call commands like `dotnet build` that have their own default arguments, extract arguments from the project files, and eventually call the F# compiler or one of the C# compilers.  What seems to be missing is documentation specifying all of the translations from `dotnet build` to `fsc`.

bender2k14:
Ready :)

cmeeren:
Thanks a lot!

PR #145, opened 2019-10-06
List to seq in UpdateModel

bender2k14:
https://github.com/elmish/Elmish.WPF/blob/4c8d85c58a6adbd7db1158be4fb3e9525f7be45c/src/Elmish.WPF/ViewModel.fs#L580-L591

I think the reason that `propsToNotify` and `cmdsToNotify` are `list`s instead of `seq`s is so that they are eagerly computed before the mutation on line 589.  However, I don't see any reason why `bindings` needs to be converted to a `list` ASAP (via lines 582 and 587).  In particular, after all bindings are initialized (in the same scope where the dictionary was instantiated), the `bindings` dictionary is only ever read (i.e. it is never mutated).  I made this clear by exposing the `bindings` dictionary as read only.

Therefore, I think it is a behavior preserving transformation to delay the calls to `Seq.toList` until the end of those expressions.

### DIFF
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -369,7 +369,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
       let binding = initializeBinding b.Name b.Data dict
       dict.Add(b.Name, binding)
       setInitialError b.Name binding
-    dict
+    dict :> IReadOnlyDictionary<string, VmBinding<'model, 'msg>>
 
   let getSelectedSubModel model vms getSelectedId getSubModelId =
       vms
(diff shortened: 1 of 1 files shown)


cmeeren:
Thanks!

cmeeren:
By the way:

> `slight Seq for List optimization in UpdateModel`

Unless you have measured, it's not guaranteed to actually be an optimization. The choice between `list`, `array`, and `seq` (or other collection types) and when to convert between them can be difficult to grasp. CPU cache locality, optimizations, JIT stuff etc. can all impact this, even for cases where one option seems like the optimal choice. [BenchmarkDotNet](https://github.com/dotnet/BenchmarkDotNet) is a great tool to get to the bottom of small performance optimizations like these.

PR #146, opened 2019-10-08
added a cast function for CmdParam binding

bender2k14:
Adds a `cast` function to `CmdParam` so that the parameter can be of type `'param` instead of type `obj`.

### DIFF
--- a/src/Elmish.WPF.Tests/BindingTests.fs
+++ b/src/Elmish.WPF.Tests/BindingTests.fs
@@ -2342,7 +2342,7 @@ module cmdParam =
     let ``sets the correct binding name`` () =
       Property.check <| property {
         let! bindingName = GenX.auto<string>
-        let binding = bindingName |> Binding.cmdParam(fail2)
+        let binding = bindingName |> Binding.cmdParam(fail, fail2)
         test <@ binding.Name = bindingName @>
       }
 
@@ -2354,7 +2354,7 @@ module cmdParam =
         let! p = GenX.auto<string>
 
         let exec (p: obj) (m: int) = unbox p + string m
-        let d = Binding.cmdParam(exec) |> getCmdParamData
+        let d = Binding.cmdParam(fail, exec) |> getCmdParamData
 
         test <@ d.Exec (box p) m = (exec p m |> ValueSome) @>
       }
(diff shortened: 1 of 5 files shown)


cmeeren:
Isn't this a breaking change?

Could you give some usage examples? I fail to see how this is an improvement over `|> unbox<'param'> |> `.

bender2k14:
Yes, it is a breaking change.  I created it to aid the discussion of #144.  In general, I think this is an improvement (since otherwise both `exec` and `canExec` would generally contain duplicate casting code).  However, the primary reason I wrote it though was to cite it from https://github.com/elmish/Elmish.WPF/pull/144#issuecomment-539605573.

cmeeren:
I agree that the typing can be improved. However, I think there's a better way to do it (written from the top of my head):

```f#
  static member cmdParam<'param>
      (exec: 'param -> 'model -> 'msg,
       ?wrapDispatch: Dispatch<'msg> -> Dispatch<'msg>)
    CmdParamData {
      Exec = fun p model -> exec (unbox<'param> p) model |> ValueSome
      CanExec = fun _ _ -> true
      AutoRequery = false
      WrapDispatch = defaultArg wrapDispatch id
    } |> createBinding
```

Usage:

```f#
Foo |> Binding.cmdParam<int>(fun p m -> SetInt p)
```

Or, in fact, since F# can infer it in this case (assuming `SetInt` is `int -> ...`):


```f#
Foo |> Binding.cmdParam(fun p m -> SetInt p)
```

I don't even think this is a breaking change, because `'param` will likely be inferred to be `obj` for existing usages using `unbox`.

Thank you for bringing this to my attention!

bender2k14:
From the perspective of the code in this PR, you are (in part) suggesting that we hard code `cast: obj -> 'param` to `unbox<'param>`.

I see trouble with that.  Looking at the `UiBoundCmdParam` sample, it seems like the parameter can be `null`:
https://github.com/elmish/Elmish.WPF/blob/2f7fc9f155b8af9dc2c735f8b63145f9f441d0d5/src/Samples/UiBoundCmdParam/Program.fs#L28-L31

Calling (say) `unbox<int> null` throws a `NullReferenceException`.

Maybe if hard coding `cast` to `Option.ofObj |> Option.map unbox<'param>` would work?  I tried and got confusing compiler errors.

Honestly though, I don't care about making this change.  I vote we abandon this PR.  We can create an issue about this change if it is important enough.

cmeeren:
> Honestly though, I don't care about making this change. I vote we abandon this PR. We can create an issue about this change if it is important enough.

Let's leave it as it is, then. I guess that `cmdParam` and `cmdParamIf` aren't used nearly as much as `cmd` and `cmdIf`. If anyone cares enough about this to come up with a good way to improve the API, I'm open to suggestions in a new issue.

Issue #137, opened 2019-10-01
Optimizations using IDs


bender2k14:
This issue spun out of #131.  Here is a brief summary.  In the `OneWaySeq` and `SubModelSeq` cases of `updateValue`, the runtime can be decreased from quadratic to nearly linear by making extensive use of the ID that each element in the sequence has.

Progress on this issue should wait until after #134 is fixed.

The hardest part to optimize is when a constant fraction of all items are removed.  For example, suppose all items in the first half of the sequence are removed.  Calling either [`Remove`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.collection-1.remove?view=netframework-4.8#System_Collections_ObjectModel_Collection_1_Remove__0_) or [`RemoveAt`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.collection-1.removeat?view=netframework-4.8#System_Collections_ObjectModel_Collection_1_RemoveAt_System_Int32_) takes linear time for each removal (regardless of the order), so the whole computation will take quadratic time.

To achieve nearly linear time in this case, we would have to derive from `ObservableCollection<>` and expose a method similar to [`List<T>.RemoveAll(Predicate<T>)`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.removeall?view=netframework-4.8).  I assumed that someone else would have done this, but I haven't found anything so far.

bender2k14:
Oh, `Move` is also troublesome.  I think each [call](https://github.com/dotnet/corefx/blob/d2506266dffcf6d83af11d192fa87435048614cb/src/System.ObjectModel/src/System/Collections/ObjectModel/ObservableCollection.cs#L161-L172) takes (in the worst case) linear time.

> I assumed that someone else would have done this, but I haven't found anything so far.

I was using the wrong search terms.  Searches with "observablecollection" and any of "range", "bulk", or "batch" have great results, including [StackOverflow](https://stackoverflow.com/questions/670577/observablecollection-doesnt-support-addrange-method-so-i-get-notified-for-each/45364074#45364074) and [dotnet/corefx's GitHub](https://github.com/dotnet/corefx/issues/10752).

Everyone seems to be concerned with raising multiple events about changes to individual elements vs raising one event about many element changes.  I think the use case there is that those events are being raised on background thread.

That is not the case for Elmish.WPF.  I was just stepping through the code in detail today for an unrelated reason, and this is what I observed.
[...]

Issue #113, opened 2019-09-12
Best practices for mutable objects in model?


reinux:
I'm working with a text/code editor control called [AvalonEdit](http://avalonedit.net/), which has a `TextDocument` class that contains all the text as well as methods for manipulating the text and events for changes to the text/filename/etc.

I thought about just wrapping it in a control along with the editor, and binding to a collection of lines, but it turns out there's no way to two-way bind to a collection, which is sensible.

So now I'm putting the `TextDocument` inside the model, which means my model is now mutable. I have a few questions about this:

- What are the caveats to having a mutable object in the model?
- How do I subscribe to events coming from an object in the model?
- Is this the right approach overall? 

Thanks in advance.

reinux:
Yeah, that seems to be the best approach. Haven't run into any horrible roadblocks yet.

Will post here if I notice anything that might be worth mentioning.

Thanks for all the input!

JDiLenarda:
@reinux : can we have a feedback, please ? I’m interested to know how you dealt with it.

bender2k14:
The general feedback would be not to do it.

Why are you considering using mutuality?

JDiLenarda:
I genuinely don’t get if it’s a technical opinion or if I’m told I broke some etiquette. If so, I apologize.

bender2k14:
Oh, sorry.  I thought you were requesting general feedback about having the model contain mutable data.  I was stating a technical opinion.  To be clear, you are welcome to ask any question.  We will do our best to answer it.

@reinux, do you have an update that you can share about your use of mutability in the model?

JDiLenarda:
So you mean mutability, not mutuality ? This word lost me. No worry, though.

More specifically, I’d like to know how reinux dealt with its use of a third-party component.

bender2k14:
Oh, yes.  I corrected my "mutuality" for "mutability" typo.  Thanks :)

reinux:
@JDiLenarda I think I've stumbled on a few repeatable patterns for factoring mutable state out of the model and dealing with non-MVVM stuff in general. @bender2k14 would love to have your input to know if I'm on the right track.

The editor and the document live happily in XAML now, and there are no direct references to it in the model.

The general idea is that there are only a handful of reasons to interact with a mutable object, and only a handful of ways to deal with them:
- You need to instantiate it. Do it in XAML if possible, bind to it if possible, but otherwise instantiate after the window constructor and use `Program.withSubscription`.
[...]

PR #149, opened 2019-10-12
Simple observable collection optimizations

### LINKED ISSUE #134: Use of ObservableCollection<T>.Remove(T)
As first stated in https://github.com/elmish/Elmish.WPF/issues/131#issuecomment-536340393, I think the use of [`ObservableCollection<T>.Remove(T)`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.collection-1.remove?view=netframework-4.8) (in three calls in `updateValue`) is a bug.  Instead of removing by item, I think the correct implementation is to remove by index via [`ObservableCollection<T>.RemoveAt(Int32)`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.collection-1.removeat?view=netframework-4.8).

This issue will serve as the tracking of this (potential) problem (instead of #131...that issue has enough going on in it :P ).

Here is the relevant part of that previous comment.

------------

Recall the update logic for the `OneWaySeq` case of `updateValue`.
https://github.com/elmish/Elmish.WPF/blob/8f78f833bef5c903e6947b23688cef30480ccf8b/src/Elmish.WPF/ViewModel.fs#L341-L343

I think that call to `Remove` contains a bug.  The precise thing to do is remove by index.  Removing by item means that the collection needs to find which item to remove.  The documentation for `ObservableCollection<T>.Remove` [says](https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.collection-1.remove?view=netframework-4.8#remarks) it removes the first item it finds for which that item and the given item return `true` when passed into the default equality comparer of type `T`.  If `itemEquals` is not the same as this default equality comparer, then this call to `Remove` will not have the intended behavior.  I can create a test that exhibits this problem and a PR with a fix.

bender2k14:
Fixes #134 and partially addresses #137 by implementing all of the "simple" optimizations.

I have not tested anything yet.  I will test later.

The two remaining more complicated optimizations are removing all elements that need to be removed in nearly linear time (as [discussed here](https://github.com/elmish/Elmish.WPF/issues/137#issue-500916222)) and ordering in nearly linear time (as [discussed here](https://github.com/elmish/Elmish.WPF/issues/137#issuecomment-537744117)).

### DIFF
--- a/src/Elmish.WPF.Tests/ViewModelTests.fs
+++ b/src/Elmish.WPF.Tests/ViewModelTests.fs
@@ -3,6 +3,7 @@
 open System
 open System.Collections.Concurrent
 open System.Collections.ObjectModel
+open System.Collections.Specialized
 open System.ComponentModel
 open System.Windows.Input
 open FSharp.Interop.Dynamic
@@ -29,7 +30,7 @@ type internal TestVm<'model, 'msg>(model, bindings) as this =
 
   let pcTriggers = ConcurrentDictionary<string, int>()
   let ecTriggers = ConcurrentDictionary<string, int>()
-  let ccTriggers = ConcurrentDictionary<string, int>()
+  let ccTriggers = ConcurrentDictionary<string, NotifyCollectionChangedEventArgs list>()
   let cecTriggers = ConcurrentDictionary<string, int>()
   let dispatchMsgs = ResizeArray<'msg> ()
 
@@ -55,7 +56,7 @@ type internal TestVm<'model, 'msg>(model, bindings) as this =
     ecTriggers.TryGetValue propName |> snd
 
   member __.NumCcTriggersFor propName =
-    ccTriggers.TryGetValue propName |> snd
+    ccTriggers.GetOrAdd(propName, []).Length
 
   member __.NumCecTriggersFor propName =
     cecTriggers.TryGetValue propName |> snd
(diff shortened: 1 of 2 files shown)


cmeeren:
Thanks a lot! Let me know when you're ready to merge.

@reinux do you want to give this PR a spin to check how much it impacts the performance you mentioned in https://github.com/elmish/Elmish.WPF/issues/113#issuecomment-541227271?

reinux:
Works great!
Down from about 250ms per update to 120ms.
Haven't set up any proper tests, but it seems to work as expected.

cmeeren:
@reinux @bender2k14 Do you think the dictionary stuff in https://github.com/elmish/Elmish.WPF/issues/137#issuecomment-541247319 could work well in this PR?

reinux:
The `Seq.find` and the `Move` call that follows are actually what causes the bug that I mentioned in that comment. As @bender2k14 notes, caching the ID->ixs of the current items up front causes it to be out of sync as the `Move` calls change the indices. His branch actually uses the dictionary for the rest of the function, where that isn't an issue.

bender2k14:
Indeed.  Of all the planning that I did for this PR, I didn't notice that replacing `Seq.fine` with dictionary lookups would cause a bug.  Good thing @reinux ran into it before me ;) :P

bender2k14:
I added a commit that calls `Clear` when it would be efficient to do so

cmeeren:
Seems that commit broke some tests. Could you investigate?

bender2k14:
Oh, I had an improvement to that commit that I forgot to include.  I force pushed a new commit.  The tests all pass for me.

The previous commit had the correct behavior, but raised more (clear) events than it needed to.  It is great to see that there are some tests that noticed the difference.

bender2k14:
> Thanks a lot

My pleasure :)

> Let me know when you're ready to merge.

I am ready.  I have reviewed my changes several times and have done some manual testing.  I think everything is working.

> PS, you can mark PRs as "drafts" or something like that when you create them, to signal if you don't want them merged right away.

Oh, great.  I will keep that in mind for next time.

cmeeren:
In https://github.com/elmish/Elmish.WPF/issues/134#issuecomment-541492334 you mention adding a test; should that be part of this PR or should I go ahead and merge?

bender2k14:
It makes sense to add the test to this PR.  Coming soon!

bender2k14:
It took me longer than I expected to write the test.  The good news though is that I think it is complete.  I have it failing on `master`.  Tomorrow I will confirm that it passes on the breach in this PR.  Then I will force push the branch with the test created first.

bender2k14:
Done! :D

Ready to merge

cmeeren:
Thanks a lot! Will try to review and merge tomorrow.

src/Elmish.WPF.Tests/ViewModelTests.fs
```diff
@@ -818,6 +815,87 @@ module OneWaySeqLazy =
...
+  type TestClass (id: int, data: string) =
+    member _.Id = id
+    member _.Data = data
+    override __.GetHashCode() = 0
+    override __.Equals that =
+      // All instances of TestClass are considered equal.
+      // Not very helpful, but a valid implementation.
+      match that with
+      | null -> false
+      | :? TestClass as x -> true
```
cmeeren: Actually, you can remove the `null` check. Type checks à la `:? TestClass` will only succeed if it's not null.

bender2k14: Sure.  Force pushed that change.

PR #156, opened 2019-10-16
improves the error handling when getId returns duplicates

### LINKED ISSUE #150: Binding.subModelWin stops Cmd.OfFunc.perform from working
Hello! This is a bit of an odd one, which might be due to some misunderstanding of how Elmish works, but here goes:

I have an app with three windows: a main window that is the entry point, and two sub windows that get opened or closed based on events. These are both controlled via code like:

```
"TableOfContents" |> Binding.subModelWin (
        (fun m -> m.tableOfContents |> WindowState.ofOption), 
        snd, 
        TableOfContentsMessage,
        TableOfContents.bindings,
        (Codex.Views.TableOfContents))
```

I also have save file functionality, triggered like so:

```
| SaveNovel -> 
        model, Cmd.OfFunc.perform saveFile "" id
```

With the saveFile function adapted from another issue answer here:

```
let saveFile fileName =
    Application.Current.Dispatcher.Invoke(fun () ->
        let dialog = SaveFileDialog (Filter = fileFilter, FileName = fileName)
        let result = dialog.ShowDialog ()
        if result.HasValue && result.Value
            then SaveFileSelected dialog.FileName
            else FileSelectCanceled
    )
```

Now my problem: the save functionality works fine when no other windows are open, but the moment I show a sub window, the result of `Cmd.OfFunc.perform saveFile "" id` is ignored. If I close the subwindow, the save functionality starts working again.

`Cmd.OfFunc.perform` creates a `[bind]` list, with bind being a func to run. It seems this list is just ignored if a subwindow is open, as if I just returned `Cmd.none`. 

Any ideas? The code is on github but private. I can open it up if that would help.

bender2k14:
Fixes #150 

I am very open to suggestions for improvement on this code.  My primary goal was to quickly give us something concrete to discuss.

I didn't try to deduplicate anything, and I am not jazzed about how I used `if-then` to skip over the happy path in `updateValue`.  However, I think this behavior is ideal.  Otherwise, no other bindings are updated after encountering this issue.

### DIFF
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -200,12 +200,12 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
   let initializeBinding name bindingData (initializedBindingsByName: Dictionary<string, VmBinding<'model, 'msg>>) =
     match bindingData with
     | OneWayData d ->
-        OneWay {
+        Some <| OneWay {
           Get = measure name "get" d.Get }
     | OneWayLazyData d ->
         let get = measure name "get" d.Get
         let map = measure name "map" d.Map
-        OneWayLazy {
+        Some <| OneWayLazy {
           Get = get
           Map = map
           Equals = measure2 name "equals" d.Equals
@@ -221,7 +221,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
           if valuesById.ContainsKey id
           then logInvalidGetId id (valuesById.[id]) value
           else valuesById.Add(id, value)
-        OneWaySeq {
+        Some <| OneWaySeq {
           Get = get
           Map = map
           Equals = measure2 name "equals" d.Equals
@@ -229,12 +229,12 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
           ItemEquals = measure2 name "itemEquals" d.ItemEquals
           Values = values }
     | TwoWayData d ->
-        TwoWay {
+        Some <| TwoWay {
           Get = measure name "get" d.Get
           Set = measure2 name "set" d.Set
           Dispatch = d.WrapDispatch dispatch }
     | TwoWayValidateData d ->
-        TwoWayValidate {
+        Some <| TwoWayValidate {
           Get = measure name "get" d.Get
           Set = measure2 name "set" d.Set
           Validate = measure name "validate" d.Validate
@@ -245,7 +245,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
         let dispatch' = d.WrapDispatch dispatch
         let execute _ = exec currentModel |> ValueOption.iter dispatch'
         let canExecute _ = canExec currentModel
-        Cmd {
+        Some <| Cmd {
           Cmd = Command(execute, canExecute, false)
           CanExec = canExec }
     | CmdParamData d ->
@@ -254,14 +254,14 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
         let dispatch' = d.WrapDispatch dispatch
         let execute param = exec param currentModel |> ValueOption.iter dispatch'
         let canExecute param = canExec param currentModel
-        CmdParam <| Command(execute, canExecute, d.AutoRequery)
+        Some <| CmdParam (Command(execute, canExecute, d.AutoRequery))
     | SubModelData d ->
         let getModel = measure name "getSubModel" d.GetModel
         let getBindings = measure name "bindings" d.GetBindings
         let toMsg = measure name "toMsg" d.ToMsg
         match getModel initialModel with
         | ValueNone ->
-            SubModel {
+            Some <| SubModel {
               GetModel = getModel
               GetBindings = getBindings
               ToMsg = toMsg
@@ -270,7 +270,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
         | ValueSome m ->
             let chain = getPropChainFor name
             let vm = ViewModel(m, toMsg >> dispatch, getBindings (), config, chain)
-            SubModel {
+            Some <| SubModel {
               GetModel = getModel
               GetBindings = getBindings
               ToMsg = toMsg
@@ -283,7 +283,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
         let onCloseRequested = fun () -> d.OnCloseRequested |> ValueOption.iter dispatch
         match getState initialModel with
         | WindowState.Closed ->
-            SubModelWin {
+            Some <| SubModelWin {
               GetState = getState
               GetBindings = getBindings
               ToMsg = toMsg
@@ -303,7 +303,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
             showNewWindow
               winRef d.GetWindow vm d.IsModal onCloseRequested
               preventClose Visibility.Hidden
-            SubModelWin {
+            Some <| SubModelWin {
               GetState = getState
               GetBindings = getBindings
               ToMsg = toMsg
@@ -323,7 +323,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
             showNewWindow
               winRef d.GetWindow vm d.IsModal onCloseRequested
               preventClose Visibility.Visible
-            SubModelWin {
+            Some <| SubModelWin {
               GetState = getState
               GetBindings = getBindings
               ToMsg = toMsg
@@ -352,7 +352,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
           if modelsById.ContainsKey id
           then logInvalidGetId id (modelsById.[id]) model
           else modelsById.Add(id, model)
-        SubModelSeq {
+        Some <| SubModelSeq {
           GetModels = getModels
           GetId = getId
           GetBindings = getBindings
@@ -361,13 +361,15 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
     | SubModelSelectedItemData d ->
         match initializedBindingsByName.TryGetValue d.SubModelSeqBindingName with
         | true, SubModelSeq b ->
-          SubModelSelectedItem {
+          Some <| SubModelSelectedItem {
             Get = measure name "get" d.Get
             Set = measure2 name "set" d.Set
             SubModelSeqBinding = b
             Dispatch = d.WrapDispatch dispatch
             Selected = ref ValueNone }
-        | _ -> failwithf "subModelSelectedItem binding referenced binding '%s', but no compatible binding was found with that name" d.SubModelSeqBindingName
+        | _ ->
+          log "subModelSelectedItem binding referenced binding '%s', but no compatible binding was found with that name" d.SubModelSeqBindingName
+          None
 
   let setInitialError name = function
     | TwoWayValidate { Validate = validate } ->
@@ -381,10 +383,13 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
     let dict = Dictionary<string, VmBinding<'model, 'msg>>(bindings.Length)
     let sortedBindings = bindings |> List.sortWith BindingData.subModelSelectedItemLast
     for b in sortedBindings do
-      if dict.ContainsKey b.Name then failwithf "Binding name '%s' is duplicated" b.Name
-      let binding = initializeBinding b.Name b.Data dict
-      dict.Add(b.Name, binding)
-      setInitialError b.Name binding
+      if dict.ContainsKey b.Name then
+        log "Binding name '%s' is duplicated. Only the first occurance will be used." b.Name
+      else
+        initializeBinding b.Name b.Data dict
+        |> Option.iter (fun binding ->
+          dict.Add(b.Name, binding)
+          setInitialError b.Name binding)
     dict :> IReadOnlyDictionary<string, VmBinding<'model, 'msg>>
 
   let getSelectedSubModel model vms getSelectedId getSubModelId =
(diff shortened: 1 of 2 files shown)


cmeeren:
The diff's quite large; could you briefly explain your thought process?

bender2k14:
> The diff's quite large...

Maybe you have GitHub set to display whitespace changes?  I recommend hiding those by default.  It can also be helpful to consider a slit view.  I find that the unified view allows me to better for single line changes while the split view helps me better understand multiline changes.
Screenshot 2019-10-16 12 38 43

> ...could you briefly explain your thought process?

Sure.  This PR shows how Elmish.WPF can log and provided much more information than Elmish can when it catches the exception and logs to its `onError` function.

If two elements/models return the same thing from `getId`, then
- if the bindings are being initialized, then an exception is thrown;
- if the view model is being updated in `updateValue`, then the issue is logged, the happy path is skipped, and `false` is returned (instead of the previous behavior of throwing an exception).

src/Elmish.WPF/ViewModel.fs
```diff
@@ -336,6 +344,12 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
                ViewModel(m, (fun msg -> toMsg (getId m, msg) |> dispatch), getBindings (), config, chain)
           )
           |> ObservableCollection
+        let modlesById = Dictionary<_,_>(vms.Count)
+        for model in vms |> Seq.map (fun vm -> vm.CurrentModel) do
+          let id = getId model
+          if modlesById.ContainsKey id
+          then failwithf "The getId function is suppose to return distinct IDs for each sub-model, but it returned the same ID %A for the two sub-models %A and %A" id (modlesById.[id]) model
```
ChrisPritchard: This should be 'supposed' not 'suppose'. Missing a 'd' :)

Also if the ids are the same, then printing those two ids will not be all that useful. Perhaps a better message would be:

"The getId function should return a unique id for all sub-models, but has returned the id '%A' for more than one item"

bender2k14: > This should be 'supposed' not 'suppose'. Missing a 'd' :)

Thanks!  I force pushed the correction.

> Also if the ids are the same, then printing those two ids will not be all that useful.

That is not what is printed.  It prints three things each once:
1. an ID returned from `getId`,
2. an input to `getId` that returns this ID, and
3. a different input to `getId` that returns this same ID.

ChrisPritchard: Ah yep I missed the intention - I was just thinking from my use case, where I used the `id` function (as does the samples); will this not return the same text three times?

src/Elmish.WPF/ViewModel.fs
```diff
@@ -211,13 +211,21 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
     | OneWaySeqLazyData d ->
         let get = measure name "get" d.Get
         let map = measure name "map" d.Map
+        let getId = measure name "getId" d.GetId
+        let values = ObservableCollection(initialModel |> get |> map)
+        let valuesById = Dictionary<_,_>(values.Count)
+        for value in values do
+          let id = getId value
+          if valuesById.ContainsKey id
+          then failwithf "The getId function is supposed to return distinct IDs for each element, but it returned the same ID %A for the two elements %A and %A" id (valuesById.[id]) value
```
cmeeren: I suggest changing `is supposed to` to `must` in all messages.

src/Elmish.WPF/ViewModel.fs
```diff
@@ -529,43 +553,53 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
         let newSubModels = newModel |> b.GetModels |> Seq.toList
```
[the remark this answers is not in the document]
bender2k14: Sure.  Maybe a better argument for using an array is that the memory allocation happens sequentially, so the caching performance is better.

bender2k14:
I am fine with all the change suggestions.

Before I proceed, I just want to make sure that we prefer the type of behavior generally exhibited by code in this branch.  I don't think it is unreasonable to stick with the current behavior for now given that I determined that the exception is caught in the Emlish dispatch loop and logged to its `onError` function.

I am fine either way.

src/Elmish.WPF/ViewModel.fs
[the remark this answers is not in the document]
cmeeren: That, too. Though I don't know nearly enough about the performance implications when it comes to CPU caches.

cmeeren:
> Before I proceed, I just want to make sure that we prefer the type of behavior generally exhibited by code in this branch. I don't think it is unreasonable to stick with the current behavior for now given that I determined that the exception is caught in the Emlish dispatch loop and logged to its `onError` function.

Are you referring to `log` vs `failwith`?

bender2k14:
> Are you referring to my log vs failwith question?

No, I was referring to the PR as a whole.

Let me try phrasing it this way.

When the user makes a mistake and provides a `getId` function that returns duplicate IDs and we "encounter" some duplicate IDs,
- are you satisfied with the current behavior, which is that [`Dictionary<,>.Add`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.add?view=netframework-4.8) throws an `ArgumentNullException` that is caught in the Elmish dispatch loop and logged to its `onError` function, or
- would you prefer that Elmish.WPF catch this exception (as in this PR) and somehow (such as via `log`) inform the user with an error message (as in this PR) that is informative (certainly more informative than what is currently being logged by Elmish) so that the user can more quickly realize and correct their mistake?

Quoting https://github.com/elmish/Elmish.WPF/issues/150#issuecomment-541533861:
> Reopening this because I'd like to know why the exception is just swallowed.

I am trying to be clear and explicitly with the way I phrased my question above because I started creating this PR before I knew where the exception was going.  I am just double checking that we really want this change now that we know what the current behavior is.

I am definitely in favor of the behavior in this PR.  The main downside that I see is that the code in this PR is more complicated, especially because I tried to quickly write something that I knew would work.

cmeeren:
I most certainly like the behaviour in this PR better than the current one.

src/Elmish.WPF/ViewModel.fs
```diff
@@ -336,6 +344,12 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
                ViewModel(m, (fun msg -> toMsg (getId m, msg) |> dispatch), getBindings (), config, chain)
           )
           |> ObservableCollection
+        let modlesById = Dictionary<_,_>(vms.Count)
```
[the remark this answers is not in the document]
bender2k14: done

src/Elmish.WPF/ViewModel.fs
```diff
@@ -395,43 +409,53 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
...
-          newValIdxPairsById
-          |> Seq.filter (Kvp.key >> oldValIdxPairsById.ContainsKey >> not)
-          |> Seq.map Kvp.value
-          |> Seq.map snd
-          |> Seq.iter b.Values.Add
+          for (newIdx, newVal) in newVals |> Seq.indexed do
+            let id = b.GetId newVal
+            if newValIdxPairsById.ContainsKey id
+            then log "The getId function is supposed to return distinct IDs for each element, but it returned the same ID %A for the two elements %A and %A" id (newValIdxPairsById.[id]) newVal
+            else  newValIdxPairsById.Add(id, (newIdx, newVal))
```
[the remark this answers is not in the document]
bender2k14: done

src/Elmish.WPF/ViewModel.fs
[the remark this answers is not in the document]
bender2k14: Done there on line 553 and also on line 409

src/Elmish.WPF/ViewModel.fs
```diff
@@ -395,43 +409,53 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
           let newVals = newModel |> b.Get |> b.Map |> Seq.toList
```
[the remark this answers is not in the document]
bender2k14: Done (same commit as above)

src/Elmish.WPF/ViewModel.fs
bender2k14: done

bender2k14:
> [What is] the difference between `failwith` at initialization, and `log` elsewhere?

`updateValue` (aka "elsewhere") was throwing a non-custom (exception and) message that was caught by the Elmish dispatch loop and logged to its `onError` function.  Even if @ChrisPritchard had noticed Elmish logging that exception, it still would have lacked clear and specific information about what went wrong.  Therefore, logging the issue to the Elmish.WPF log (with clear and specific information) is the only way to improve the situation in `updateValue`.

The reason I decided to use `failwithf` instead of `log` in `initializeBinding` is because `failwithf` was used elsewhere when initializing bindings:
https://github.com/elmish/Elmish.WPF/blob/2f7fc9f155b8af9dc2c735f8b63145f9f441d0d5/src/Elmish.WPF/ViewModel.fs#L363-L372

and
https://github.com/elmish/Elmish.WPF/blob/2f7fc9f155b8af9dc2c735f8b63145f9f441d0d5/src/Elmish.WPF/ViewModel.fs#L345-L354

In the second case, that call to `failwithf` used to be in `TryGetMember` and `TrySetMember` until I moved it to this location in PR #138.  Looking through that again, I think it always could have been a call to `log` instead of throwing an exception.  Maybe that code was written before Elmish.WPF had a logger.

> Bindings may be initialized at any time during the app's lifetime, so I can't immediately see why we should not use the same failing/logging both places.

If either of those calls to `failwithf` are executed at startup, then the application will crash.  If they are executed after startup, then I think the exception will be caught by the Elmish dispatch loop (and logged to its `onError` function).  The advantage here over the case in `updateValue` is that the exception message clearly states what went wrong.

I think a more consistent and lenient behavior is to replace both of those calls to `failwithf` with calls to `log` (as well as some additional logic to ignore those bindings).  I will make this change.  Please review.

bender2k14:
> I will make this change. Please review.

I deleted a test that checked for a thrown exception when duplicate binding names are used.  I would prefer to have a test for the new behavior, but I don't see an easy way to test it.

bender2k14:
> ...it would be great if there is a way to reduce whatever duplication there is.

Sure.  I removed some duplication (in https://github.com/elmish/Elmish.WPF/pull/156/commits/e4ac2e3ca7a707ab97b2b788d147f62acb24d412).

cmeeren:
> `updateValue` (aka "elsewhere")

😂

> > Bindings may be initialized at any time during the app's lifetime, so I can't immediately see why we should not use the same failing/logging both places.
> 
> If either of those calls to `failwithf` are executed at startup, then the application will crash. If they are executed after startup, then I think the exception will be caught by the Elmish dispatch loop (and logged to its `onError` function).

No, exceptions thrown during binding initialization will bubble up anyway. To see this, modify the SubModelSeq example to have a duplicated binding in `counterBindings`, and remove the initial counter (i.e., set `DummyRoot = [] |> asDummyRoot` in `init`). When you add a counter after startup, you'll get an exception.

> I think a more consistent and lenient behavior is to replace both of those calls to `failwithf` with calls to `log` (as well as some additional logic to ignore those bindings). I will make this change. Please review.

Having duplicated bindings is a critical error. Devs might not be watching the output log closely (partiularly since it may be very noisy), or even have enabled logging in Elmish.WPF for that matter (which, after all, is opt-in). I think binding initialization errors should be properly "in your face", i.e. throw an exception, just as invalid arguments do in most APIs. Given that they will indeed be thrown also after app startup (contrary to your previous assumption), perhaps you agree?

bender2k14:
> > If either of those calls to `failwithf` are executed at startup, then the application will crash.  If they are executed after startup, then I think the exception will be caught by the Elmish dispatch loop (and logged to its `onError` function).  The advantage here over the case in `updateValue` is that the exception message clearly states what went wrong.
> 
> No, exceptions thrown during binding initialization will bubble up anyway. ... When you add a counter after startup, you'll get an exception...

What part are you disagreeing with?  Bubble up to where?  My soft claim there was that they bubble up to the Elmish dispatch loop.

I justed tested now.  I upgrade my soft claim to a hard claim by replacing "...I think..." with "...it is definitely the case that...".  In the `SumModelSeq` sample, add the line
```
|> Program.withErrorHandler (fun (s, e) -> printf "%s\n%A" s e)
```

src/Elmish.WPF/ViewModel.fs
```diff
@@ -381,10 +383,13 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
     let dict = Dictionary<string, VmBinding<'model, 'msg>>(bindings.Length)
     let sortedBindings = bindings |> List.sortWith BindingData.subModelSelectedItemLast
     for b in sortedBindings do
-      if dict.ContainsKey b.Name then failwithf "Binding name '%s' is duplicated" b.Name
-      let binding = initializeBinding b.Name b.Data dict
-      dict.Add(b.Name, binding)
-      setInitialError b.Name binding
+      if dict.ContainsKey b.Name
+      then log "Binding name '%s' is duplicated. Only the first occurance will be used." b.Name
```
cmeeren: Could you move `then` up to the previous line here?

```f#
if ... then
  log ...
```

cmeeren:
> Bubble up to where?

I was wrong.

I meant that it bubbles up to the debugger, crashing the program (which I think is good for the reasons mentioned). However, that only happens with a project reference to Elmish.WPF (which is then debugged, too). I tried with a NuGet reference to Elmish.WPF, and then there's no exception in the debugger, only in the output.

I left a single nitpicky comment; otherwise I'm happy with your changes.

src/Elmish.WPF/ViewModel.fs
bender2k14: done

bender2k14:
I squashed the commits.  I am ready to merge as well.

cmeeren:
Thanks a lot!

Issue #143, opened 2019-10-04
Lazy subModelSeq (and subModel)?


cmeeren:
Unless I'm mistaken, currently `subModelSeq` will loop through and update the sub-bindings for every item on every iteration of the update loop. This may have performance implications for large sequences. If the input sequence has not changed, the bindings should ideally be skipped altogether. This calls for a `subModelSeqLazy` binding.

The same might be said for `subModel`, though there's only ever one model there, so it's not as critical. But given that the model might have further sub-models (or just a lot of bindings), it should be possible to skip `subModel`, too, if the input hasn't changed.

Nobody has requested this, and I don't have any immediate plans on implementing it, but I'm leaving this issue here in case anyone else wants to look at it, or just as a future reminder for myself.

bender2k14:
Yep!  I want this now.

> Unless I'm mistaken, currently subModelSeq will loop through and update the sub-bindings for every item on every iteration of the update loop. This may have performance implications for large sequences.

Yep, you are correct.  We noticed the performance of our application decrease significantly after I added the `SubModelSeq` binding from https://github.com/elmish/Elmish.WPF/issues/131#issuecomment-538398038 even though it only contained around 200 sub-models.  The CPU analyzer in Visual Studio showed that the binding functions were taking a significant percentage of the time.  Upgrading to a version with the `SubModelSeq` optimizations from #149 regained about half of the lost performance while removing all sub-models or removing the entire `SubModelSeq` binding regained all of the lost performance.

Just like how we have `Binding.oneWaySeq` as the special case of `Binding.oneWaySeqLazy` with `equals = refEq` and `map = id` (as of #157), I think we should also use `refEq` and `id` as the defaults in these new lazy bindings for `SubModel[Seq]`.

Quoting https://github.com/elmish/Elmish.WPF/issues/157#issuecomment-543021860
> I'm not sure whether this should bump the patch (implementation detail), minor (new feature), or major (breaking change) version. I'm leaning towards the latter, but am open to other opinions.

Again, I am indifferent to which version bump you go with.  However, I recommend that we implement these lazy bindings while using `refEq` and `id` as the defaults and release them in the same version bump since it is the same kind of change.  In particular, then you only have to do that version bump once.

cmeeren:
I agree with everything. Thanks! 👍 

[...]

PR #158, opened 2019-10-17
optimize default OneWaySeq using refEq and id

### LINKED ISSUE #157: OneWaySeq[Lazy] with equals = refEq and map = id
This is my second attempt to improve the lazy bindings.  (The first was in #117.)  I have more experience with Elmish.WPF now, and I am more confident this time that what follows makes sense :)

I now believe that I see the great value in the `OneWaySeqLazy` binding (instead of just the `OneWaySeq` binding).  Every message results in a new model, and every new model is passed to `ViewModel<'model, 'msg>.Update`.  Within that method, each binding is inspected to see if its data needs to be updated.  If so, then the new data is passed to WPF through the appropriate changed events.

For a `OneWaySeq[Lazy]` binding, the appropriate changed notifications are raised by mutating an `ObservableCollection<>`.  Checking if the data needs to be updated in a collection like this can take a long time (though it now takes less time thanks to #149).  This is why `OneWaySeqLazy` exists; if `equals` returns `true`, then the user is telling us that we can skip doing all the work needed to check if any element involved in the `OneWaySeq` binding has changed.

https://github.com/elmish/Elmish.WPF/blob/4b4151b8ffcc97fdbcb8647acb1f337e301caf74/src/Elmish.WPF/Binding.fs#L419-L425

I was thinking about how to define `equals` and `map`.  A good choice seems to be
- `equals = refEq` and
- `map = id`.

If `get newModel` returns something that is reference equal to `get oldModel`, then clearly there are no changed notifications that need to be raised for the corresponding binding.   ..._Unless_ `get` returned a mutable data structure.  However, the user shouldn't be mutating, so this shouldn't happen.

As such, I suggest that we modify `Binding.oneWaySeq`, which is currently...
https://github.com/elmish/Elmish.WPF/blob/4b4151b8ffcc97fdbcb8647acb1f337e301caf74/src/Elmish.WPF/Binding.fs#L382-L393

...so that `equals = refEq` and `map = id`.  That way, the default/simplest way to create a `OneWaySeq` binding is also very efficient for the vast majority of cases (which is all the cases that are immutable).  If the user decides to play with the fire known as mutability, then they can create their binding using `Binding.oneWaySeqLazy` and pass in `equal = fun _ _ -> false` (like the current implementation of `Binding.oneWaySeq` does) or something else that makes sense in their case.

bender2k14:
Fixes #157 

> I get no compile-time error using your code. Note that you need to move `Utils.fs` up above `Bindings.fs` to make `refEq` accessible.

Yep, I did that move, but this commit still doesn't build for me :(

Let's see if it the automated build likes it.

### DIFF
--- a/src/Elmish.WPF/Binding.fs
+++ b/src/Elmish.WPF/Binding.fs
@@ -370,7 +370,9 @@ type Binding private () =
   ///   binding is backed by a persistent <c>ObservableCollection</c>, so only
   ///   changed items (as determined by <paramref name="itemEquals" />) will be
   ///   replaced. If the items are complex and you want them updated instead of
-  ///   replaced, consider using <see cref="subModelSeq" />.
+  ///   replaced, consider using <see cref="subModelSeq" />. This method is
+  ///   implemented by calling <see cref="oneWaySeqLazy" /> with <c>refEq</c>
+  ///   for <c>equals</c> and <c>id</c> for <c>map</c>.
   /// </summary>
   /// <param name="get">Gets the collection from the model.</param>
   /// <param name="itemEquals">
@@ -384,13 +386,7 @@ type Binding private () =
        itemEquals: 'a -> 'a -> bool,
        getId: 'a -> 'id)
       : string -> Binding<'model, 'msg> =
-    OneWaySeqLazyData {
-      Get = box
-      Map = unbox<'model> >> get >> Seq.map box
-      Equals = fun _ _ -> false
-      GetId = unbox<'a> >> getId >> box
-      ItemEquals = fun a b -> itemEquals (unbox<'a> a) (unbox<'a> b)
-    } |> createBinding
+    Binding.oneWaySeqLazy(get, refEq, id, itemEquals, getId)
 
 
   /// <summary>
--- a/src/Elmish.WPF/Elmish.WPF.fsproj
+++ b/src/Elmish.WPF/Elmish.WPF.fsproj
@@ -27,11 +27,11 @@
     <Compile Include="AssemblyInfo.fs" />
     <Compile Include="InternalUtils.fs" />
     <Compile Include="InternalTypes.fs" />
+    <Compile Include="Utils.fs" />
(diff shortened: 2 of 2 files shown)


bender2k14:
The automated build doesn't like my commit either.

@cmeeren, when you said that my code worked for you, you must have done something a bit different.  Did you make changes to the tests?  That is where I get three compile errors.

bender2k14:
Read https://github.com/elmish/Elmish.WPF/issues/143#issuecomment-543446567 before deciding to release after completing this PR.

cmeeren:
Quoting from https://github.com/elmish/Elmish.WPF/issues/143#issuecomment-543904477:

> I suggest we wait with merging #158 until this is ready, too. Alternatively, if it makes the job easier for you, we can create a dedicated branch in this repo for these changes, and merge toward that instead, and when it's ready, merge that into master.

I'll have a look at the compilation errors.

cmeeren:
The "error" is already present in the bindings as warnings (an example of why I always enable warnings as errors).

In any case, the way the types are defined, I don't really see a way to have `Map` be `id`. The closest I come is this:

```f#
static member oneWaySeq
      (get: 'model -> #seq<'a>,
       itemEquals: 'a -> 'a -> bool,
       getId: 'a -> 'id)
  OneWaySeqLazyData {
      Get = unbox<'model> >> get >> box
      Map = unbox<seq<'a>> >> Seq.map box
      Equals = refEq
      GetId = unbox<'a> >> getId >> box
      ItemEquals = fun x y -> itemEquals (unbox<'a> x) (unbox<'a> y)
    } |> createBinding
```

If you play around with it and closely inspect the parameter types, the types used by the `OneWaySeqLazyData` type and the way the `OneWaySeqLazyData` DU case requires some of the parameters to be `obj`, I'm sure you can convince yourself that it needs to be this way (or discover why I'm wrong).

There may of course be other changes you can do to make it work.

bender2k14:
Replaced by PR #162.

PR #159, opened 2019-10-18
upgraded Microsoft.Xaml.Behaviors.Wpf to version 1.1.3

bender2k14:
Upgrades [Microsoft.Xaml.Behaviors.Wpf](https://www.nuget.org/packages/Microsoft.Xaml.Behaviors.Wpf/) to its latest version.  Compared to the previous version that we were using that only targeted .NET Framework, this version also targets .NET Core.  This removes two build warnings for us since we target both of those now.

I tested the `EventBindingsAndBehaviors` sample, which is the only code that depends on this NuGet package.  Everything still works.


cmeeren:
Excellent, thanks!

PR #160, opened 2019-10-18
simplified type check syntax

bender2k14:
This syntax is strictly better:
1. It is shorter.
2. It is idiomatic.
3. It doesn't have a warning.  The previous code had a warning because `x` was unused (and we are using `--warnon:1182`) and [`x` can't be replaced by `_`](https://github.com/dotnet/fsharp/issues/7740).


cmeeren:
Thanks! I didn't look that hard at the test code in the PR that added this originally. 😅

PR #162, opened 2019-10-21
[Internal] optimize default OneWaySeq using refEq

bender2k14:
Replaces PR #158.  Will merge many similar changes into this internal branch as describe in https://github.com/elmish/Elmish.WPF/issues/143#issuecomment-543904477.

### DIFF
--- a/src/Elmish.WPF/Binding.fs
+++ b/src/Elmish.WPF/Binding.fs
@@ -367,42 +367,12 @@ type Binding private () =
   /// <summary>
   ///   Creates a one-way binding to a sequence of items, each uniquely
   ///   identified by the value returned by <paramref name="getId"/>. The
-  ///   binding is backed by a persistent <c>ObservableCollection</c>, so only
-  ///   changed items (as determined by <paramref name="itemEquals" />) will be
-  ///   replaced. If the items are complex and you want them updated instead of
-  ///   replaced, consider using <see cref="subModelSeq" />.
-  /// </summary>
-  /// <param name="get">Gets the collection from the model.</param>
-  /// <param name="itemEquals">
-  ///   Indicates whether two collection items are equal. Good candidates are
-  ///   <c>elmEq</c>, <c>refEq</c>, or simply <c>(=)</c>.
-  /// </param>
-  /// <param name="getId">Gets a unique identifier for a collection
-  /// item.</param>
-  static member oneWaySeq
-      (get: 'model -> #seq<'a>,
-       itemEquals: 'a -> 'a -> bool,
-       getId: 'a -> 'id)
-      : string -> Binding<'model, 'msg> =
-    OneWaySeqLazyData {
-      Get = box
-      Map = unbox<'model> >> get >> Seq.map box
-      Equals = fun _ _ -> false
-      GetId = unbox<'a> >> getId >> box
-      ItemEquals = fun a b -> itemEquals (unbox<'a> a) (unbox<'a> b)
-    } |> createBinding
-
-
-  /// <summary>
-  ///   Creates a one-way binding to a sequence of items, each uniquely
-  ///   identified by the value returned by <paramref name="getId"/>. The
-  ///   binding will only be updated if the output of <paramref name="get" />
-  ///   changes, as determined by <paramref name="equals" />. The binding is
-  ///   backed by a persistent
-  ///   <c>ObservableCollection</c>, so only changed items (as determined by
-  ///   <paramref name="itemEquals" />) will be replaced. If the items are
-  ///   complex and you want them updated instead of replaced, consider using
-  ///   <see cref="subModelSeq" />.
+  ///   binding will not be updated if the output of <paramref name="get"/>
+  ///   does not change, as determined by <paramref name="equals"/>.
+  ///   The binding is backed by a persistent <c>ObservableCollection</c>, so
+  ///   only changed items (as determined by <paramref name="itemEquals"/>)
+  ///   will be replaced. If the items are complex and you want them updated
+  ///   instead of replaced, consider using <see cref="subModelSeq"/>.
   /// </summary>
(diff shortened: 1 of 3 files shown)


bender2k14:
Continuing conversation from https://github.com/elmish/Elmish.WPF/pull/158#issuecomment-543932328
> In any case, the way the types are defined, I don't really see a way to have `Map` be `id`.
> ...
> If you play around with it..., then I'm sure you can convince yourself that it needs to be this way (or discover why I'm wrong).
>
> There may of course be other changes you can do to make it work.

There isn't a compile error if we have F# infer the type parameters to a function instead of a method.

But now I see some tests fail.  I will look into that later.

bender2k14:
> But now I see some tests fail. I will look into that later.

There were three tests that failed because they were checking for behavior specific to the previous default behavior.  I replaced/modified those those tests.  The new tests now check for behavior specific to the new default behavior.

I think this PR is ready to be merged.

src/Elmish.WPF/Binding.fs
```diff
@@ -232,6 +232,21 @@ module internal Helpers =
     { Name = binding.Name
       Data = BindingData.box binding.Data }
 
+  let createOneWaySeqLazyData
+    (get: 'model -> 'a)
+    (equals: 'a -> 'a -> bool)
+    (map: 'a -> #seq<'b>)
+    (itemEquals: 'b -> 'b -> bool)
+    (getId: 'b -> 'id) =
+      let x = OneWaySeqLazyData {
```
cmeeren: Style-wise, I normally follow the rule that the function body should be indented one level (2 spaces), and the parameters, if appearing on separate lines such as this, should be indented three levels (one level further than the function name):

```f#
let createOneWaySeqLazyData
      (get: 'model -> 'a)
      (equals: 'a -> 'a -> bool)
      (map: 'a -> #seq<'b>)
      (itemEquals: 'b -> 'b -> bool)
      (getId: 'b -> 'id) =
  let x = OneWaySeqLazyData {
```

bender2k14:
> I'm happy you found a workaround.

I figured out another work around.  Instead of defining a function that both methods call (which, of course, is defined before both methods), it suffices to define the more general method first.  I think this reveals the fundamental issue.  If these were functions, then instead working as intended (like in `master`) or a compiler warning (like in my previous commit), we would have a compiler error.  This behavior with methods is not what I expected.

I would prefer this workaround (i.e. switching the order of `Binding.oneWaySeq` and `Binding.oneWaySeqLazy`).

Do you have a preference @cmeeren?

cmeeren:
> This behavior with methods is not what I expected.

Me neither.

As I understand your comment, your preference is to have `oneWaySeq` call `oneWaySeqLazy`. That is my preference, too.

bender2k14:
I found an existing (and closed) issue about this behavior.
https://github.com/dotnet/fsharp/issues/5787

src/Elmish.WPF/Binding.fs
```diff
@@ -232,6 +232,21 @@ module internal Helpers =
...
+        Get = get >> box
+        Map = unbox<'a> >> map >> Seq.map box
+        Equals = fun x y -> equals (unbox<'a> x) (unbox<'a> y)
+        GetId = unbox<'b> >> getId >> box
+        ItemEquals = fun x y -> itemEquals (unbox<'b> x) (unbox<'b> y)
+      }
+      x |> createBinding
```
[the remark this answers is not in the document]
bender2k14: Oh, sure.  That was left over from me trying various things.  Thanks for finding that.

src/Elmish.WPF/Binding.fs
```diff
@@ -432,6 +403,31 @@ type Binding private () =
...
+  ///   <c>elmEq</c>, <c>refEq</c>, or simply <c>(=)</c>.
+  /// </param>
+  /// <param name="getId">Gets a unique identifier for a collection
+  /// item.</param>
+  static member oneWaySeq
+      (get: 'model -> #seq<'a>,
+       itemEquals: 'a -> 'a -> bool,
+       getId: 'a -> 'id)
+      : string -> Binding<'model, 'msg> =
+    Binding.oneWaySeqLazy(get, refEq, id, itemEquals, getId)
```
bender2k14: > As I understand your comment, your preference is to have `Binding.oneWaySeq` call `Binding.oneWaySeqLazy`. That is my preference, too.

Yep.  In particular by defining `Binding.oneWaySeqLazy` before `Binding.oneWaySeq` like this.

I will leave this for you to resolve to ensure that we are on the same page here.

src/Elmish.WPF/Binding.fs
```diff
@@ -370,7 +385,9 @@ type Binding private () =
   ///   binding is backed by a persistent <c>ObservableCollection</c>, so only
   ///   changed items (as determined by <paramref name="itemEquals" />) will be
   ///   replaced. If the items are complex and you want them updated instead of
-  ///   replaced, consider using <see cref="subModelSeq" />.
+  ///   replaced, consider using <see cref="subModelSeq" />. This method is
+  ///   implemented by calling <see cref="oneWaySeqLazy" /> with <c>refEq</c>
+  ///   for <c>equals</c> and <c>id</c> for <c>map</c>.
```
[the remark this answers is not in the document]
bender2k14: Ah, great wording.  I modified the documentation for both methods.  Please check.

src/Elmish.WPF/Binding.fs
bender2k14: Ok.  I will keep that in mind for future reference.  It doesn't matter anymore here because I replace the define-an-addition-function workaround with the reorder-the-methods workaround.

bender2k14:
I would like to squash some commits before merging.

cmeeren:
Looks good. Let me know when you're ready.

bender2k14:
Ready

cmeeren:
Thanks a lot!

PR #164, opened 2019-10-25
now using single instead of double underscore per new F# 4.7 feature

bender2k14:
One of the [new features of F# 4.7](https://devblogs.microsoft.com/dotnet/announcing-f-4-7/) is the ability to use a single underscore as an (ignored) instance identifier in method definitions.  This PR replaces all of our double underscores with single underscores.

### DIFF
--- a/src/Elmish.WPF.Tests/ViewModelTests.fs
+++ b/src/Elmish.WPF.Tests/ViewModelTests.fs
@@ -46,25 +46,25 @@ type internal TestVm<'model, 'msg>(model, bindings) as this =
 
   new(model, binding) = TestVm(model, [binding])
 
-  member private __.Dispatch x =
+  member private _.Dispatch x =
     dispatchMsgs.Add x
 
-  member __.NumPcTriggersFor propName =
+  member _.NumPcTriggersFor propName =
     pcTriggers.TryGetValue propName |> snd
 
-  member __.NumEcTriggersFor propName =
+  member _.NumEcTriggersFor propName =
     ecTriggers.TryGetValue propName |> snd
 
-  member __.NumCcTriggersFor propName =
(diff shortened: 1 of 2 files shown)


bender2k14:
It builds on my machine, but the automated build didn't like it :(

bender2k14:
Wait...what?  Now it doesn't build on my machine. :/

I would still prefer to get this working...in both places.

bender2k14:
> # Get started
> First, install either:
> 
> - The latest .NET Core 3.0
> - The latest Visual Studio
> If you are a Visual Studio user, you will get an appropriate .NET Core installed by default. Once you have installed either .NET Core or Visual Studio 2019, you can use F# 4.7...

Huh?  I have both of those.  I will try to get F# 4.7 working on my system later.

inosik:
The `private` modifier breaks something. This works:

``` fsharp
namespace fslib
type T () =
  member _.N = 0
```

This doesn't:

``` fsharp
namespace fslib
type T () =
  member private _.N = 0
```

cmeeren:
Sounds like this should be reported as an F# bug (if it isn't already).

inosik:
Looks like this was fixed in dotnet/fsharp#7631. However, that was after SDK 3.0.100 was released.

bender2k14:
Thank you @inosik.  I thought I was going crazy.

In hindsight, I should have tried to use another feature of 4.7.  That would have gotten me closer to realizing that this behavior is a bug.

cmeeren:
Should I close this, or should I leave it around for later (meaning it might have to be rebased/fixed etc.)?

bender2k14:
Let's close this.  I will keep a record in my notes to make this change when the bug fix is available.

PR #167, opened 2019-10-29
Example usage of PassEventArgsToCommand

bender2k14:
Partially addresses #33.

Only changes the `EventBindingsAndBehaviors` sample.  I tested it, and everything works.


cmeeren:
Excellent, thanks! What do you mean by "partially"? (Moreover, do you think that issue should be re-opened?)

bender2k14:
> What do you mean by "partially"?

#33 has one OP and one title but many related but separate problems have been discussed and resolved there.  I was just trying to say that this PR only addresses one of them.

> Moreover, do you think that issue should be re-opened?

I suppose it could have been previously open because of the problem resolved by this PR.  I don't know of any remaining problems in that issue, so I think it can remain closed.

Issue #151, opened 2019-10-13
Implement the strategy design pattern for VmBinding


bender2k14:
Writing the code in #149 was more difficult than it would have been if we had stronger types in `ViewModel<'model, 'msg>`.  Nearly all the types in all the bindings are boxed to `obj` before they get used by library authors in `ViewModel<'model, 'msg>`.  Replacing one thing of (compile-time) type `obj` for another satisfies the compiler but would (typically) throw an exception at runtime.

I think the way forward to to have a view model type for every binding type.  For example, the view model for `SubModelSeqBinding` would be `SubModelSeqViewModel<'model, 'msg, 'bindingModel, 'bindingMsg, 'id>`.  These view models still need to expose a common set of behaviors, and there is common code that still needs somewhere to live.  Having `SubModelSeqViewModel<'model, 'msg, 'bindingModel, 'bindingMsg, 'id>` extend `ViewModel<'model, 'msg>` would help to minimize the change while also exposing the common set of methods and giving a home some common behavior.

There are many ways to group things.  I often see things grouped either syntactically or semantically.  Currently, I would describe the view model code as being grouped syntactically.  For example, all of the different cases of `updateValue` are located next to each other in the source code (within a pattern match on the binding type).  I am suggesting that we group the code semantically.  For example, `SubModelSeqViewModel<'model, 'msg, 'bindingModel, 'bindingMsg, 'id>` would implement `updateValue` and other methods like `TryGetMember` and `TrySetMember` by just having the code that is currently contained in the `SubModelSeq` cases of the respective pattern matches.

Speaking more technically, the branching currently expressed through pattern matching would be pushed "down" and performed by the runtime when it goes through the virtual method table while trying to determine which implementation to execute among the types in the resulting inheritance hierarchy.  I first encountered this idea in the Google Clean Code talk titled [Inheritance, Polymorphism, & Testing](https://www.youtube.com/watch?v=4F72VULWFvc).

This change would be invisible to users.  The main disadvantage I see is that pattern match cases that currently share an implemention would either have to repeat/duplicate the implementation or extract the behavior to some function that they all call.  However, I don't think this matters much because there are very few cases that share an implementation, and when they do, the implementations are very short.

BentTranberg:
[...]

PR #169, opened 2019-11-01
combined validation error set and update logic

bender2k14:
I found these two pattern matches on `VmBinding<'model, 'msg>` that were almost identical.  This PR extracts the common code.

The main difference was the model (`initialModel` vs `currentModel`), which is now passed in to the new function.

The other difference was what to do when `ValueNone` is matched.  The update case called `removeError name` while the initial case did nothing.  I unified this by having the initial case now also call `removeError name` since `removeError` will also do nothing after not finding `name` in the `errors` dictionary.

In addition to the extracted function being called, this adds one call to `Dictionary.Remove` and one `if` check per `twoWayValidate` initialization.  Probably fine.

I am looking ahead to #151.  I expect this change to help a bit since this decreases the number of pattern matches on `VmBinding<'model, 'msg>`.

### DIFF
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -152,6 +152,13 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
       log "[%s] ErrorsChanged \"%s\"" propNameChain propName
       errorsChanged.Trigger([| box this; box <| DataErrorsChangedEventArgs propName |])
 
+  let updateValidationError model name = function
+    | TwoWayValidate { Validate = validate } ->
+        match validate model with
+        | ValueNone -> removeError name
+        | ValueSome error -> setError error name
+    | _ -> ()
+
   let measure name callName f =
     if not config.Measure then f
     else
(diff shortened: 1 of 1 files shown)


src/Elmish.WPF/ViewModel.fs
```diff
@@ -152,6 +152,16 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
...
 
+  let updateValidationError name model = function
+    | TwoWayValidate { Validate = validate } ->
+        match validate model with
+        | ValueNone -> removeError name
+        | ValueSome error -> setError error name
+    | _ -> ()
+
+  let setInitialValidationError name binding = updateValidationError name initialModel binding
+  let updateValidationError name binding = updateValidationError name currentModel binding
```
cmeeren: Are these necessary? Since they're only used once each, I think it's more clear if they are removed.

bender2k14: Nope.  I removed them and swapped the order of the `model` and `name` arguments.  I merged this change with the previous commit and force pushed.

cmeeren:
Thanks! Go ahead and merge whenever you're ready.

bender2k14:
> The main difference was the model (`initialModel` vs `currentModel`), which is now passed in to the new function.

In fact, it suffices to always use `currentModel` since `currentModel` is immediately initiated with `initialModel` and there is no mutation yet.  I don't mind leaving it as is though.

cmeeren:
> In fact, it suffices to always use `currentModel` since `currentModel` is immediately initiated with `initialModel` and there is no mutation yet. I don't mind leaving it as is though.

Ah, noted. I don't care much either way. Feel free to change it if you want, or leave it as it is.

PR #171, opened 2019-11-02
Extracts all pattern matches on VmBindings

bender2k14:
Preparing for #143.

After adding a recursive lazy binding case in `VmBinding<'model, 'msg>`, the logic that handles the different cases will also change to be recursive.  This PR extracts all of those pattern matches into their own functions.  I also removed the default case on one pattern match to make sure that I don't forget to add the new lazy case when the time comes.

The functions that will so be recursive are
1. `updateValidationError`
2. `updateValue`
3. `getCmdIfCanExecChanged`
4. `tryGetMember`
5. `trySetMember`


cmeeren:
Looks good, thanks!

PR #172, opened 2019-11-02
Strengthen types in boxBindingData

bender2k14:
Strengthens the types in `boxBindingData` from `obj` to either `'bModel` or `'bMsg`, which are short for "boxed model" and "boxed message" respectively.  This alone is a nice improvement.

Furthermore, this seems necessary for how I have started to implement lazy bindings (c.f. #143).  My current plan is to have the lazy binding data look like
```
type internal LazyData<'model, 'msg, 'bindingModel> = {
  Get: 'model -> 'bindingModel
  Equals: 'bindingModel -> 'bindingModel -> bool
  Binding: Binding<'bindingModel, 'msg>
}
```

When creating the lazy binding, `bindingModel` needs to be boxed but `'msg` should not be boxed.  Thus, that sub binding needs to be boxed by passing in `unbox` for `unboxModel` but `id` for `unboxMsg` and `boxMsg`.

### DIFF
--- a/src/Elmish.WPF/Binding.fs
+++ b/src/Elmish.WPF/Binding.fs
@@ -144,79 +144,91 @@ module internal BindingData =
     subModelSelectedItemDataLast a b
 
 
-  let boxDispatch (strongDispatch: Dispatch<'msg>) : Dispatch<obj> =
-    unbox<'msg> >> strongDispatch
+  let boxDispatch
+      (unboxMsg: 'boxedMsg -> 'msg)
+      (strongDispatch: Dispatch<'msg>) : Dispatch<'boxedMsg> =
+    unboxMsg >> strongDispatch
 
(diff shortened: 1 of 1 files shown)


cmeeren:
I suggest to change `bModel` and `bMsg` to `boxedModel` and `boxedMsg` to make it more self-documenting.

Other than that I can't see anything obviously wrong here. If everything still works, feel free to merge (after considering my request above). :)

bender2k14:
> > ...to either `'bModel` or `'bMsg`, which are short for "boxed model" and "boxed message" respectively.
>
> I suggest changing `bModel` and `bMsg` to `boxedModel` and `boxedMsg` to make it more self-documenting; this stuff is dense enough as it is.

Oh, of course.  This is why PRs / code reviews are so useful.  It is so much easier to see the "mistakes" in other's code than in your own code 😛 Thanks!

Issue #110, opened 2019-09-08
Nested voption in type of selected VM in SubModelSelectedItem


bender2k14:
I just noticed that the `selected` property in the case `SubModelSelectedItem` of the discriminated union `VmBinding<'model, 'msg>` contains nested `voption`s.

https://github.com/elmish/Elmish.WPF/blob/c0ec0d3a9a605054895d10e8ee0980cddbd44ed0/src/Elmish.WPF/ViewModel.fs#L60-L61

What does it mean when the outer `voption` is in the case `ValueSome` but the inner `voption` is in the case `ValueNone`?

cmeeren:
The inner is part of the actual value (i.e.`null` or not), and the outer is whether it's been set/initialized at all. So `ValueSome ValueNone` means "this sub-model binding has a `null` value" and `ValueNone` means "this sub-model binding does not have a value".

bender2k14:
Thanks for providing this short insight.

> It's probably more clear if you see where and how `SubModelSelectedItem` is used throughout that file. :)

I completely admit that I didn't look into how it was used.  I noticed it as I was working on logging and wanted to fire off this quick question instead of getting sidetracked.

PR #176, opened 2019-11-04
Initializing SubModelSelectedItem.Selected in initializeBinding instead of tryGetMember

bender2k14:
When resolving #135, I did the most narrow improvement at the time, which was "just" to get the `SubModelSeq` binding to the `SubModelSelectedItem` binding in order to decrease the number of calls to `failwith` in `ViewModel<,>`.

Now I see that I could have gone further.  As described in https://github.com/elmish/Elmish.WPF/issues/110#issuecomment-529373684, the outer option wrapping on the `Selected` field only exists because the `SubModelSeq` binding didn't used to be available during initialization of the `SubModelSelectedItem` binding.  Now that it is, we can remove this outer use of option.

### DIFF
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -78,7 +78,7 @@ and internal SubModelSelectedItemBinding<'model, 'msg, 'bindingModel, 'bindingMs
   Get: 'model -> 'id voption
   Set: 'id voption -> 'model -> unit
   SubModelSeqBinding: SubModelSeqBinding<'model, 'msg, obj, obj, obj>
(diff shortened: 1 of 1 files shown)


bender2k14:
I tested the `SubModelSeqSelectedItem` sample, and everything still works.

cmeeren:
The rationale behind the existing design is that it's a lazy operation; if the UI doesn't need the selected item for some reason, it's not necessary to get it. Similar to e.g. `oneWayLazy`.

Does that change your view of this?

bender2k14:
This PR conflicts with PR #177, and I was thinking this change was obviously better.  I will abandon this PR and create an issue for further discussion.

Issue #178, opened 2019-11-04
On the laziness of SubModelSelectedItem


bender2k14:
This is a continuation of #176 

Quoting https://github.com/elmish/Elmish.WPF/pull/176#issuecomment-549292167.
> The rationale behind the existing design is that it's a lazy operation; if the UI doesn't need the selected item for some reason, it's not necessary to calculate it. Similar to e.g. `oneWayLazy`. (Though admittedly, I'm not sure how relevant/beneficial this really is for `subModelSelectedItem`.)
> 
> Does that change your view of this?

It certainly does change my view in the sense that I didn't realize that the laziness was intended.  To be clear, my PR removed the laziness and made `SubModelSelectedItem` strict/eager.

I thought the laziness was an artifact of the implementation that existed before PR #138 (where each `SubModelSelectedItem` binding was initialized without access to the corresponding `SubModelSeq` binding) instead of a feature that tangibly benefits the user.

The lazy behaviors of `SubModelSelectedItem` and `OneWaySeq` are similar in the sense that they are both forms of [non-strict evaluation strategies](https://en.wikipedia.org/wiki/Evaluation_strategy#Non-strict_evaluation), but I think the similarity ends there.  Instead, I think the lazy behavior of `SubModelSelectedItem` is similar to the behavior of the evaluation strategy [call by name](https://en.wikipedia.org/wiki/Evaluation_strategy#Call_by_name) in the sense that the entire lifetime of the binding is like a single function call and some (user-provided) computation is skipped if the bound data is never requested.  If the bound data is ever requested, then the behavior of the binding "switches" to the strict/eager behavior explicit in my PR for all future interactions (which is really just the behavior in `updateValue` _after_ PR #177 is completed).

In contrast, I think the lazy behavior of `OneWaySeq` is similar to the [short-circuit evaluation](https://en.wikipedia.org/wiki/Short-circuit_evaluation) of most implementations of the logical operations `AND` and `OR`.  In this case, the possibility of skipping (user-defined) computation in the future is independent of the past.  This behavior is both similar to and a dual of the use of hash codes in equality functions as described in more detail in https://github.com/elmish/Elmish.WPF/issues/117#issuecomment-546344499:
> ...lazy bindings [like `OneWayLazy`] are able to short circuit computation when things are equal. In contrast, hashcodes are used in structural equality functions to short circuit computation when things are unequal.

[...]

PR #179, opened 2019-11-04
In SubModelSelectedItem, replaced the outer voption with Lazy<>

bender2k14:
Related to #178 (c.f. https://github.com/elmish/Elmish.WPF/issues/178#issuecomment-549569400).

The outer `voption` was being used to achieve some semblance of lazy behavior, but barely any.

This PR replaces it with `Lazy<>` and uses it the same way that the `OneWayLazy` binding does.  This PR also optimizes `updateValue` by first comparing the IDs of `newModel` and `currentModel` and continuing if and only if this expression is `false`, which is also something that the `OneWayLazy` binding does.

I tested the `SubModelSelectedItem` sample, and everything still works.

Side note:
I pushed three minor commits to `master` before making this PR to help keep this PR clean.


cmeeren:
Thanks!

PR #180, opened 2019-11-06
Properly dispose of Command instances

2sComplement:
I was noticing a memory leak in cases where sequences of sub-models containing command bindings were being removed. This was reproduced in a sample (included in this MR) where, once a sub-model containing a command binding is dynamically removed, the reference to the `ViewModel<'model,'msg>` instance is still present in memory. This was not the case for sub-models without command bindings.

Using Visual Studio's memory profiler, it appeared to be `execute` and `canExecute` inside the `Command` that were still hanging on the to `ViewModel` reference even after its underlying model had been removed. My solution was to have the `Command` class implement `IDisposable`, in which it removes references to `execute` and `canExecute`.

### DIFF
--- a/src/Elmish.WPF.sln
+++ b/src/Elmish.WPF.sln
@@ -57,6 +57,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "FileDialogs.CmdMsg.Views", 
 EndProject
 Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "FileDialogs.CmdMsg", "Samples\FileDialogs.CmdMsg\FileDialogs.CmdMsg.fsproj", "{F095E7E4-28ED-4223-A92F-4E86922E34EF}"
 EndProject
+Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "SubModelSeqMemory", "Samples\SubModelSeqMemory\SubModelSeqMemory.fsproj", "{97CAA5E1-B059-474B-9530-3CD7ECAC9675}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SubModelSeqMemory.Views", "Samples\SubModelSeqMemory.Views\SubModelSeqMemory.Views.csproj", "{E1527FE8-E653-40FD-9ADD-8CC54EF2C3D1}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -167,6 +171,14 @@ Global
 		{F095E7E4-28ED-4223-A92F-4E86922E34EF}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{F095E7E4-28ED-4223-A92F-4E86922E34EF}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{F095E7E4-28ED-4223-A92F-4E86922E34EF}.Release|Any CPU.Build.0 = Release|Any CPU
+		{97CAA5E1-B059-474B-9530-3CD7ECAC9675}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{97CAA5E1-B059-474B-9530-3CD7ECAC9675}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{97CAA5E1-B059-474B-9530-3CD7ECAC9675}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{97CAA5E1-B059-474B-9530-3CD7ECAC9675}.Release|Any CPU.Build.0 = Release|Any CPU
+		{E1527FE8-E653-40FD-9ADD-8CC54EF2C3D1}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{E1527FE8-E653-40FD-9ADD-8CC54EF2C3D1}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{E1527FE8-E653-40FD-9ADD-8CC54EF2C3D1}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{E1527FE8-E653-40FD-9ADD-8CC54EF2C3D1}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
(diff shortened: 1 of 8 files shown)


src/Elmish.WPF/ViewModel.fs
```diff
@@ -583,39 +591,44 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
...
+            |> Seq.iter b.Vms.Add
+
+            // Reorder according to new model list
+            for Kvp (newId, (newIdx, _)) in newSubModelIdxPairsById do
+              let oldIdx =
+                b.Vms
+                |> Seq.indexed
+                |> Seq.find (fun (_, vm) -> newId = b.GetId vm.CurrentModel)
+                |> fst
+              if oldIdx <> newIdx then b.Vms.Move(oldIdx, newIdx)
```
cmeeren: Everything here is indented one space too much

2sComplement: Fixed here, though I've noticed inconsistent indentation throughout the code base.

bender2k14: I have noticed that too.  I might be partially to blame for any inconsistencies.

2sComplement:
I'm happy to remove the sample, though it was helpful in debugging this issue. The existing subModelSeq sample had some added complexity that went further than a simple/typical subModelSeq sample. Let me know if you have a preference.

bender2k14:
> I'll review it carefully. @bender2k14, you're welcome to have a look at this, too. The more eyes, the better.

Sure.  I will also carefully review.

bender2k14:
Still haven't started my careful review.

Just wanted to say that I will be considering if using [weak references](https://docs.microsoft.com/en-us/dotnet/api/system.weakreference?view=netframework-4.8) would be a good solution the memory leak.

bender2k14:
I am still trying to verify your claim that there is a memory leak.

Whatever the case, I don't see how `Dispose` helps any.  It removes references by mutating fields that are not mutated anywhere else but those fields were initialized using references are still held in the constructor.

src/Elmish.WPF/InternalTypes.fs
```diff
@@ -4,26 +4,39 @@ module internal Elmish.WPF.InternalTypes
...
-type Command(execute, canExecute, autoRequery) as this =
+type Command(execute: obj -> unit, canExecute: obj -> bool, autoRequery) as this =
+
+  let mutable _execute = Some execute
+
+  let mutable _canExecute = Some canExecute
 
-  let canExecuteChanged = Event<EventHandler,EventArgs>()
-  let handler = EventHandler(fun _ _ -> this.RaiseCanExecuteChanged())
+  let mutable _canExecuteChanged = new Event<EventHandler,EventArgs>()
```
bender2k14: Why did you add `new`?

I thought that adding `new` here would cause a warning, but it didn't.  Through a warning, the compiler suggests that `new` should only be used to [construct classes](https://fsharpforfunandprofit.com/posts/classes/#constructing-and-using-a-class) that implement `IDisposable`, which [`Event<,>`](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/control.event%5b'delegate,'args%5d-class-%5bfsharp%5d) does not.

Is the lack of warning for this code is a bug in the compiler?  If so, then this code would produce a warning when this bug is fixed.

src/Elmish.WPF.sln
```diff
@@ -57,6 +57,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "FileDialogs.CmdMsg.Views",
 EndProject
 Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "FileDialogs.CmdMsg", "Samples\FileDialogs.CmdMsg\FileDialogs.CmdMsg.fsproj", "{F095E7E4-28ED-4223-A92F-4E86922E34EF}"
 EndProject
+Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "SubModelSeqMemory", "Samples\SubModelSeqMemory\SubModelSeqMemory.fsproj", "{97CAA5E1-B059-474B-9530-3CD7ECAC9675}"
```
bender2k14: The changes to this file didn't make it into your first commit.

Can you move the changes to this file from your second commit to your first commit and then force push the resulting branch?

2sComplement:
Why I believe there is a memory leak (assuming you've pulled the code from this PR):

1. Remove the call to `Dispose()` in `BindingManager.disposeBindings`.
2. Run the `SubModelSeqMemory` sample with profiling in Visual Studio
3. Click the "Add sub model" button under "With Command"
4. Click the "Remove sub model" button under "With Command"
5. Take a snapshot of the heap and notice that there is still a reference to the SubModel, originating from the button EventHandler

Notice that this does not happen when adding/removing subModels without command bindings. Add the `Dispose()` call back in, and repeat 2-5, you should see that the heap back down to baseline memory levels.

I could be misinterpreting this behaviour but it feels like a memory leak to me.

bender2k14:
The garbage collector is lazy.  You need to add a step in there where you force the garbage collector to reclaim as much memory as possible.

Try using [this branch](https://github.com/bender2k14/Elmish.WPF/tree/SubModelSeqMemory_with_manual_garbage_collection) and add explicit garbage collection to your steps.  When I added in this step, then In both cases, the memory usage returned to baseline.

2sComplement:
That worked. When I was testing, I was forcing a `GC.Collect()` inside the `updateValue` function itself, after the subModel was removed, and that was not cleaning up properly.  Maybe that was too soon for the GC to realize it needed to collect that instance?

cmeeren:
In any case, in your repro @2sComplement, what happens if you add/remove many times? If there truly is a memory leak, then it should grow unbounded. If it plateaus but then does not increase any further, there's no memory leak and GC is working (lazily) as it should.

bender2k14:
> ...what happens if you add/remove many times?

His program already does that.  Each time the user clicks either add button, so many objects are created that the memory usage increases by 1 GB.  After removing everything, calling the garbage collector reclaims all that memory.

cmeeren:
What I meant was, if continuing to remove/add sub-models, the GC should kick in sooner or later and the memory should not increase by 1GB for every click (not for long, at least). That would have been a sign that it's not a memory leak. Just a tip for the future. :)

bender2k14:
> Maybe that was too soon for the GC to realize it needed to collect that instance?

It seems so.  Maybe something still had a reference to it at that point or maybe it was in the second [generation](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/fundamentals#generations).  However, I don't think generations matter when doing a forced garbage collection.

> If there truly is a memory leak, then it should grow unbounded. If it plateaus but then does not increase any further, there's no memory leak and GC is working (lazily) as it should.
> 
> ...if continuing to remove/add sub-models, the GC should kick in sooner or later and the memory should not increase by 1GB for every click (not for long, at least). That would have been a sign that it's not a memory leak.

Before I added the forced garbage collection, I found it very difficult to tell if there was a memory leak or not.  On the one hand, all I needed to see is memory allocated from sub-models with a command not being reclaimed, but that hard part is know what percentage of the 1 GB is leaking.  The larger the percentage, the easier it would be to tell without forcing a garbage collection.

One complication is that the garbage collector not only lazily collects but also adjusts how much memory it wants to reclaim.  This makes the amount of memory reclaimed seem a bit random.

Another complication is expecting to see a difference between the sub-models with and without a command.

If we had a memory leak, my goal was to find a way to write a test that failed for that reason.  Not sure if that is possible, but I am glad that I don't have to try now :)

2sComplement:
The fact that was most concerning to me was that a snapshot of the heap still showed a subModel reference after it had been removed from the view, and only in the withCommand case. With @bender2k14's manual GC instead of the "eager" one I was testing with, it appears that is no longer the case. In theory, the command may have delayed the GC in marking that instance for collection.

@cmeeren when I add/remove/add/remove, the memory grows to the initial 1GB but does not continually grow from there, so it seems like this whole thing was a 70kg red herring.

Thanks for the review guys, and for the continued maintenance of this repo!

bender2k14:
> Thanks for the review guys, and for the continued maintenance of this repo!

It is my pleasure!  Working on Elmish.WPF is my favorite thing to do right now :D

PR #183, opened 2019-11-08
Recursive cached binding

### LINKED ISSUE #181: Add recursive cached binding
In https://github.com/elmish/Elmish.WPF/issues/143#issuecomment-549944305, I expressed some frustration with our use of `Lazy<>`.  It seems to me that the only benefit of our use of `Lazy<>` is if WPF is given a binding but never reads it.

Issue #178 was about improvements to the `SubModelSelectedItem` binding.  I thought of another improvement that also addresses my critique of our use of `Lazy<>`.

The improvement that I want to see for `SubModelSelectedItem` is not duplicating its "get" logic.  Specifically I am referring to lines 391 and 628.
```F#
// in initializeBindings
391 Selected = ref <| lazy (getSelectedSubViewModel b.Vms b.GetId get initialModel) }
// in updateValue
627 b.Selected := lazy (
628  let selected = getSelectedSubViewModel b.SubModelSeqBinding.Vms b.SubModelSeqBinding.GetId b.Get newModel
629  log "[%s] Setting selected VM to %A" propNameChain (selected |> ValueOption.map (fun vm -> b.SubModelSeqBinding.GetId vm.CurrentModel))
630  selected)
// in tryGetMember
672 | SubModelSelectedItem b ->
673    (!b.Selected).Value |> ValueOption.toObj |> box
```

My idea to fix this "problem" is twofold.  For simplicity, I will ignore logging.

First, remove the duplication by moving the "get" logic to `tryGetMember` _without_ any caching.  The above code would become something like this.
```F#
// in initializeBindings
// no more Selected field
// in updateValue
// no more Selected field
// in tryGetMember
| SubModelSelectedItem b ->
    getSelectedSubViewModel b.SubModelSeqBinding.Vms b.SubModelSeqBinding.GetId b.Get currentModel
    |> ValueOption.toObj |> box
```

Second, add caching back in via a new recursive binding type.  It would look something like this.

```F#
type internal CachedBinding<'model, 'msg> = {
  Cache: obj option ref
  Binding: VmBinding<'model, 'msg>
}
// in initializeBinding
Some <| Cached {
  Cache = ref None
  Binding = // initialize SubModelSelectedItem binding
}
// in updateValue
let valueChanged = // recursive call to updateValue
if valueChanged then b.Cache := None
valueChanged
// in tryGetMember
match b.Cached with
| Some v -> v
| None ->
  let v = // recursive call to tryGetMember
  b.Cached := Some v
  v
```

Let's review.

First, this both removes the duplicate "get" logic current present in `SubModelSelectedItem` as well as moves the "get" logic to the "get" function `tryGetMember`.

Second, the `Cached` case of `tryGetMember` effectively replicates duplicates the behavior of `Lazy<>`.  In particular, if WPF never requests the bound value of a given binding, then this code never executes the "get" logic.  So although I still don't know if WPF would ever do such a thing, now I don't care because I really like this code and it handles this case optimally.

Third, I have been trying to implement a recursive lazy binding in the style of [Elm's Html.Lazy](https://guide.elm-lang.org/optimization/lazy.html) (since about https://github.com/elmish/Elmish.WPF/issues/143#issuecomment-546219080).  This binding type is more complicated both because I was trying to have the user decide when to use it and because it is (or, at least, it currently seems to me to be) inherently more complicated than this idea of a recursive cached binding.  I am excited and eager to implement our first recursive binding, so I like that this cached binding seems simpler than the lazy binding.

Fourth, I ignored logging above, but we can add that back via a recursive logging binding.  This also helps us efficiently log because if (trace) logging is enabled, then we add recursive logging bindings during binding initialization.

bender2k14:
Fixes #181 

Creates a recursive binding type called `CachedBinding<'model, 'msg>`.  The caching in the `OneWayLazy` and `SubModelSelectedItem` cases are removed and now use this cached binding instead.

All the tests pass, and there is an existing test for `OneWayLazy` that ensures that caching is working.

I left one comment in the code to make sure we have a conversation about it.  I will create a comment in this PR on that line next.

### DIFF
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -671,7 +671,9 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
           b.SubModelSeqBinding.Vms 
           |> Seq.tryFind (fun (vm: ViewModel<obj, obj>) ->
             selectedId = ValueSome (b.SubModelSeqBinding.GetId vm.CurrentModel))
-        log "[%s] Setting selected VM to %A" propNameChain (selected |> Option.map (fun vm -> b.SubModelSeqBinding.GetId vm.CurrentModel))
+        log "[%s] Setting selected VM to %A"
+          propNameChain
+          (selected |> Option.map (fun vm -> b.SubModelSeqBinding.GetId vm.CurrentModel))
         selected |> Option.toObj |> box
     | Cached b ->
         match !b.Cache with
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -78,7 +78,6 @@ and internal SubModelSelectedItemBinding<'model, 'msg, 'bindingModel, 'bindingMs
   Get: 'model -> 'id voption
   Set: 'id voption -> 'model -> unit
   SubModelSeqBinding: SubModelSeqBinding<'model, 'msg, obj, obj, obj>
-  Selected: Lazy<ViewModel<'bindingModel, 'bindingMsg> voption> ref
 }
 
 and internal CachedBinding<'model, 'msg, 'value> = {
@@ -119,6 +118,9 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
 
   /// Error messages keyed by property name.
   let errors = Dictionary<string, string>()
+
+
+  let withCaching b = Cached { Binding = b; Cache = ref None }
 
 
   let log fmt =
@@ -391,11 +393,11 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
           let get = measure name "get" d.Get
           let set = measure2 name "set" d.Set
           let dispatch' = d.WrapDispatch dispatch
-          Some <| SubModelSelectedItem {
+          SubModelSelectedItem {
             Get = get
             Set = fun obj m -> set obj m |> dispatch'
             SubModelSeqBinding = b
-            Selected = ref <| lazy (getSelectedSubViewModel b.Vms b.GetId get initialModel) }
+          } |> withCaching |> Some
         | _ ->
           log "subModelSelectedItem binding referenced binding '%s', but no compatible binding was found with that name" d.SubModelSeqBindingName
           None
@@ -628,13 +630,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
             if oldIdx <> newIdx then b.Vms.Move(oldIdx, newIdx)
         false
     | SubModelSelectedItem b ->
-        if b.Get newModel = b.Get currentModel then false
-        else
-          b.Selected := lazy (
-            let selected = getSelectedSubViewModel b.SubModelSeqBinding.Vms b.SubModelSeqBinding.GetId b.Get newModel
-            log "[%s] Setting selected VM to %A" propNameChain (selected |> ValueOption.map (fun vm -> b.SubModelSeqBinding.GetId vm.CurrentModel))
-            selected)
-          true
+        b.Get newModel <> b.Get currentModel
     | Cached b ->
         let valueChanged = updateValue bindingName newModel b.Binding
         if valueChanged then
@@ -681,7 +677,9 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
         | WindowState.Hidden vm | WindowState.Visible vm -> box vm
     | SubModelSeq { Vms = vms } -> box vms
     | SubModelSelectedItem b ->
-        (!b.Selected).Value |> ValueOption.toObj |> box
+        let selected = getSelectedSubViewModel b.SubModelSeqBinding.Vms b.SubModelSeqBinding.GetId b.Get model
+        log "[%s] Setting selected VM to %A" propNameChain (selected |> ValueOption.map (fun vm -> b.SubModelSeqBinding.GetId vm.CurrentModel))
+        selected |> ValueOption.toObj |> box
     | Cached b ->
         match !b.Cache with
         | Some v -> v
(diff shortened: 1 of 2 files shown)


src/Elmish.WPF/ViewModel.fs
```diff
@@ -684,6 +696,11 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
           |> ValueOption.map (fun vm -> b.SubModelSeqBinding.GetId vm.CurrentModel)
         b.Set id model
         true
+    | Cached b ->
+        let successful = trySetMember model value b.Binding
+        if successful then
+          b.Cache := None // `None` is safter, but consider using `Some value` instead
```
bender2k14: This is in `trySetMember`.  If the set was successful, then `true` is returned and the cache is stale.  We can either assign `None` to empty the cache or assign `Some value` to populate the cache with the value that was just set.

The behavior of the caching currently in `master` for `OneWayLazy` and `SubModelSelectedItem` equivalent to assigning `None` here.  Assigning `Some value` instead would be a further optimization.

The risk with setting `Some value` is if anyone would purposefully return a different value than just set or if the user has a bug in their code that causes a different value to be returned than just set.  This "improved" caching would "hide" that bug--make it difficult to identify.

The Elm architecture is only supposed to store state in the model.  If we cache `Some value` here, then it might not be the same data that would be returned from `tryGetMember`.  This would definitely happen to someone eventually, and I don't expect they like that we made that choice.

So, assigning `None` mean the cache will be populated from the model and user-provided functions, which keeps things very understandable and predictable.  This is what I would prefer here.

cmeeren: Thanks for the detailed explanation. Let's definitely keep `None`. There are valid use-cases for returning different data than what is being set (e.g. format-as-you-type for mobile numbers).

src/Elmish.WPF/ViewModel.fs
```diff
@@ -78,7 +77,11 @@ and internal SubModelSelectedItemBinding<'model, 'msg, 'bindingModel, 'bindingMs
   Get: 'model -> 'id voption
   Set: 'id voption -> 'model -> unit
   SubModelSeqBinding: SubModelSeqBinding<'model, 'msg, obj, obj, obj>
-  Selected: Lazy<ViewModel<'bindingModel, 'bindingMsg> voption> ref
+}
+
+and internal CachedBinding<'model, 'msg> = {
+  Binding: VmBinding<'model, 'msg>
+  Cache: obj option ref
```
cmeeren: Should we parametrize `CachedBinding` by the `Cache` type, too? And only assign `obj` when used below in `VmBinding`? This would make it consistent with the other bindings.

src/Elmish.WPF/ViewModel.fs
```diff
@@ -684,6 +695,11 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
           |> ValueOption.map (fun vm -> b.SubModelSeqBinding.GetId vm.CurrentModel)
         b.Set id model
         true
+    | Cached b ->
+        let successful = trySetMember model value b.Binding
+        if successful then
+          b.Cache := None // `None` is safter, but consider using `Some value` instead
```
cmeeren: This can be moved up after `then` since it's so short (but I won't fight you if you prefer it like this)

cmeeren:
Thanks a lot! There's a lot to take in here, but I can't see anything obviously wrong. Could you also just verify that all relevant samples still work? If they do, and all tests still pass, then I think it's good to go (after resolving the other conversations).

src/Elmish.WPF/ViewModel.fs
bender2k14: Excellent example!

And that reminds me, I forgot to say this..

For each of the `OneWay`, `TwoWay`, and `TwoWayValidation` bindings, it is now a one-line change to add caching.  The example you gave for auto-formatting user input (like, say, making the first character capital if it is a letter) requires two-way functionality.

I suggest that we add this caching to `OneWay`, `TwoWay`, and `TwoWayValidation`.  Then I can document using a test that assigning `None` is the behavior that we want (i.e. the test will fail if `Some value` were assigned instead).

How does that sound?

cmeeren: Hmm... I'm not sure if we should make all those bindings caching by default. There's some overhead involved in caching, and for trivial getters, non-cached bindings may be more performant. I have no idea how much overhead there is, though. As far as I know, it may improve performance, or it may not.

Thoughts?

bender2k14: Your hesitation to add caching to the other three binding types is good.  We should to see how the performance compares.  I will open a separate issue for that.

I could make such a test for `SubModelSelectedItem`, but I think it would only be a pathological one.  One advantage of writing that test for, say, a `TwoWayBinding` is that I could use the excellent example behavior that you suggested.

bender2k14:
> There's a lot to take in here, but I can't see anything obviously wrong.

I tried to separate conceptual changes by commit.  Reviewing one commit at a time could help.

> Could you also just verify that all relevant samples still work? If they do, and all tests still pass, then I think it's good to go (after resolving the other conversations).

Will do.

src/Elmish.WPF/ViewModel.fs
bender2k14: Sure.  I squashed this change into the first commit and force pushed.

src/Elmish.WPF/ViewModel.fs
bender2k14: > Your hesitation to add caching to the other three binding types is good. We should test to see how the performance compares. I will open a separate issue for that.

Just did that with #184.

> Regarding the None test: Definitely 👍

How about we proceed without a test for now but I make an issue about adding such a test?

bender2k14:
> >Could you also just verify that all relevant samples still work? If they do, and all tests still pass, then I think it's good to go (after resolving the other conversations).
> 
> Will do.

For good measure, I tried all the samples.  Everything seemed to work.

src/Elmish.WPF/ViewModel.fs
cmeeren: > How about we proceed without a test for now but I make an issue about adding such a test?

Sure!

cmeeren:
> For good measure, I tried all the samples. Everything seemed to work.

Great, thanks!

src/Elmish.WPF/ViewModel.fs
bender2k14: Created issue #185

src/Elmish.WPF/ViewModel.fs
bender2k14: I will keep it like this because it is a mutation.

cmeeren:
Looks good to me. Merge when you're ready :)

PR #187, opened 2019-11-18
Update Readme to fix call to Binding.twoWay

nukecoder:
Binding.twoWay takes a tuple. This code example would not run as it was.


bender2k14:
Thanks @nukecoder! :)

PR #201, opened 2020-04-13
Bug fix 200 current application

### LINKED ISSUE #200: Problem when opening a subModelWin window as the initial command
Straightforward repro: in the NewWindow sample, if I submit the ShowWin1 message in the init function, an exception is thrown and the message is swallowed.

More specifically, I change `mkSimpleWpf` to `mkProgram` or `mkProgramWPF`, fix up the signature of the `bindings` function, add `, Cmd.none` to each branch of the `update` function, and then change the `init` function to
```
  let init () =
    { Win1State = WindowState.Closed
      Win1Input = ""
      Win2 = None }, Cmd.ofMsg ShowWin1
```

bender2k14:
Fixes #200

`Application.Current` in
https://github.com/elmish/Elmish.WPF/blob/3f0d3552203dcf3152f1088014ee608f2926f8a9/src/Elmish.WPF/ViewModel.fs#L199
was `null` when processing an initial state as described in #200.  After some clean up of the relevant code, this is a minimal change to fix that bug.

I didn't create an automated test for this bug.  As is, it needs to display one or two windows.  I have created a WPF test before (which needs to execute with a dedicated thread), but I have never tried to display a window in a test before, and I don't feel like trying to do that now.

There is another way to fix this bug that I will suggest in a new issue.

### DIFF
--- a/src/Elmish.WPF/Program.fs
+++ b/src/Elmish.WPF/Program.fs
@@ -33,10 +33,13 @@ let startElmishLoop
   |> Program.run
 
 
-/// Instantiates Application if it is not already running, and then runs the
-/// specified window. This is a blocking function.
-let private startApp window =
+/// Instantiates Application if it is not already running.
+let private instantiateApp () =
   if isNull Application.Current then Application () |> ignore
+
+
+/// Runs the specified window. This is a blocking function.
+let private startApp window =
   Application.Current.Run window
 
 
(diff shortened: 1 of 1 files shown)


cmeeren:
Thanks a lot! Should this be merged and released, or should I wait for the other issue you mention?

bender2k14:
No need to wait.  You can merge and release.

Issue #202, opened 2020-04-13
References to Application.Current


bender2k14:
This is the issue I promised in https://github.com/elmish/Elmish.WPF/pull/201#issue-402463416 that I would create.

`Application.Current` in
https://github.com/elmish/Elmish.WPF/blob/3f0d3552203dcf3152f1088014ee608f2926f8a9/src/Elmish.WPF/ViewModel.fs#L199
was `null` when processing an initial state as described in #200.  Indeed, this property is static with a private setter.  The fix in PR #201 implies to me that the constructor of `Application` sets that static property (probably like `Current = this`).  So that PR is a bug fix, but I also believe it does not satisfy the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).  I mean, look at how
https://github.com/elmish/Elmish.WPF/blob/3f0d3552203dcf3152f1088014ee608f2926f8a9/src/Elmish.WPF/Program.fs#L36-L40
creates an `Application` instance and then throws it away.  Very unexpected if you ask me.

Instead, I would prefer if we accessed the current `Application` instance in a way that I could statically verify in the Elmish.WPF code.  There is more than one way to achieve this.  Here is the way that I think is best.

We could implement the [F#-equivalent of a static constructor](https://fsharpforfunandprofit.com/posts/classes/#static-constructors) to initialize a static `let`-bound `Application` instance like
```F#
type CurrentApplication()
  let value =
    let current = Application.Current
    if isNull current then Application () else current
```

Then we replace all references to `Application.Current` with `CurrentApplication.value`.  As always, naming is impossibly hard, so we could certainly go with better names.

cmeeren:
> Very unexpected if you ask me.

I disagree; I don't find it unexpected in a WPF world. You have to instantiate `Application` once when starting your app (unless it's done implicitly in XAML), and then you access it using `Application.Current` because it's a singleton in WPF.

[...]

PR #203, opened 2020-04-14
Use window dispatcher

bender2k14:
This is a better fix to issue #200, which has already been fixed by PR #201.  This change also allows additional windows to be created in different threads.

Some related discussion is also in issue #202 (but I will include here the parts that I think are most relevant).

In https://github.com/elmish/Elmish.WPF/issues/202#issuecomment-613543070, I said
> ...the thought about which `Dispatcher` instance to use crossed my mind, but I don't know the difference between using one vs another.

I read up on this, and I think I understand this now.  The short answer is that each WPF object instance can only be modified by the thread that created it.  Don't know what thread created a certain WPF object instance?  No worries; just mutate that instance using its `Dispatcher` instance.

This is what @JDiLenarda said in https://github.com/elmish/Elmish.WPF/issues/202#issuecomment-613527232:
> ...the use of FrameworkElement instead of just Window, and replaced every instance of Application.Current.Dispatcher with Dispatcher from the current FrameworkElement in ViewModel module.
> 
> Currently, I cannot use windows submodel in such a plugin. Since it has not been big deal for now, I didn’t investigate the reason or opened issues, but I strongly suspect it’s because of Application.Current.Dispatcher calls. To my mind, such uses should be replaced by the Dispatcher of the main FrameworkElement.

Yes, @JDiLenarda.  That is exactly what the first commit in this PR does.  The `Window` returned by `getWindow` in
https://github.com/elmish/Elmish.WPF/blob/b62f4ed28f806847d6d8521216886e82fb64f6fd/src/Elmish.WPF/ViewModel.fs#L191-L199
is now mutated using its `Dispatcher` instead of the one returned from `Application.Current.Dispatcher`.

As such, this change avoids the `NullReferenceException` bug from issue #200.  To see that, on top of this branch, cherry pick the only commit on [this branch](https://github.com/mbenoni7/Elmish.WPF/tree/initial-cmd-bug) and run the `NewWindow` sample (because the second commit in the branch for this PR essentially reverts the fix from PR #201).

Furthermore, to see that this change also allows windows to be created in different threads,
- first checkout `master`, cherry pick the only commit on [this brach](https://github.com/bender2k14/Elmish.WPF/tree/testing/create_window_in_different_thread), run the `NewWindow` sample, click `Hide`, and observe an `InvalidOperationException`,
- then repeat all that but with the branch for this PR checked out instead of `master`.

### DIFF
--- a/src/Elmish.WPF/Program.fs
+++ b/src/Elmish.WPF/Program.fs
@@ -33,13 +33,10 @@ let startElmishLoop
   |> Program.run
 
 
-/// Instantiates Application if it is not already running.
-let private instantiateApp () =
-  if isNull Application.Current then Application () |> ignore
-
-
-/// Runs the specified window. This is a blocking function.
+/// Instantiates Application if it is not already running then runs the
+/// specified window. This is a blocking function.
 let private startApp window =
+  if isNull Application.Current then Application () |> ignore
   Application.Current.Run window
 
 
@@ -47,7 +44,6 @@ let private startApp window =
 /// Will instantiate Application if it is not already running, and then run the
 /// specified window. This is a blocking function.
 let runWindowWithConfig config window program =
-  instantiateApp ()
   startElmishLoop config window program
   startApp window
 
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -196,11 +196,11 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
       (onCloseRequested: unit -> unit)
       (preventClose: bool ref)
       initialVisibility =
-    Application.Current.Dispatcher.Invoke(fun () ->
+    let win = getWindow currentModel dispatch
+    winRef.SetTarget win
+    win.Dispatcher.Invoke(fun () ->
       let guiCtx = System.Threading.SynchronizationContext.Current
       async {
-        let win = getWindow currentModel dispatch
-        winRef.SetTarget win
         win.DataContext <- dataContext
         win.Closing.Add(fun ev ->
           ev.Cancel <- !preventClose
@@ -502,7 +502,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
           | true, w ->
               log "[%s] Closing window" winPropChain
               b.WinRef.SetTarget null
-              w.Close ()
+              w.Dispatcher.Invoke(fun () -> w.Close ())
           b.WinRef.SetTarget null
 
         let hide () =
@@ -511,7 +511,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
               log "[%s] Attempted to hide window, but did not find window reference" winPropChain
           | true, w ->
               log "[%s] Hiding window" winPropChain
-              w.Visibility <- Visibility.Hidden
+              w.Dispatcher.Invoke(fun () -> w.Visibility <- Visibility.Hidden)
 
         let showHidden () =
           match b.WinRef.TryGetTarget () with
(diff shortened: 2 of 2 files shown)


cmeeren:
Thanks a lot! I'm confused by what's changed from a user's perspective - could you give me a one-liner for the changelog?

bender2k14:
How about this?
> Now mutating each additional window using its own dispatcher, which allows it to be created on a different thread.

cmeeren:
Does this impact a user in any way? When would a user need to create windows directly, instead of managed through the Elmish.WPF API?

bender2k14:
> Does this impact a user in any way?

Technically the performance is worse, but not in a way that matters (because no one would open, close, and/or change the visibility of a window at a high rate).

Otherwise, no API change, everything that worked before still works, and this one thing that didn't work now works.

> When would a user need to create windows directly, instead of managed through the Elmish.WPF API?

I don't know what you mean by that difference (i.e. "create windows directly" vs "managed through the Elmish.WPF API").  The relevant Elmish.WPF API for this situation is the `Binding.subModelWin` methods, specifically the `getWindow` functions:
https://github.com/elmish/Elmish.WPF/blob/b62f4ed28f806847d6d8521216886e82fb64f6fd/src/Elmish.WPF/Binding.fs#L1386
https://github.com/elmish/Elmish.WPF/blob/b62f4ed28f806847d6d8521216886e82fb64f6fd/src/Elmish.WPF/Binding.fs#L1448
https://github.com/elmish/Elmish.WPF/blob/b62f4ed28f806847d6d8521216886e82fb64f6fd/src/Elmish.WPF/Binding.fs#L1502
https://github.com/elmish/Elmish.WPF/blob/b62f4ed28f806847d6d8521216886e82fb64f6fd/src/Elmish.WPF/Binding.fs#L1556
https://github.com/elmish/Elmish.WPF/blob/b62f4ed28f806847d6d8521216886e82fb64f6fd/src/Elmish.WPF/Binding.fs#L1604
https://github.com/elmish/Elmish.WPF/blob/b62f4ed28f806847d6d8521216886e82fb64f6fd/src/Elmish.WPF/Binding.fs#L1653

Elmish.WPF does not instantiate the window; it is returned by these `getWindow` functions.  It is user code that "create[s] windows directly".  If that instantiation happens on a different thread, then the use of `Application.Current.Dispatcher` is not correct.

I agree that the code I wrote to demonstrate this problem looks contrived.  I was just trying to create a minimal working example.  @JDiLenarda seems to have an actual use case mind, which is about a plugin.  See https://github.com/elmish/Elmish.WPF/issues/202#issuecomment-613527232 (which I also quoted a bit in my initial comment of this PR).

bender2k14:
> Furthermore, to see that this change also allows windows to be created in different threads,
> - first checkout `master`, cherry pick the only commit on [this brach](https://github.com/bender2k14/Elmish.WPF/tree/testing/create_window_in_different_thread), run the `NewWindow` sample, click `Hide`, and observe an `InvalidOperationException`,
> - then repeat all that but with the branch for this PR checked out instead of `master`.

If you want, I can try to make this into an automated test before we merge this PR in.

bender2k14:
> I can try to make this into an automated test before we merge this PR in.

I would attempt to do this with a `Window` class defined in C# that I would put in a project named something like `Elmish.WPF.Tests.CSharp`.

Does that sounds ok?

JDiLenarda:
I’ll try to see if it solves the issue I brought in #202 as soon as possible. I know I’m in a very corner case with my use of Elmish.Wpf for writing plugin, but I think one should not assume it would be used only for stand alone application, neither that an elmish program is always bound to a Window (in my case, it’s bound to a UserControl - no choice, there, that how you put a WPF interface in an AutoCAD palette).

By the way, [one link](https://stackoverflow.com/questions/20025143/application-current-dispatcher-vs-dependencyobject-dispatcher-for-ui-thread-acce) I gave in #202 promotes the use of Dispatcher from framework elements rather than from Application.

cmeeren:
> Elmish.WPF does not instantiate the window; it is returned by these `getWindow` functions. It is user code that "create[s] windows directly". If that instantiation happens on a different thread, then the use of `Application.Current.Dispatcher` is not correct.

Oh, of course. Thanks :)

I'm ready to merge this and don't see the need for a test. (Neither do I see the need to have a separate C# test project, but it's been a while since I've touched the code in this project.) I'll leave that up to you.

bender2k14:
Ok, let's merge then :)

PR #205, opened 2020-04-18
logo

MrMattSim:
Various formats and use-cases under /logo
Replaced old versions (readme and nuget src)

Modified layout on readme page. I think a slightly smaller scale and centered with generous padding works best (esp. on mobile).
Markdown Preview: https://github.com/MrMattSim/Elmish.WPF/tree/%23204_preview-logo-layout


MrMattSim:
Okay, I'll get started. 
The only point I would contest is removing the `afdesign` file. It's the master vector format (Affinity Design) which batch exports to all other formats, scales, and lockups with a single click. 

As for the other vectors, keeping just the `svg` is probably adequate here, so you're not dependant on Affinity Designer. (I'm just in the habit of exporting a few formats because graphics conversion is often unreliable.)

Sorry about the markdown reformatting. I blame VS Code.

cmeeren:
Thanks! I trust your judgement regarding which files to keep.

MrMattSim:
done

cmeeren:
Excellent, thank you so much for all the work! <3

cmeeren:
Acknowledgement added in https://github.com/elmish/Elmish.WPF/commit/66c909041a09c3d3e64326893459a8cdc1ea9a45

Issue #206, opened 2020-04-23
Problems setting Window.Owner when starting application


bender2k14:
Issue original reported in https://github.com/elmish/Elmish.WPF/issues/200#issuecomment-618089747.

For the `NewWindow` sample, if the message `ShowWin2` is returned by `init`, then there is a `NullReferenceException` because `Application.Current` is `null` in
https://github.com/elmish/Elmish.WPF/blob/66c909041a09c3d3e64326893459a8cdc1ea9a45/src/Samples/NewWindow/Program.fs#L104

MWE given in the branch https://github.com/mbenoni7/Elmish.WPF/tree/init-cmd-bug-pt-2

bender2k14:
PR #207 makes some progress, but fixing the `NullReferenceException` uncovers another issue.

After making that change, there is a `InvalidOperationException` because [`Window.Owner` ](https://docs.microsoft.com/en-us/dotnet/api/system.windows.window.owner?view=netframework-4.8) is being
> ...set with a window that has not been previously shown.

It is not obvious to me how to avoid this.  I will keep thinking about it.

cmeeren:
Brainwave, this may not make sense (I'm not that into the code at the moment). But won't everything be fine as long as we guarantee that `Application` is instantiated and the main window shown before we start the Elmish loop?

bender2k14:
Yes, that should work.  I will try that.

PR #207, opened 2020-04-23
Replace call to Application.Current with function argument

bender2k14:
This is a step on the road to fixing issue #206 

However, after this change, issue #206 is still not fixed.  Instead of a `NullReferenceException` there is now a `InvalidOperationException`.

### DIFF
--- a/src/Samples/NewWindow/Program.fs
+++ b/src/Samples/NewWindow/Program.fs
@@ -78,7 +78,7 @@ module App =
         | None -> m
 
 
-  let bindings () : Binding<Model, Msg> list = [
+  let bindings mainWindow () : Binding<Model, Msg> list = [
     "ShowWin1" |> Binding.cmd ShowWin1
     "HideWin1" |> Binding.cmd HideWin1
     "CloseWin1" |> Binding.cmd CloseWin1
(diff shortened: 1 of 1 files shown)


bender2k14:
See https://github.com/elmish/Elmish.WPF/issues/206#issuecomment-618111981 for more information about why issue #206 is still unresolved even after completing PR.

bender2k14:
BTW, this PR just changes code in the `NewWindow` sample, so it is rather safe to complete.

cmeeren:
Thanks! I think perhaps we should wait with merging this, since it doesn't really fix the issue.

bender2k14:
After resolving #200, I had been thinking that we should replace (almost) all calls to `Application.Current` to improve robustness.

So instead of this PR, how about I create another PR that does that?

cmeeren:
How is that needed? This problem here stems from a call in a sample. i.e. in user code. There is nothing we can do about user code, so it seems this is unrelated to calling `Application.Current` in Elmish.WPF. I still think everything in [this comment](https://github.com/elmish/Elmish.WPF/issues/202#issuecomment-613005544) (which you later agreed with) applies just as much as it did back then.

bender2k14:
Not actually needed.  Even though it is user code, I was trying to make the samples better.

But thanks for pushing back.  I think I found a better solution.

PR #208, opened 2020-04-23
initialize main Window before starting Elmish dispatch loop #206

### LINKED ISSUE #206: Problems setting Window.Owner when starting application

bender2k14:
Fixes #206

As @cmeeren suggested in https://github.com/elmish/Elmish.WPF/issues/206#issuecomment-618504164, this PR initializes `Application` and shows the provided `Window` before starting the Elmish dispatch loop.

One additional change is also needed though, which is to set the provided window to `Applicaiton.Current.MainWindow`.  The [documentation](https://docs.microsoft.com/en-us/dotnet/framework/wpf/app-development/how-to-get-and-set-the-main-application-window) says that...
> The first Window that is instantiated within a Windows Presentation Foundation (WPF) application is automatically set by `Application` as the main application window.

...but this is only true if `Application` has been instantiated.  Therefore, this PR also assigns the provided `Window` to `Application.Current.MainWindow` if `Application` had not yet been instantiated.

### DIFF
--- a/src/Elmish.WPF/Program.fs
+++ b/src/Elmish.WPF/Program.fs
@@ -33,24 +33,28 @@ let startElmishLoop
   |> Program.run
 
 
-/// Instantiates Application if it is not already running then runs the
-/// specified window. This is a blocking function.
-let private startApp window =
-  if isNull Application.Current then Application () |> ignore
-  Application.Current.Run window
+/// Instantiates Application and sets its MainWindow if it is not already
+/// running.
+let private initializeApplication window =
+  if isNull Application.Current
+  then
+    Application () |> ignore
+    Application.Current.MainWindow <- window
 
 
 /// Starts the Elmish and WPF dispatch loops with the specified configuration.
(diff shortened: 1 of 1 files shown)


cmeeren:
Thanks a lot! Everything's ready for release then, I suppose?

bender2k14:
Thanks for the suggestion about how to fix the issue! :)

Yes, I think everything is ready for release.

Issue #195, opened 2020-01-18
Q: Are there any Undo/Redo suggestions?


KingKnecht:
How could I implement a undo/redo functionality?

Approaches that I was thinking about:
* Store the state of the model every time a messages has been sent.
  * Problem: Memory consumption
* Store each message sent
  * Undo = replay messages from init state to the desired undo state
  * Problems:
    * Could be slow
    * If some non pure functional code is involved e.g. Guid.NewGuid(), the result will not be the same
 * Store an inverting function for every message that has been processed
    * Problems:
      * How to store them if there are many different types of inverter functions?
      * What if the invert function can't be found? Does a inverter function always exist?

Are there any other ideas? Or even examples? How does the time-warp debugger of Elmish work?

cmeeren:
While Elm/Elmish architectures make undo/redo simpler, there is no silver bullet. An undo strategy is completely up to the implementer, and will vary based on application requirements. Relevant questions are: What in particular should be undoable? Do you want to undo several things separately? And many more.

Undo based on whole state: Probably not a good idea. Do you really want to undo UI stuff such as hiding/showing of dialogs etc.?

Regarding GUIDs: You could keep `update` deterministic and make sure that a GUID is included in the message.

Closing since the choice of a redo strategy is not related to Elmish.WPF and thus not something to be "solved" here, but feel free to continue the discussion. In general though, you might be better off asking this on a general Elmish or Elm discussion forum.

bender2k14:
Making a prototype with this feature sounds fun.  I might try this sometime.  I will share here if I do.

bender2k14:
I created a demo here https://github.com/bender2k14/Elmish.WPF/commits/demo/undo_redo

It adds undo-redo functionality to the simple counter sample.  It stores the model every time a message is sent via a function called `push`.  However, I also added a function called `map` that changes the current model without also changing the undo-redo state.

I am going to use this code in my application at work.  My plan is for the parent context to use either `map` or `push` to perfectly control the undo-redo state.

I will comment again to provide an update on that use case.

bender2k14:
I forced pushed many times to that branch.  I am satisfied with that code now.

All changes to the current state are done via `map`.

The functionality that used to be called `push` is now slightly simpler and called `mark`.  You call `mark` when you want to mark the current state and push the previously marked state onto the undo stack.  However, this only happens if the current state and the marked state are different as determined by a given `equals` function.  Otherwise, nothing happens.

[...]

PR #212, opened 2020-05-05
Add Bindings.mapModel

bender2k14:
I recently realized that `Binding<'model, 'msg>` is a ~co~ _contravariant_ functor in `'model`.  As such, there exists a function `map` with signature
```F#
('a -> 'b) -> Binding<'b, 'c> -> Binding<'a, 'c>
```
satisfying the ~co~ _contravariant_ functor laws.

This PR adds that function under the name `Binding.mapModel` (and the `list` variant as `Bindings.mapModel`) and then simplifies the boxing of a binding by making use of this new function.

The scope of both of those functions is public since I expect these could be useful to users.  (I am trying to make use of it as I clean up my Undo/Redo example https://github.com/elmish/Elmish.WPF/issues/195#issuecomment-621443558).

I tested the `SubModelSeq` sample and it still works.

(Side note: I want to define a similar function `mapMsg` under the premise that `Binding<'model, 'msg>` is a functor in `'msg`...but I don't think it is....because of `WrapDispatch` has signature `Dispatch<'msg> -> Dispatch<'msg>`.  I need to think more about this.)

### DIFF
--- a/src/Elmish.WPF/Binding.fs
+++ b/src/Elmish.WPF/Binding.fs
@@ -1,4 +1,4 @@
-﻿namespace Elmish.WPF
+namespace Elmish.WPF
 
 open System.Windows
 
@@ -230,6 +230,83 @@ module internal BindingData =
         SubModelSeqBindingName = d.SubModelSeqBindingName
         WrapDispatch = boxWrapDispatch unboxMsg boxMsg d.WrapDispatch
       }
+
+
+module internal Binding =
+
+  let mapModel f binding =
+    let binaryHelper binary x m = (x, f m) ||> binary
+    let newData =
+      match binding.Data with
+      | OneWayData d ->
+          { Get = f >> d.Get
+          } |> OneWayData
+      | OneWayLazyData d ->
+          { Get = f >> d.Get
+            Map = d.Map;
+            Equals = d.Equals
+          } |> OneWayLazyData
+      | OneWaySeqLazyData d ->
+          { Get = f >> d.Get
+            Map = d.Map
+            Equals = d.Equals
+            GetId = d.GetId
+            ItemEquals = d.ItemEquals
+          } |> OneWaySeqLazyData
+      | TwoWayData d ->
+          { Get = f >> d.Get
+            Set = binaryHelper d.Set
+            WrapDispatch = d.WrapDispatch
+          } |> TwoWayData
+      | TwoWayValidateData d ->
+          { Get = f >> d.Get
+            Set = binaryHelper d.Set
+            Validate = f >> d.Validate
+            WrapDispatch = d.WrapDispatch
+          } |> TwoWayValidateData
+      | CmdData d ->
+          { Exec = f >> d.Exec
+            CanExec = f >> d.CanExec
+            WrapDispatch = d.WrapDispatch
+          } |> CmdData
+      | CmdParamData d ->
+          { Exec = binaryHelper d.Exec
+            CanExec = binaryHelper d.CanExec
+            AutoRequery = d.AutoRequery
+            WrapDispatch = d.WrapDispatch
+          } |> CmdParamData
+      | SubModelData d ->
+          { GetModel = f >> d.GetModel
+            GetBindings = d.GetBindings
+            ToMsg = d.ToMsg
+            Sticky = d.Sticky
+          } |> SubModelData
+      | SubModelWinData d ->
+          { GetState = f >> d.GetState
+            GetBindings = d.GetBindings
+            ToMsg = d.ToMsg
+            GetWindow = f >> d.GetWindow
+            IsModal = d.IsModal
+            OnCloseRequested = d.OnCloseRequested
+          } |> SubModelWinData
+      | SubModelSeqData d ->
+          { GetModels = f >> d.GetModels
+            GetId = d.GetId
+            GetBindings = d.GetBindings
+            ToMsg = d.ToMsg
+          } |> SubModelSeqData
+      | SubModelSelectedItemData d ->
+          { Get = f >> d.Get
+            Set = binaryHelper d.Set
+            SubModelSeqBindingName = d.SubModelSeqBindingName
+            WrapDispatch = d.WrapDispatch
+          } |> SubModelSelectedItemData
+    { Name = binding.Name; Data = newData }
+
+
+module internal Bindings =
+
+  let mapModel f bindings = bindings |> List.map (Binding.mapModel f)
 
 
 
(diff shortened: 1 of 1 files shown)


cmeeren:
Thanks!

1. Regarding making it public: Could you give me a use-case where it's useful for users to map the model in a binding?
2. The signature you posted
  ```f#
   ('a -> 'b) -> Binding<'b, 'c> -> Binding<'a, 'c>
   ```
   does't make sense to me. Are you sure you haven't switched `'a` and `'b` in the `Binding` types? I can't see how a function `'a -> 'b` can possibly take you from `Binding<'b, 'c>` to `Binding<'a, 'c>`.

bender2k14:
I will separate my lengthy answers into separate comments for clarity.

Actually, I will make the scope private for now.

My goal was to first simplify the undo-redo example and then make this PR, but that hasn't happened yet.  However, I did find an internal use for this mapping function, so I thought I could contribute it for that reason.

I took Mark Seemann's workshop [From Design Patterns to universal abstractions](https://www.ndcconferences.com/slot/from-design-patterns-to-universal-abstractions) last summer (at NDC Minnesota 2019).  In brief, one of my takeaways is that abstractions motivated by mathematics are good abstractions.  So I believe that this is a good code to add even if I can't (yet!) demonstrate its usefulness to a user.

Nonetheless, what I am trying to achieve is what I think of as a lightweight alternative to a `SubModel` binding.  I could be wrong, but...if I create a `SubModel` binding, then I have to compensate in the XAML by changing the `DataContext`.  Most of the time I am ok with this, but I don't like this when I want to intersperse the bindings for what would normally be the bindings in two different submodels.  Conceptually, think of separating the F# code for a counter and its bindings from the F# code for undo-redo and its bindings while also allowing the XAML code to appear as though all that F# code was combined.

To circle back, making these two functions private is still helpful to me.  In my undo-redo work, I can just remove the private keyword, which is trivial.  Until now, I have essentially had this code in a stash that I keep applying as I try one approach, discard all my changes, try another approach, and so on.

bender2k14:
Ah, yes.  You are right.  It is not a co-functor.  It is a contravariant functor.  That same [StackOverflow answer](https://stackoverflow.com/questions/34732571/why-there-is-no-cofunctor-typeclass-in-haskell/34732721#34732721) you linked to points out this concept at the end.

While I was an object oriented programmer, one thing I learned was [type variance](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)).  If `F<'a>` is a (covariant) functor in `'a`, then I think it is also covariant in `'a`.  Similarly, if `F<'a>` is a contravariant functor in `'a`, then I think it is also contravariant in `'a`.

For a (covariant) functor `F<'a>`, we think of it as containing or having the ability to produce or can return an instance of type `'a`.  In this case, producing a `F<'b>` from a `F<'a>` requires a function of type `a -> b`.  In contrast, a contravariant functor `F<'a>` would be thought of as having the ability to consume an instance of type `'a`.  As such, producing a `F<'b>` from a `F<'a>` requires a function of type `b -> a`.

A simple example of a contravariant functor is the function type `'a -> unit`.  Producing `'b -> unit` from `'a -> unit` requires a function of type `b -> a`.

And guess what?  [`Dispatcher<'msg>`](https://github.com/elmish/elmish/blob/12648f57092ac6c3dca383e1c25995882a646447/src/cmd.fs#L13), which is just a type alias for `'msg -> unit`, is exactly that.  `Dispatcher<'msg>` is a contravariant functor in `'msg`.

cmeeren:
Ah, thanks. Got it now. I always think of functors as "containers" of some other type, which is obviously not a general truth.

bender2k14:
Yes, there are two adjectives, covariant and contravariant.  For perfect clarity, exactly one would always be used.  Instead, since we almost always are talking about covariant things, we never use that adjective and instead only use contravariant when want that case.

I think the exact same problem exists in linear algebra with vectors.  There are row vectors and column vectors, but (I think) a vector without a provided adjective is meant to be a column vector.

cmeeren:
I see. Feel free to merge when you're ready!

bender2k14:
> Regarding making it public: Could you give me a use-case where it's useful for users to map the model in a binding?

For future reference, here is an exmaple of how I would like to use `Bindings.mapModel` and `Bindings.mapMsg`.

In my program at work, we have a drawing editor screen that currently has 17 bindings.  Here are are how five of them look now.
```
fun () ->
  [ "MouseLeftButtonDown" |> Binding.cmdParam(RawDown >> RawInputMsg |> onMouseEvent)
    "MouseLeftButtonUp" |> Binding.cmdParam(RawUp >> RawInputMsg |> onMouseEvent)
    "MouseMove" |> Binding.cmdParam(RawMove >> RawInputMsg |> onMouseEvent)
    "Undo" |> Binding.cmdIf(Undo |> UndoRedoMsg, (fun m -> m.Core) >> canUndo)
    "Redo" |> Binding.cmdIf(Redo |> UndoRedoMsg, (fun m -> m.Core) >> canRedo)
  ]
```

Here is how I could simplify them with `Bindings.mapModel` and `Bindings.mapMsg`:
```
fun () ->
  [ "MouseLeftButtonDown" |> Binding.cmdParam(RawDown |> onMouseEvent)
    "MouseLeftButtonUp" |> Binding.cmdParam(RawUp |> onMouseEvent)
    "MouseMove" |> Binding.cmdParam(RawMove |> onMouseEvent)
  ] |> Bindings.mapMsg RawInputMsg
  @
  [ "Undo" |> Binding.cmdIf(Undo, canUndo)
    "Redo" |> Binding.cmdIf(Redo, canRedo)
  ] |> Bindings.mapMsg UndoRedoMsg |> Bindings.mapModel (fun m -> m.Core)
```

Even though reasonable people could probably disagree about which of these two is "better", I still think this is a decedent example of how `Bindings.mapModel` and `Bindings.mapMsg` could be used by a user.

cmeeren:
Thanks for the example. Personally I think the former is easier to read and understand. Also, I'd guess that most cases where it's relevant to map messages would be the sub-model bindings, where this is already built in.

bender2k14:
Indeed, these mappings are built in with the `SubModel` binding.

I could be wrong, but I think using two `SubModel` bindings means that bindings within each sub-model must either
1. satisfy the constraint of being in sub-trees of the XAML with different settings for `DataContext` (bound to their respective sub-models) or
2. be more complex by being in a sub-tree of the XAML with the `DataContext` bound to a parent sub-model and then have more complicated XAML binding expressions that also include the mappings that could otherwise have been applied to the binding in F# via `Binding.mapModel`.

Most of the time, it is not an issue satisfying the constraint (1), so I go with that option.  On the other hand, I haven't even looked into how to express the XAML bindings in the more complicated case (2) because I don't want to push any complexity into the XAML.  And even though I think it is probably possible to express a "`mapModel`" mapping in the XAML, I doubt that it is possible to express a "`mapMsg`" mapping in the XAML.

PR #214, opened 2020-05-15
Use Elm-style merge

### LINKED ISSUE #137: Optimizations using IDs

bender2k14:
Fixes #137 

This branch implements [merge like Elm does](https://github.com/elm/virtual-dom/blob/master/src/Elm/Kernel/VirtualDom.js#L980-L1226).  I have tested it in my application at work, and the performance is clearly better.

However, I don't want to merge this PR yet.  I want to write more tests to cover all the branches.  Then I will write performance tests with BenchmarkDotNet and share those results here.

I am creating this PR before I am ready to merge as an invitation for feedback.

### DIFF
--- a/src/Elmish.WPF.Tests/MergeTests.fs
+++ b/src/Elmish.WPF.Tests/MergeTests.fs
@@ -9,11 +9,10 @@ open Swensen.Unquote
 open Elmish.WPF
 
 
-let logNoOp _ _ _ = ()
 let getIdAsId = id
 let createAsId a _ = a
 let updateNoOp _ _ _ = ()
-let merge x = x |> historicalMerge logNoOp logNoOp
+let merge x = x |> elmStyleMerge
 
 
 let private trackCC (observableCollection: ObservableCollection<_>) =
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -69,6 +69,157 @@ module internal ViewModelHelpers =
           |> fst
         if tIdx <> sIdx then target.Move(tIdx, sIdx)
 
+  let elmStyleMerge
+        getSourceId
+        getTargetId
+        create
+        update
+        (target: ObservableCollection<_>)
+        (source: _ array) =
+    (*
+     * Based on Elm's HTML.keyed
+     * https://guide.elm-lang.org/optimization/keyed.html
+     * https://github.com/elm/virtual-dom/blob/5a5bcf48720bc7d53461b3cd42a9f19f119c5503/src/Elm/Kernel/VirtualDom.js#L980-L1226
+     *)
+    let removals = Dictionary<_, _> ()
+    let additions = Dictionary<_, _> ()
+
+    let recordRemoval curTargetIdx curTarget curTargetId =
+      removals.Add(curTargetId, (curTargetIdx, curTarget))
+    let recordAddition curSourceIdx curSource curSourceId =
+      additions.Add(curSourceId, (curSourceIdx, curSource))
+
+    let mutable curSourceIdx = 0
+    let mutable curTargetIdx = 0
+
+    let mutable shouldContinue = true
+
+    let sourceCount = source.Length
+    let targetCount = target.Count
+
+    while (shouldContinue && curSourceIdx < sourceCount && curTargetIdx < targetCount) do
+      let curSource = source.[curSourceIdx]
+      let curTarget = target.[curTargetIdx]
+
+      let curSourceId = getSourceId curSource
+      let curTargetId = getTargetId curTarget
+
+      if curSourceId = curTargetId then
+        update curTarget curSource curTargetIdx
+
+        curSourceIdx <- curSourceIdx + 1
+        curTargetIdx <- curTargetIdx + 1
+      else
+        let mNextSource =
+          source
+          |> Array.tryItem (curSourceIdx + 1)
+          |> Option.map (fun s ->
+            let id = getSourceId s
+            s, id, id = curTargetId) // true => need to add
+
+        let mNextTarget =
+          if curTargetIdx + 1 < targetCount then target.[curTargetIdx + 1] |> Some else None
+          |> Option.map (fun t ->
+            let id = getTargetId t
+            t, id, id = curSourceId) // true => need to remove
+
+        match mNextSource, mNextTarget with
+        | Some (nextSource, _, true), Some (nextTarget, _, true) -> // swap adjacent
+            target.[curTargetIdx] <- nextTarget
+            target.[curTargetIdx + 1] <- curTarget
+
+            update curTarget nextSource (curTargetIdx + 1)
+            update nextTarget curSource curTargetIdx
+            
+            curSourceIdx <- curSourceIdx + 2
+            curTargetIdx <- curTargetIdx + 2
+        |               None, Some (nextTarget, _, true)
+        | Some (_, _, false), Some (nextTarget, _, true) -> // remove
+            recordRemoval curTargetIdx curTarget curTargetId
+            
+            update nextTarget curSource curTargetIdx
+
+            curSourceIdx <- curSourceIdx + 1
+            curTargetIdx <- curTargetIdx + 2
+        | Some (nextSource, _, true), None
+        | Some (nextSource, _, true), Some (_, _, false) -> // add
+            recordAddition curSourceIdx curSource curSourceId
+            
+            update curTarget nextSource (curTargetIdx + 1)
+            
+            curSourceIdx <- curSourceIdx + 2
+            curTargetIdx <- curTargetIdx + 1
+        | Some (_, _, false),               None
+        |               None, Some (_, _, false)
+        |               None,               None -> // source and target have different lengths and we have reached the end of one
+            shouldContinue <- false 
+        | Some (nextSource, nextSourceId, false), Some (nextTarget, nextTargetId, false) ->
+            if nextSourceId = nextTargetId then // replace
+              recordRemoval curTargetIdx curTarget curTargetId
+              recordAddition curSourceIdx curSource curSourceId
+              
+              update nextTarget nextSource (curTargetIdx + 1)
+
+              curSourceIdx <- curSourceIdx + 2
+              curTargetIdx <- curTargetIdx + 2
+            else // collections very different
+              shouldContinue <- false
+
+    // replace many
+    while (curSourceIdx < sourceCount && curTargetIdx < targetCount) do
+      let curSource = source.[curSourceIdx]
+      let curTarget = target.[curTargetIdx]
+
+      let curSourceId = getSourceId curSource
+      let curTargetId = getTargetId curTarget
+
+      recordRemoval curTargetIdx curTarget curTargetId
+      recordAddition curSourceIdx curSource curSourceId
+      
+      curSourceIdx <- curSourceIdx + 1
+      curTargetIdx <- curTargetIdx + 1
+
+    // remove many
+    for i in targetCount - 1..-1..curTargetIdx do
+      let t = target.[i]
+      let id = getTargetId t
+      recordRemoval i t id
+
+    // add many
+    for i in curSourceIdx..sourceCount - 1 do
+      let s = source.[i]
+      let id = getSourceId s
+      recordAddition i s id
+
+    let moves =
+      additions
+      |> Seq.toList
+      |> List.collect (fun (Kvp (id, (sIdx, s))) ->
+        match removals.TryGetValue id with
+        | (false, _) -> []
+        | (true, (tIdx, t)) ->
+            removals.Remove id |> ignore
+            additions.Remove id |> ignore
+            (tIdx, sIdx, t, s) |> List.singleton)
+
+    // actually remove
+    Seq.empty
+    |> Seq.append (removals |> Seq.map (Kvp.value >> fst))
+    |> Seq.append (moves |> Seq.map (fun (tIdx, _, _, _) -> tIdx))
+    |> Seq.sortDescending
+    |> Seq.iter target.RemoveAt
+
+    // actually add
+    Seq.empty
+    |> Seq.append (additions |> Seq.map (fun (Kvp (id, (idx, s))) -> idx, create s id))
+    |> Seq.append (moves |> Seq.map (fun (_, sIdx, t, _) -> sIdx, t))
+    |> Seq.sortBy fst
+    |> Seq.iter target.Insert
+
+    // update moved elements
+    moves |> Seq.iter (fun (_, sIdx, t, s) -> update t s sIdx)
+
+
 type internal OneWayBinding<'model, 'a> = {
   Get: 'model -> 'a
 }
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -338,8 +338,6 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
       if config.LogTrace then Diagnostics.Trace.WriteLine(str)
     Printf.kprintf innerLog fmt
 
-  let logInvalidGetId = log "The getId function must return distinct IDs, but it returned the same ID %A for %A and %A"
-
   let getPropChainFor bindingName =
     sprintf "%s.%s" propNameChain bindingName
 
@@ -444,12 +442,6 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
         let map = measure name "map" d.Map
         let getId = measure name "getId" d.GetId
         let values = ObservableCollection(initialModel |> get |> map)
-        let valuesById = Dictionary<_,_>(values.Count)
-        for value in values do
-          let id = getId value
-          if valuesById.ContainsKey id
-          then logInvalidGetId id (valuesById.[id]) value
-          else valuesById.Add(id, value)
         Some <| OneWaySeq {
           Get = get
           Map = map
@@ -577,12 +569,6 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
                ViewModel(m, (fun msg -> toMsg (getId m, msg) |> dispatch), getBindings (), config, chain)
           )
           |> ObservableCollection
-        let modelsById = Dictionary<_,_>(vms.Count)
-        for model in vms |> Seq.map (fun vm -> vm.CurrentModel) do
-          let id = getId model
-          if modelsById.ContainsKey id
-          then logInvalidGetId id (modelsById.[id]) model
-          else modelsById.Add(id, model)
         Some <| SubModelSeq {
           GetModels = getModels
           GetId = getId
@@ -639,7 +625,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
             if not (b.ItemEquals newVal oldVal) then
               b.Values.[oldIdx] <- newVal
           let newVals = intermediate |> b.Map |> Seq.toArray
-          historicalMerge logInvalidGetId logInvalidGetId b.GetId b.GetId create update b.Values newVals
+          elmStyleMerge b.GetId b.GetId create update b.Values newVals
         false
     | Cmd _
     | CmdParam _ ->
@@ -734,14 +720,13 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
             vm.UpdateModel m
             false
     | SubModelSeq b ->
-        let logInvalidGetTargetId a b (vm: ViewModel<_, _>) = logInvalidGetId a b vm.CurrentModel
         let getTargetId (vm: ViewModel<_, _>) = b.GetId vm.CurrentModel
         let create m id = 
           let chain = getPropChainForItem bindingName (id |> string)
           ViewModel(m, (fun msg -> b.ToMsg (id, msg) |> dispatch), b.GetBindings (), config, chain)
         let update (vm: ViewModel<_, _>) m _ = vm.UpdateModel m
         let newSubModels = newModel |> b.GetModels |> Seq.toArray
-        historicalMerge logInvalidGetId logInvalidGetTargetId b.GetId getTargetId create update b.Vms newSubModels
+        elmStyleMerge b.GetId getTargetId create update b.Vms newSubModels
         false
     | SubModelSelectedItem b ->
         b.Get newModel <> b.Get currentModel
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -202,19 +202,33 @@ module internal ViewModelHelpers =
             additions.Remove id |> ignore
             (tIdx, sIdx, t, s) |> List.singleton)
 
-    // actually remove
-    Seq.empty
-    |> Seq.append (removals |> Seq.map (Kvp.value >> fst))
-    |> Seq.append (moves |> Seq.map (fun (tIdx, _, _, _) -> tIdx))
-    |> Seq.sortDescending
-    |> Seq.iter target.RemoveAt
+    let actuallyRemove () =
+      Seq.empty
+      |> Seq.append (removals |> Seq.map (Kvp.value >> fst))
+      |> Seq.append (moves |> Seq.map (fun (tIdx, _, _, _) -> tIdx))
+      |> Seq.sortDescending
+      |> Seq.iter target.RemoveAt
 
-    // actually add
-    Seq.empty
-    |> Seq.append (additions |> Seq.map (fun (Kvp (id, (idx, s))) -> idx, create s id))
-    |> Seq.append (moves |> Seq.map (fun (_, sIdx, t, _) -> sIdx, t))
-    |> Seq.sortBy fst
-    |> Seq.iter target.Insert
+    let actuallyAdd () =
+      Seq.empty
+      |> Seq.append (additions |> Seq.map (fun (Kvp (id, (idx, s))) -> idx, create s id))
+      |> Seq.append (moves |> Seq.map (fun (_, sIdx, t, _) -> sIdx, t))
+      |> Seq.sortBy fst
+      |> Seq.iter target.Insert
+
+    match moves, removals.Count, additions.Count with
+    | [ (tIdx, sIdx, _, _) ], 0, 0 -> // single move
+        target.Move(tIdx, sIdx)
+    | [ (t1Idx, s1Idx, _, _); (t2Idx, s2Idx, _, _) ], 0, 0 when t1Idx = s2Idx && t2Idx = s1Idx-> // single swap
+        let temp = target.[t1Idx]
+        target.[t1Idx] <- target.[t2Idx]
+        target.[t2Idx] <- temp
+    | _, rc, _ when rc = targetCount && rc > 0 -> // remove everything (implies moves = [])
+        target.Clear ()
+        actuallyAdd ()
+    | _ ->
+        actuallyRemove ()
+        actuallyAdd ()
 
     // update moved elements
     moves |> Seq.iter (fun (_, sIdx, t, s) -> update t s sIdx)


cmeeren:
Thanks! It's possible to mark PRs as drafts when submitting - thankfully, GH has now added the capability to do that after creation, too. So I went ahead and marked this as a draft. :)

bender2k14:
I am now satisfied with the diff in this PR.  All the other incidental changes have now been merged in via other PRs.

This PR adds `elmStyleMerge` and modifies `update` for `OneWaySeq` and `SubModelSeq` to use it instead of `historicalMerge`.

Each merge algorithm has cases where it is faster.
- `elmStyleMerge` is faster when the change to the list is "small".  Specifically, it is essentially optimal in the following cases.
    - No change*
    - One element is added
    - One element is removed
    - One element is replaced
    - Any combination of the above that have at least one space between them
    - Two adjacent elements are swapped
    - Many elements are added at the end
    - Many elements are removed at the end
    - Many elements are replaced at the end
- `historicalMerge` is faster when the changes to the list are not "small" _and_ creating a "target" element is expensive.  This is because it will only call `create` when necessary whereas `elmStyleMerge` will start recreating all elements in the `ObservableCollection<>` if the difference is not "small".  A good example of a "non-small" change to the list is reordering.

*The cost of the "No change" case can be avoided for a `OneWaySeqBinding` if its `equals` function returns `true`.  However, there is no such lazy behavior for a `SubModelSeqBinding`, so the high cost of executing any merge algorithm is currently being paid more often than it should be.

I will provide some benchmarking results soon in a future comment.

In the meantime, the performance difference is noticeable via our `OneWaySeq` sample.  When testing on `master`, there is a slight delay between the button click and the new element becoming visible.  When testing the same thing on the branch associated with this PR, the new element instantly becomes visible when the button is clicked.

The runtime difference between these two functions is not the whole story.  Another advantage of `historicalMerge` is that it raises the least possible number of change events.  The more change events, the more time WPF will take to update the UI.  This is essentially the point of the `OneWaySeq` sample (though I think the events involved are different, I think the resulting behavior would be the same).  Telling WPF that every element in the list has changed takes noticeably longer than spending the extra time to raise a small number of specific change events.  I am not sure how to systematically do performance testing of this integration scenario.

In the end though, we should optimize for the small or common cases and be willing to suffer worse performance when large or uncommon cases occur.  I think `elmStyleMerge` achieves a better tradeoff than `historicalMerge`.

bender2k14:
When there is no change to the elements or their order, `elmStyleMerge` is much faster than `historicalMerge`.  Of couse no events are raised by either algorithm in this case.

[Zip of HTML version](https://github.com/elmish/Elmish.WPF/files/4646506/Program.NoChange-report.zip)

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18362.836 (1903/May2019Update/19H1)
Intel Core i7-4700HQ CPU 2.40GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.202
  [Host]     : .NET Core 3.1.4 (CoreCLR 4.700.20.20201, CoreFX 4.700.20.22101), X64 RyuJIT DEBUG  [AttachedDebugger]
  DefaultJob : .NET Core 3.1.4 (CoreCLR 4.700.20.20201, CoreFX 4.700.20.22101), X64 RyuJIT


```
|          Method | Length |          Mean |       Error |        StdDev |        Median | Ratio |
|---------------- |------- |--------------:|------------:|--------------:|--------------:|------:|
|   **ElmStyleMerge** |     **10** |      **1.198 μs** |   **0.0186 μs** |     **0.0248 μs** |      **1.189 μs** |  **0.11** |
| HistoricalMerge |     10 |     10.889 μs |   0.2165 μs |     0.4976 μs |     10.645 μs |  1.00 |
|                 |        |               |             |               |               |       |
|   **ElmStyleMerge** |     **20** |      **1.798 μs** |   **0.0246 μs** |     **0.0205 μs** |      **1.790 μs** |  **0.06** |
| HistoricalMerge |     20 |     29.500 μs |   0.5893 μs |     1.2301 μs |     28.897 μs |  1.00 |
|                 |        |               |             |               |               |       |
|   **ElmStyleMerge** |     **50** |      **3.742 μs** |   **0.0748 μs** |     **0.1424 μs** |      **3.727 μs** |  **0.03** |
| HistoricalMerge |     50 |    137.681 μs |   1.9465 μs |     1.7255 μs |    137.012 μs |  1.00 |
|                 |        |               |             |               |               |       |
|   **ElmStyleMerge** |    **100** |      **6.887 μs** |   **0.1374 μs** |     **0.2987 μs** |      **6.733 μs** |  **0.01** |
| HistoricalMerge |    100 |    517.128 μs |  10.2091 μs |    23.4572 μs |    512.359 μs |  1.00 |
|                 |        |               |             |               |               |       |
|   **ElmStyleMerge** |    **200** |     **12.719 μs** |   **0.1187 μs** |     **0.0991 μs** |     **12.688 μs** | **0.007** |
| HistoricalMerge |    200 |  1,928.130 μs |  32.7418 μs |    59.0402 μs |  1,899.871 μs | 1.000 |
|                 |        |               |             |               |               |       |
|   **ElmStyleMerge** |    **500** |     **31.537 μs** |   **0.6264 μs** |     **1.0637 μs** |     **31.004 μs** | **0.003** |
| HistoricalMerge |    500 | 11,475.965 μs | 115.1450 μs |   102.0731 μs | 11,432.639 μs | 1.000 |
|                 |        |               |             |               |               |       |
|   **ElmStyleMerge** |   **1000** |     **60.805 μs** |   **0.7303 μs** |     **0.5702 μs** |     **60.603 μs** | **0.001** |
| HistoricalMerge |   1000 | 48,070.375 μs | 958.4668 μs | 1,957.8937 μs | 47,079.836 μs | 1.000 |

bender2k14:
When the the order of the elements is reversed, `elmStyleMerge` is still faster when `create` is fast (in this case, essentially implemented as the identity function).

[ZIP of HTML version](https://github.com/elmish/Elmish.WPF/files/4646684/Program.ReverseWithoutDelay-report.zip)

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18362.836 (1903/May2019Update/19H1)
Intel Core i7-4700HQ CPU 2.40GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.202
  [Host]     : .NET Core 3.1.4 (CoreCLR 4.700.20.20201, CoreFX 4.700.20.22101), X64 RyuJIT DEBUG  [AttachedDebugger]
  DefaultJob : .NET Core 3.1.4 (CoreCLR 4.700.20.20201, CoreFX 4.700.20.22101), X64 RyuJIT


```
|          Method | Length |          Mean |         Error |        StdDev |        Median | Ratio |
|---------------- |------- |--------------:|--------------:|--------------:|--------------:|------:|
|   **ElmStyleMerge** |     **10** |      **1.503 μs** |     **0.0299 μs** |     **0.0745 μs** |      **1.466 μs** |  **0.10** |
| HistoricalMerge |     10 |     15.528 μs |     0.2428 μs |     0.2028 μs |     15.448 μs |  1.00 |
|                 |        |               |               |               |               |       |
|   **ElmStyleMerge** |     **20** |      **2.560 μs** |     **0.0458 μs** |     **0.0382 μs** |      **2.562 μs** |  **0.05** |
| HistoricalMerge |     20 |     46.883 μs |     0.3474 μs |     0.2713 μs |     46.891 μs |  1.00 |
|                 |        |               |               |               |               |       |
|   **ElmStyleMerge** |     **50** |      **5.908 μs** |     **0.1172 μs** |     **0.2940 μs** |      **5.751 μs** |  **0.02** |
| HistoricalMerge |     50 |    244.850 μs |     1.9722 μs |     1.6469 μs |    245.433 μs |  1.00 |
|                 |        |               |               |               |               |       |
|   **ElmStyleMerge** |    **100** |     **11.061 μs** |     **0.2195 μs** |     **0.5507 μs** |     **10.791 μs** |  **0.01** |
| HistoricalMerge |    100 |    945.470 μs |    18.7576 μs |    44.9420 μs |    921.279 μs |  1.00 |
|                 |        |               |               |               |               |       |
|   **ElmStyleMerge** |    **200** |     **21.776 μs** |     **0.3232 μs** |     **0.2523 μs** |     **21.739 μs** | **0.006** |
| HistoricalMerge |    200 |  3,501.043 μs |    33.8470 μs |    28.2638 μs |  3,497.167 μs | 1.000 |
|                 |        |               |               |               |               |       |
|   **ElmStyleMerge** |    **500** |     **53.677 μs** |     **0.9562 μs** |     **0.7465 μs** |     **53.414 μs** | **0.002** |
| HistoricalMerge |    500 | 22,207.941 μs |   401.6917 μs |   922.9535 μs | 21,768.681 μs | 1.000 |
|                 |        |               |               |               |               |       |
|   **ElmStyleMerge** |   **1000** |    **105.488 μs** |     **1.0464 μs** |     **0.8169 μs** |    **105.476 μs** | **0.001** |
| HistoricalMerge |   1000 | 88,494.344 μs | 1,746.4804 μs | 2,770.1030 μs | 87,143.483 μs | 1.000 |

bender2k14:
But when the elements are reversed create takes a long time--specifically 1 ms, then `historicalMerge` is faster up through `1000` element (but would eventually become slower).

[ZIP of HTML version](https://github.com/elmish/Elmish.WPF/files/4646718/Program.ReverseWithDelay-report.zip)

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18362.836 (1903/May2019Update/19H1)
Intel Core i7-4700HQ CPU 2.40GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.202
  [Host]     : .NET Core 3.1.4 (CoreCLR 4.700.20.20201, CoreFX 4.700.20.22101), X64 RyuJIT DEBUG  [AttachedDebugger]
  DefaultJob : .NET Core 3.1.4 (CoreCLR 4.700.20.20201, CoreFX 4.700.20.22101), X64 RyuJIT


```
|          Method | Length |             Mean |        Error |       StdDev |           Median |    Ratio | RatioSD |
|---------------- |------- |-----------------:|-------------:|-------------:|-----------------:|---------:|--------:|
|   **ElmStyleMerge** |     **10** |    **155,593.30 μs** |   **710.298 μs** |   **664.414 μs** |    **155,758.05 μs** | **9,569.20** |  **493.04** |
| HistoricalMerge |     10 |         16.15 μs |     0.322 μs |     0.819 μs |         15.96 μs |     1.00 |    0.00 |
|                 |        |                  |              |              |                  |          |         |
|   **ElmStyleMerge** |     **20** |    **309,632.68 μs** | **3,740.118 μs** | **3,498.509 μs** |    **310,834.50 μs** | **6,450.02** |  **186.10** |
| HistoricalMerge |     20 |         47.65 μs |     0.919 μs |     1.289 μs |         47.18 μs |     1.00 |    0.00 |
|                 |        |                  |              |              |                  |          |         |
|   **ElmStyleMerge** |     **50** |    **778,531.12 μs** | **3,542.184 μs** | **3,313.361 μs** |    **779,868.50 μs** | **3,610.61** |  **545.66** |
| HistoricalMerge |     50 |        248.66 μs |     6.888 μs |    20.202 μs |        247.37 μs |     1.00 |    0.00 |
|                 |        |                  |              |              |                  |          |         |
|   **ElmStyleMerge** |    **100** |  **1,560,257.50 μs** | **2,155.219 μs** | **2,015.993 μs** |  **1,560,461.00 μs** | **1,558.34** |   **91.17** |
| HistoricalMerge |    100 |        959.95 μs |    19.128 μs |    54.881 μs |        936.05 μs |     1.00 |    0.00 |
|                 |        |                  |              |              |                  |          |         |
|   **ElmStyleMerge** |    **200** |  **3,122,721.17 μs** | **2,739.147 μs** | **2,562.200 μs** |  **3,123,494.60 μs** |   **871.62** |   **14.07** |
| HistoricalMerge |    200 |      3,583.50 μs |    66.171 μs |    58.659 μs |      3,565.33 μs |     1.00 |    0.00 |
|                 |        |                  |              |              |                  |          |         |
|   **ElmStyleMerge** |    **500** |  **7,810,132.60 μs** | **2,304.521 μs** | **2,155.650 μs** |  **7,810,425.00 μs** |   **438.67** |   **60.37** |
| HistoricalMerge |    500 |     21,471.47 μs |   772.298 μs | 2,178.281 μs |     21,337.40 μs |     1.00 |    0.00 |
|                 |        |                  |              |              |                  |          |         |
|   **ElmStyleMerge** |   **1000** | **15,622,315.15 μs** | **2,232.167 μs** | **2,087.971 μs** | **15,622,890.20 μs** |   **176.32** |    **6.72** |
| HistoricalMerge |   1000 |     87,669.06 μs | 1,461.796 μs | 3,502.370 μs |     85,947.81 μs |     1.00 |    0.00 |

bender2k14:
It is not a dichotomy though.  We could also consider other options, like a hybrid of these two that has the advantages of both.

cmeeren:
Thanks for your thorough and excellent work!

> It is not a dichotomy though. We could also consider other options, like a hybrid of these two that has the advantages of both.

Yes, I was thinking about that. `elmStyleMerge` is better in many common cases, but vastly slower in at least one probably-not-that-uncommon (?) case.

If you have an idea of what a hybrid implementation could look like, specifically one that has the advantages of both and the drawbacks of neither, feel free to experiment! From a user perspective, that is certainly the best option.

Another option is to parametrize relevant bindings to accept an optional parameter telling Elmish.WPF which method to use. (It could be a function parameter, allowing users to pass anything but typically one of the two built-in functions, or it could just be a bool.)

cmeeren:
By the way, before merging, it would be great to have at least a few simple unit tests of the new merge function (whatever form it takes) to ensure the behavior is correct.

bender2k14:
> If you have an idea of what a hybrid implementation could look like, specifically one that has the advantages of both and the drawbacks of neither, feel free to experiment! From a user perspective, that is certainly the best option.

I do.  One difference between [Elm's code](https://github.com/elm/virtual-dom/blob/master/src/Elm/Kernel/VirtualDom.js#L980-L1226) and my port is that my code mutates the `ObservableCollection<>` "right away" but the Elm code doesn't.  Instead, the Elm code records what changes to make, and then makes them later.

For example, suppose that some element has moved from index `i` to index `j` with `i < j` (so the elements before `i` and after `j` have not moved but the elements between `i` and `j` have all shifted down by one spot).  This case and its inverse would cover anything the user could do via a drag-and-drop interface.  If `i = j + 1` (i.e. the element only moved by one spot), then my `elmStyleMerge` code sees this as a swap of two adjacent elements and behaves optimally.  In particular, the elements are swapped in the `ObservableCollection<>` as well; `create` is never called.  On the other hand, if `i < j + 1` (i.e. the element moved by more than one spot), then my `elmStyleMerge` code handles this by both removing the element at index `i` and adding a new instance obtained via `create` at index `j`.

In contrast, the Elm code first records the change as a "remove" when inspecting index `i` but later change this record to a "move" when it inspects index `j`.

Until now, I have purposefully ignore that extra layer of complexity because getting this far was hard enough.  I will go look into this extra layer of complexity now.

bender2k14:
> Another option is to parametrize relevant bindings to accept an optional parameter telling Elmish.WPF which method to use. (It could be a function parameter, allowing users to pass anything but typically one of the two built-in functions, or it could just be a bool.)

That is another option, but I don't think that added complexity is worth it.  I wouldn't expect the user to have a good idea of what a better merge algorithm would be.  We are having a hard enough time getting it right ourselves.

bender2k14:
> before merging (whenever this PR is done), it would be great to have at least a few simple unit tests of the new merge function (whatever form it takes) to ensure the behavior is correct.

The existing tests are also testing `elmStyleMerge` now.  That is why I have been improving the tests so much lately.

After my initial implementation, the tests helped me identify three improvements.  Two were failing tests that caught small bugs.  One was from test code coverage (via my friend who has a subscription to [NCrunch](https://www.ncrunch.net/)) that showed the "add" case was never executed.  I added test code coverage for that case when I expanded the add test (in PR #222) from only adding at the end to adding at a randomly selected spot in the list.

Here are some potential ways I see to improve the relevant tests.
1. The tests that consider many different cases only check if an event of the correct type was raised.  The assertion could be strengthened to verify that the correct number of events were raised and that they contain the correct data.
2. Tests considering the many different cases could also verify that the elements in the `ObservableCollection<>` end up in the correct order.  There are some general tests like this, but I am not completely convinced of their thoroughness.
3. We could add tests specifically for `elmStyleMerge` instead of higher up for `OneWaySeqBinding` or `SubModelSeqBinding`.  Such tests could, for example, count how many times `update` is called and assert that it is the correct number.

Overall, I am currently satisfied with the existing tests, but I can add some if there is something specific that you want tested.

cmeeren:
> Until now, I have purposefully ignore that extra layer of complexity because getting this far was hard enough. I will go look into this extra layer of complexity now.

Excellent, thank you so much!

> That is another option, but I don't think that added complexity is worth it. I wouldn't expect the user to have a good idea of what a better merge algorithm would be. We are having a hard enough time getting it right ourselves.

Totally agree.

> Overall, I am currently satisfied with the existing tests, but I can add some if there is something specific that you want tested.

No, I think it's fine as it is. (Though if you want to extend the tests at some later time, knock yourself out!)

bender2k14:
I improved `elmStyleMerge` by delaying the the additions and removals until the end.  At which point, it searches the removals for elements that were also recorded as an addition.  That way, it only calls `create` if necessary.

I will share the benchmarking comparisons in separate comments below.

bender2k14:
`elmStyleMerge` is still faster when there is no change

[ZIP of NoChange-report.html](https://github.com/elmish/Elmish.WPF/files/4653137/Program.NoChange-report.zip)
``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18362.836 (1903/May2019Update/19H1)
Intel Core i7-4700HQ CPU 2.40GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.202
  [Host]     : .NET Core 3.1.4 (CoreCLR 4.700.20.20201, CoreFX 4.700.20.22101), X64 RyuJIT DEBUG  [AttachedDebugger]
  DefaultJob : .NET Core 3.1.4 (CoreCLR 4.700.20.20201, CoreFX 4.700.20.22101), X64 RyuJIT


```
|          Method | Length |          Mean |       Error |        StdDev |        Median | Ratio | RatioSD |
|---------------- |------- |--------------:|------------:|--------------:|--------------:|------:|--------:|
|   **ElmStyleMerge** |     **10** |      **4.383 μs** |   **0.1767 μs** |     **0.4926 μs** |      **4.169 μs** |  **0.40** |    **0.05** |
| HistoricalMerge |     10 |     11.106 μs |   0.3226 μs |     0.9256 μs |     10.829 μs |  1.00 |    0.00 |
|                 |        |               |             |               |               |       |         |
|   **ElmStyleMerge** |     **20** |      **4.671 μs** |   **0.0917 μs** |     **0.1744 μs** |      **4.607 μs** |  **0.16** |    **0.01** |
| HistoricalMerge |     20 |     29.178 μs |   0.5141 μs |     0.4557 μs |     29.140 μs |  1.00 |    0.00 |
|                 |        |               |             |               |               |       |         |
|   **ElmStyleMerge** |     **50** |      **7.020 μs** |   **0.1899 μs** |     **0.5387 μs** |      **6.874 μs** |  **0.05** |    **0.00** |
| HistoricalMerge |     50 |    133.742 μs |   2.6639 μs |     4.6656 μs |    132.588 μs |  1.00 |    0.00 |
|                 |        |               |             |               |               |       |         |
|   **ElmStyleMerge** |    **100** |     **10.264 μs** |   **0.3020 μs** |     **0.8061 μs** |     **10.064 μs** |  **0.02** |    **0.00** |
| HistoricalMerge |    100 |    594.034 μs |  28.3192 μs |    82.1591 μs |    567.486 μs |  1.00 |    0.00 |
|                 |        |               |             |               |               |       |         |
|   **ElmStyleMerge** |    **200** |     **16.745 μs** |   **0.4252 μs** |     **1.2267 μs** |     **16.398 μs** | **0.009** |    **0.00** |
| HistoricalMerge |    200 |  1,926.195 μs |  46.0642 μs |   131.4236 μs |  1,892.110 μs | 1.000 |    0.00 |
|                 |        |               |             |               |               |       |         |
|   **ElmStyleMerge** |    **500** |     **34.684 μs** |   **0.6881 μs** |     **1.3257 μs** |     **34.338 μs** | **0.003** |    **0.00** |
| HistoricalMerge |    500 | 11,650.765 μs | 231.7274 μs |   572.7728 μs | 11,389.890 μs | 1.000 |    0.00 |
|                 |        |               |             |               |               |       |         |
|   **ElmStyleMerge** |   **1000** |     **67.180 μs** |   **1.4632 μs** |     **4.2216 μs** |     **64.927 μs** | **0.001** |    **0.00** |
| HistoricalMerge |   1000 | 45,385.228 μs | 817.8480 μs | 1,595.1497 μs | 44,711.045 μs | 1.000 |    0.00 |

bender2k14:
`elmStyleMerge` is essentially still faster when the elements are reversed and `create` is implemented as the identity function.

[ZIP of ReverseWithoutDelay-report.html](https://github.com/elmish/Elmish.WPF/files/4653158/Program.ReverseWithoutDelay-report.zip)

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18362.836 (1903/May2019Update/19H1)
Intel Core i7-4700HQ CPU 2.40GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.202
  [Host]     : .NET Core 3.1.4 (CoreCLR 4.700.20.20201, CoreFX 4.700.20.22101), X64 RyuJIT DEBUG  [AttachedDebugger]
  DefaultJob : .NET Core 3.1.4 (CoreCLR 4.700.20.20201, CoreFX 4.700.20.22101), X64 RyuJIT


```
|          Method | Length |         Mean |        Error |       StdDev |       Median | Ratio | RatioSD |
|---------------- |------- |-------------:|-------------:|-------------:|-------------:|------:|--------:|
|   **ElmStyleMerge** |     **10** |     **17.50 μs** |     **0.347 μs** |     **0.341 μs** |     **17.44 μs** |  **1.09** |    **0.06** |
| HistoricalMerge |     10 |     15.91 μs |     0.314 μs |     0.702 μs |     15.62 μs |  1.00 |    0.00 |
|                 |        |              |              |              |              |       |         |
|   **ElmStyleMerge** |     **20** |     **29.71 μs** |     **0.289 μs** |     **0.226 μs** |     **29.68 μs** |  **0.64** |    **0.01** |
| HistoricalMerge |     20 |     46.56 μs |     0.494 μs |     0.412 μs |     46.55 μs |  1.00 |    0.00 |
|                 |        |              |              |              |              |       |         |
|   **ElmStyleMerge** |     **50** |     **65.17 μs** |     **1.081 μs** |     **0.844 μs** |     **65.16 μs** |  **0.26** |    **0.01** |
| HistoricalMerge |     50 |    250.62 μs |     4.993 μs |    11.768 μs |    247.38 μs |  1.00 |    0.00 |
|                 |        |              |              |              |              |       |         |
|   **ElmStyleMerge** |    **100** |    **131.27 μs** |     **2.602 μs** |     **4.051 μs** |    **129.86 μs** |  **0.14** |    **0.01** |
| HistoricalMerge |    100 |    957.53 μs |    18.588 μs |    51.197 μs |    947.36 μs |  1.00 |    0.00 |
|                 |        |              |              |              |              |       |         |
|   **ElmStyleMerge** |    **200** |    **251.11 μs** |     **3.142 μs** |     **2.453 μs** |    **250.79 μs** |  **0.06** |    **0.00** |
| HistoricalMerge |    200 |  3,819.11 μs |   111.243 μs |   320.961 μs |  3,688.24 μs |  1.00 |    0.00 |
|                 |        |              |              |              |              |       |         |
|   **ElmStyleMerge** |    **500** |    **637.47 μs** |    **12.367 μs** |    **20.662 μs** |    **631.40 μs** |  **0.03** |    **0.00** |
| HistoricalMerge |    500 | 23,258.87 μs |   569.190 μs | 1,669.336 μs | 22,498.70 μs |  1.00 |    0.00 |
|                 |        |              |              |              |              |       |         |
|   **ElmStyleMerge** |   **1000** |  **1,312.55 μs** |    **26.143 μs** |    **57.385 μs** |  **1,301.16 μs** |  **0.01** |    **0.00** |
| HistoricalMerge |   1000 | 89,762.44 μs | 2,014.715 μs | 5,616.221 μs | 87,476.48 μs |  1.00 |    0.00 |

bender2k14:
Since neither `elmStyleMerge` nor `historicalMerge` call `create` when the elements are reversed, this benchmark comparison, in which `create` is implemented with a `1` ms delay, is now identical the previous benchmark comparison.

Unlike in the previous benchmark comparison, `elmStyleMerge` was measured to be strictly faster than `historicalMerge` for all parameter values...including `10`.  Either way, the runtime for both algorithms for parameter value `10` is in the fractions of a nano second, so the variance there is not very significant.

[ZIP of ReverseWithDelay-report.html](https://github.com/elmish/Elmish.WPF/files/4653181/Program.ReverseWithDelay-report.zip)

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18362.836 (1903/May2019Update/19H1)
Intel Core i7-4700HQ CPU 2.40GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.202
  [Host]     : .NET Core 3.1.4 (CoreCLR 4.700.20.20201, CoreFX 4.700.20.22101), X64 RyuJIT DEBUG  [AttachedDebugger]
  DefaultJob : .NET Core 3.1.4 (CoreCLR 4.700.20.20201, CoreFX 4.700.20.22101), X64 RyuJIT


```
|          Method | Length |         Mean |        Error |       StdDev |       Median | Ratio | RatioSD |
|---------------- |------- |-------------:|-------------:|-------------:|-------------:|------:|--------:|
|   **ElmStyleMerge** |     **10** |     **17.25 μs** |     **0.292 μs** |     **0.244 μs** |     **17.26 μs** |  **0.92** |    **0.04** |
| HistoricalMerge |     10 |     18.94 μs |     0.374 μs |     0.644 μs |     18.82 μs |  1.00 |    0.00 |
|                 |        |              |              |              |              |       |         |
|   **ElmStyleMerge** |     **20** |     **30.99 μs** |     **0.619 μs** |     **1.797 μs** |     **30.24 μs** |  **0.64** |    **0.03** |
| HistoricalMerge |     20 |     48.50 μs |     0.565 μs |     0.472 μs |     48.41 μs |  1.00 |    0.00 |
|                 |        |              |              |              |              |       |         |
|   **ElmStyleMerge** |     **50** |     **69.00 μs** |     **1.673 μs** |     **4.692 μs** |     **67.51 μs** |  **0.27** |    **0.02** |
| HistoricalMerge |     50 |    257.61 μs |     5.424 μs |    15.474 μs |    252.42 μs |  1.00 |    0.00 |
|                 |        |              |              |              |              |       |         |
|   **ElmStyleMerge** |    **100** |    **131.54 μs** |     **2.611 μs** |     **5.677 μs** |    **130.12 μs** |  **0.13** |    **0.01** |
| HistoricalMerge |    100 |  1,015.06 μs |    31.823 μs |    89.236 μs |    989.71 μs |  1.00 |    0.00 |
|                 |        |              |              |              |              |       |         |
|   **ElmStyleMerge** |    **200** |    **259.60 μs** |     **5.092 μs** |    **12.869 μs** |    **254.16 μs** |  **0.07** |    **0.00** |
| HistoricalMerge |    200 |  3,574.66 μs |    56.403 μs |    82.675 μs |  3,550.49 μs |  1.00 |    0.00 |
|                 |        |              |              |              |              |       |         |
|   **ElmStyleMerge** |    **500** |    **615.94 μs** |     **4.649 μs** |     **3.882 μs** |    **615.45 μs** |  **0.03** |    **0.00** |
| HistoricalMerge |    500 | 21,639.86 μs |   361.185 μs |   281.989 μs | 21,539.92 μs |  1.00 |    0.00 |
|                 |        |              |              |              |              |       |         |
|   **ElmStyleMerge** |   **1000** |  **1,323.94 μs** |    **26.337 μs** |    **67.513 μs** |  **1,306.11 μs** |  **0.02** |    **0.00** |
| HistoricalMerge |   1000 | 88,676.72 μs | 1,742.355 μs | 4,106.935 μs | 86,620.66 μs |  1.00 |    0.00 |

bender2k14:
I am trying to think of any more ways in which `elmStyleMerge` can be improved or specifically of cases in which `historicalMerge` is better.

One way in which `historicalMerge` is better than `elmStyleMerge` is its use of `ObservableCollection<>.MoveTo(int, int)`.  This would raise one change event while `elmStyleMerge` would currently handle this case with a call to `RemoveAt` and then one to `Insert`, each of which raises one event for a total of two raised events.  The runtime of `MoveTo` is also never worse than a pair of calls to `RemoveAt` and `Insert`, but all three are linear time in the worst case.

`elmStyleMerge` could be improved in the case that the data as been completely refreshed (i.e. no IDs match between the existing elements and the new elements).  `historicalMerge` does that in a special case of that.

Another case in which `elmStyleMerge` could be improved is when a pair of elements are swapped (but those elements are not adjacent), but not by much....and this case doesn't seem all that common to me.

cmeeren:
Thanks for your excellent work! 👍 The way I understand it, you want to work some more on it before merging?

bender2k14:
Sure, let's put one more day into this.  I can make improvements for all three of those cases.

bender2k14:
> Another advantage of `historicalMerge` is that it raises the least possible number of change events [related to ordering].

Oh, that is not true.

What I was trying to say there was that the number of `CollectionChanged` events also matters, and since `historicalMerge` uses `MoveTo` extensively, it could raise fewer `CollectionChanged` events (because one event is raised by `MoveTo` or achieving the same behavior via `RemoveAt` and `Insert` raises two events).

The reason `historicalMerge` doesn't necessary raise the least number of change events related to ordering is because (in pseudo code)
- `[ b; c; a ] |> moveTo 2 0` and
- `[ b; c; a ] |> moveTo 1 2 |> moveTo 0 1`

both yield `[ a; b; c ]` but they raise a different number of `CollectionChanged` events.

bender2k14:
> One way in which `historicalMerge` is better than `elmStyleMerge` is its use of `ObservableCollection<>.MoveTo(int, int)`.  This would raise one change event while `elmStyleMerge` would currently handle this case with a call to `RemoveAt` and then one to `Insert`, each of which raises one event for a total of two raised events.  The runtime of `MoveTo` is also never worse than a pair of calls to `RemoveAt` and `Insert`, but all three are linear time in the worst case.
> 
> `elmStyleMerge` could be improved in the case that the data as been completely refreshed (i.e. no IDs match between the existing elements and the new elements).  `historicalMerge` does that in a special case.
> 
> Another case in which `elmStyleMerge` could be improved is when a pair of elements are swapped (but those elements are not adjacent), but not by much....and this case doesn't seem all that common to me.

It was easy to optimize for these cases.  [Here is the code](https://github.com/bender2k14/Elmish.WPF/blob/6830e86f1511a63f3a6dedc7a71f533af233beff/src/Elmish.WPF/ViewModel.fs#L219-L231) from the most recent commit that does so.

There are explicit tests for the move and swap cases.  The remove all case is likely being executed in one of the tests that merges in an arbitrary array (even though it was not required to be disjoint with the original items).  If there is any concern, then we should just add an explicit test for this case.  Either way, I will get an updated report on test code coverage tomorrow when I speak with my NCrunch friend.

I have not written benchmark comparisons related to these optimizations.  I think it is easy to see that the performance could hardly be worse.  The extra work when not in any of those cases is a few predicate evaluations.

I am satisified with this branch.

cmeeren:
Excellent, thanks! Merge when you're ready.

> Either way, I will get an updated report on test code coverage tomorrow when I speak with my NCrunch friend.

If you want, I can get you an OSS license for ReSharper Ultimate (including dotCover). They give out free licenses (1 year at a time) to regular contributors on active OSS projects. I had one through this project (and used dotTrace a bit for performance profiling), but the license just expired and I didn't renew it because I don't really use it now. It's easy to renew, though, so let me know if you want it.

cmeeren:
After merging, should we release right away? Mentioning performance improvements in the changelog? Could you tell me, in release-note-friendly terms, where the improvements are relevant?

bender2k14:
> I will get an updated report on test code coverage tomorrow when I speak with my NCrunch friend.

Test code coverage is good.  Every line of `elmStyleMerge` is executed.

> If you want, I can get you an OSS license for ReSharper Ultimate (including dotCover). They give out free licenses (1 year at a time) to regular contributors on active OSS projects. I had one through this project...

Oh, that is good to know.  I will look into that at some point.  Thanks! :)

> After merging, should we release right away?

Yes, I think we should release.

> Mentioning performance improvements in the changelog?

Yes, there is no API change and no bug fixes, just performance improvements.

> Could you tell me, in release-note-friendly terms, where the improvements are relevant?

The amount of time used to update `OneWaySeq` and `SubModelSeq` bindings has been significantly decreased.  This includes all cases of a `SubModelSeq` binding and all cases of a `OneWaySeq` binding for which `equals` returns `false`.

cmeeren:
Thanks! Excellent release notes. v3.5.6 is in the pipeline now.

PR #217, opened 2020-05-17
Remove one call to `get`

bender2k14:

I think this code should only call `get` twice, once for the current model and once for the new model.

The previous code called `get` three times, once for the current model and _twice_ for the new model.

This PR saves the return value from the first time the new model is passed into `get` and uses that in place of another call to `get`.


bender2k14:
(The force push is just a rebase onto `master`)

I do want this change.

To be honest, I think of this as more of a separation between the lazy behavior and the seq behavior than a performance optimization.  If I remember correctly, I previously tried (and failed) to extract the lazy behavior into its own binding effect.  I think I am more likely to succeed if I tried again now that `Binding.mapModel` exists.

PR #220, opened 2020-05-18
Improve test names

bender2k14:

This PR primarily improves test names.  It also splits one test in two (because after which, the name each name is easier to understand).  It also removes two more type annotations I missed before (after extracting the corresponding functions from within a quoted expression).


cmeeren:
Thanks!

PR #221, opened 2020-05-18
Unify OneWaySeq and SubModel merges

bender2k14:

This PR extracts the aspects of `oneWaySeqMerge` and `SubModelSeqMerge` that differ.

The next PR in this sequence will rename identifiers in these methods to confirm they are really the same function by observing they have exactly the same source code.  Then the PR after that will delete one of these functions and keep the other.


cmeeren:
Thanks! :)

PR #222, opened 2020-05-18
Test any permutation of elements

bender2k14:

Two existing tests tested update after reversing the order of the elements.  This PR increases the scope of that test using `GenX.shuffle` to consider any permutation of the elements instead of just the reversal (though the identity permutation is excluded).


cmeeren:
Thanks a lot!

PR #224, opened 2020-05-18
adds tests that replace an element

bender2k14:

This PR adds two tests (one for `OneWaySeqBinding` and one for `SubModelSeqBinding`) that replaces an element and then calls `update`.


cmeeren:
Thanks for all your excellent work! 😀

PR #226, opened 2020-05-18
move historicalMerge out of ViewModel

bender2k14:

This PR moves `historicalMerge` out of ViewModel into its own _`internal`_ module in the same file.  This makes it easier to test since now an instance of `ViewModel<,>` is not required.

### DIFF
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -10,6 +10,65 @@ open System.Windows
 open Elmish
 
 
+[<AutoOpen>]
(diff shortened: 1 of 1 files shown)


cmeeren:
You know what, you're doing such a thorough job (with these PRs and otherwise) that I trust you and am barely looking at the changes at all before merging all this stuff anyway. 😛 I don't know how much is left, but I think that you can just go ahead and merge this and other similar uncontroversial PRs and let me know if there is something specific you want me to look at. 👍

bender2k14:
Will do :)

Issue #105, opened 2019-08-25
Potential logging improvements


bender2k14:
I have some ideas about how to improve logging.  I think they will lay a nice foundation to achieve the logging goals that @cmeeren has in mind (as briefly mentioned in #84).

### 1. Ability to send logs anywhere

Currently Elmish.WPF only allows its log events to be sent to [`System.Console`](https://github.com/elmish/Elmish.WPF/blob/master/src/Elmish.WPF/ViewModel.fs#L78) or [`System.Diagnostics.Trace`](https://github.com/elmish/Elmish.WPF/blob/master/src/Elmish.WPF/ViewModel.fs#L79).  Similar to [`Elmish.Program.withTrace`](https://github.com/elmish/elmish/blob/master/src/program.fs#L78), I suggest that Elmish.WPF accept a callback to allow users to send logs to a location of their choice.  Let's call such a location a logging [sink](https://github.com/serilog/serilog/wiki/Provided-Sinks).

I still think it is a good idea to have functionality like [`ElmConfig.LogConsole`](https://github.com/elmish/Elmish.WPF/blob/master/src/Elmish.WPF/Config.fs#L5) and [`ElmConfig.LogTrace`](https://github.com/elmish/Elmish.WPF/blob/master/src/Elmish.WPF/Config.fs#L7) in the sense that it is helpful to users to be able to trivially employ those two logging sinks.  However, in terms of implementation, I would prefer if it were like [`Emlish.Program.withConsoleTrace`](https://github.com/elmish/elmish/blob/master/src/program.fs), which creates a new record instance with the appropriate change instead of requiring users to create a new record themselves.

Whatever API is used, I strongly suggest that the build-in support for logging to `System.Console` and `System.Diagnostics.Trace` be implemented via the the previously suggested callback API.  This ensures that these build-in logging sinks are merely (very helpful) syntactic sugar for users instead of taking advantage of their proximity to Elmish.WPF and logging additional information not given to the logging callback (which is currently the situation with Elmish and something I intend to point out in a future issue there).

### 2. Use strong types to communicate all log event data

Currently Elmish.WPF handles each log event by sending a computed `string` to `System.Console.WriteLine` or `System.Diagnostics.Trace.WriteLine`.  For the user-supplied logging callback discussed above, I suggest its input be a discriminated union with cases for each type of log event and containing all the relevant data.  This gives users full control to employ techniques such as [structured logging](https://github.com/serilog/serilog/wiki/Structured-Data).

[...]

PR #235, opened 2020-05-31
DRAFT: Replace ElmConfig with ProgramWpf

bender2k14:
This _draft_ PR is initial work towards resolving #105.  See especially https://github.com/elmish/Elmish.WPF/issues/105#issuecomment-635701641.

This PR contains two commits.  The first commit changes the indentation in most of Program.fs, so it is helpful to review each commit separately while ignoring whitespace changes.

The first commit makes the `Program` module into a module within the namespace `Elmish.WPF`.

The second commit replaces `ElmConfig` with `ProgramWpf`.  The constructor of `ProgramWpf` is private and public setters/mapers are provided, just like Elmish does with its `Program` record.  The `ProgramWpf` record contains the `Elmish.Program` record.  Modifications to this underlying `Elmish.Prgoram` record are allowed via `Program.mapElmishProgram`.

I tried to make the second commit as small as possible to help emphasize the change to the API.  There are many other internal changes that would naturally follow that I have excluded for now.

What are you thoughts on this approach where the `ProgramWpf` record contains the `Elmish.Program` record?

As for naming and the public API, one change that I think would be nice is to rename
- the module `Program` module to `WpfProgram`,
- the record `ProgramWpf` to `WpfProgram`,
- the function `mkProgramWpf` to `mkWpfProgram`, and
- the function `mkProgramWpfWithCmdMsg` to `mkWpfProgramWithCmdMsg`.

This would help distinguish which functions take `Elmish.Program` as input and which ones take `Elmish.WPF.WpfProgram` as input...while also putting `Wpf` before `Program`, which is the natural placement of the adjective `Wpf` of the noun `Program`.  Of course we would deprecate instead of remove the existing functions and module to facilitate migration.

### DIFF
--- a/src/Elmish.WPF/Program.fs
+++ b/src/Elmish.WPF/Program.fs
@@ -4,6 +4,12 @@ open System.Windows
 open Elmish
 
 
+type ProgramWpf<'arg, 'model, 'msg> = private {
+  ElmishProgram: Program<'arg, 'model, 'msg, Binding<'model, 'msg> list>
+  ElmConfig: ElmConfig
+}
+
+
 [<RequireQualifiedAccess>]
 module Program =
 
@@ -11,16 +17,15 @@ module Program =
   /// for the specified FrameworkElement. Non-blocking. This is a low-level function;
   /// for normal usage, see runWindow and runWindowWithConfig.
   let startElmishLoop
-      (config: ElmConfig)
       (element: FrameworkElement)
-      (program: Program<unit, 'model, 'msg, Binding<'model, 'msg> list>) =
+      (program: ProgramWpf<unit, 'model, 'msg>) =
     let mutable lastModel = None
   
     let setState model dispatch =
       match lastModel with
       | None ->
-          let bindings = Program.view program model dispatch
-          let vm = ViewModel<'model,'msg>(model, dispatch, bindings, config, "main")
+          let bindings = Program.view program.ElmishProgram model dispatch
+          let vm = ViewModel<'model,'msg>(model, dispatch, bindings, program.ElmConfig, "main")
           element.DataContext <- vm
           lastModel <- Some vm
       | Some vm ->
@@ -29,7 +34,7 @@ module Program =
     let uiDispatch (innerDispatch: Dispatch<'msg>) : Dispatch<'msg> =
       fun msg -> element.Dispatcher.Invoke(fun () -> innerDispatch msg)
   
-    program
+    program.ElmishProgram
     |> Program.withSetState setState
     |> Program.withSyncDispatch uiDispatch
     |> Program.run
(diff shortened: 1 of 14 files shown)


cmeeren:
I might not be able to have a good look at this for a while (at least next weekend), but one thing I noticed right away is that we may want to duplicate all (or at least the most common) Elmish `Program.` functions, e.g. `Program.withSubscription` and `Program.withConsoleTrace` (but perhaps still keep `Program.mapElmishProgram` as a workaround for future Elmish additions, and to use in the implementation of the duplicates).

bender2k14:
Yes, great suggestion.  In addition to providing `mapElmishProgram`, I like the idea of duplicating all the `Elmish.Program` functions.  There are not that many such functions and it does make the API much better.

I pushed a commit that makes this change for the two `Elmish.Program` functions used by the samples.

src/Elmish.WPF/Program.fs
```diff
@@ -1,101 +1,129 @@
...
+
+[<RequireQualifiedAccess>]
+module Program =
+
+  /// Starts the Elmish dispatch loop, setting the bindings as the DataContext
+  /// for the specified FrameworkElement. Non-blocking. This is a low-level function;
+  /// for normal usage, see runWindow and runWindowWithConfig.
+  let startElmishLoop
+      (element: FrameworkElement)
+      (program: ProgramWpf<unit, 'model, 'msg>) =
```
2sComplement, line 21: The `'arg` type in `Progrram` is used in traditional Elmish apps to pass an initial state to the top-level `init` function, typically from a URL parser. This doesn't apply to Elmish.WPF applications. hence the reason `'arg` is always `unit`. With `ProgramWpf` now abstracting `Program`, is it worth excluding `'arg` from the definition of `ProgramWpf`?

bender2k14: I am definitely in favor of this.  Thanks for the suggestion.

Dropping the `'arg` type was also considered for `Elmish.Program` https://github.com/elmish/elmish/issues/138, but it didn't go anywhere.

Do you know of an example of a traditional Elmish/Fable app that passes initial state into the top-level `init` function?  I would like to try a refactor attempt like Don Syme did.

bender2k14: I pushed a commit with this change

2sComplement: Are you referring to the refactor in elmish/elmish#138? The thing is, I don't think we will even need a non-unit argument passed to `init`, will we? The only use-case I've ever seen in Elmish is when a URL is parsed during state initialization.

The [sample-react-navigation](https://github.com/elmish/sample-react-navigation) shows a typical example, specifically [here](https://github.com/elmish/sample-react-navigation/blob/master/src/app.fs#L73).

bender2k14: That file was edited slightly.  Here is the [corresponding permalink](https://github.com/elmish/sample-react-navigation/blob/c5fe721b29ffcf4e75ae55d7f718495fe0d705d5/src/app.fs#L73).

bender2k14:
This change is now being made in PR #255

Issue #237, opened 2020-06-30
Tests in master failing due to MissingMethodException


bender2k14:
There are four tests in `master` failing because due to a `MissingMethodException`.

> System.MissingMethodException : Method not found: 'System.Double System.Numerics.BigInteger.op_Explicit(System.Numerics.BigInteger)'

Below I have included the full test detail summary for all four tests.  I am troubleshooting this issue now.  I will comment again soon with more information.

1)
```
Test Name:	Elmish.WPF.Tests.MergeTests.starting with random items, when merging after shuffling, should contain the merged items and never call create and call update eactly once for each item
Test FullName:	Elmish.WPF.Tests.Elmish.WPF.Tests.MergeTests.Elmish.WPF.Tests.MergeTests.starting with random items, when merging after shuffling, should contain the merged items and never call create and call update eactly once for each item
Test Source:	C:\Code\ThirdParty\Elmish.WPF\src\Elmish.WPF.Tests\MergeTests.fs : line 207
Test Outcome:	Failed
Test Duration:	0:00:00

Test Name:	Elmish.WPF.Tests.MergeTests.starting with random items, when merging after shuffling, should contain the merged items and never call create and call update eactly once for each item
Test Outcome:	Failed
Result StackTrace:	
at Elmish.WPF.Tests.MergeTests.starting with random items\, when merging after shuffling\, should contain the merged items and never call create and call update eactly once for each item@208-2.Invoke(Int32 n0)
   at Elmish.WPF.Tests.MergeTests.starting with random items\, when merging after shuffling\, should contain the merged items and never call create and call update eactly once for each item@208-1.Invoke(Int32 sz)
   at Hedgehog.Gen.List@338-1.Invoke(Seed seed, Int32 size)
   at Hedgehog.Random.bind@45.Invoke(Seed seed, Int32 size)
   at Hedgehog.Gen.bindRandom@40.Invoke(Seed seed0, Int32 size)
   at Hedgehog.Property.loop@332-21(Int32 n, Random`1 random, Seed seed, Int32 size, Int32 tests, Int32 discards)
   at Hedgehog.Property.Report(Int32 n, Property`1 p)
   at Hedgehog.Property.Check(Property`1 p)
[...]

PR #240, opened 2020-06-30
Replaced calls to Range.exponential with Range.constant to avoid issue #237

bender2k14:
Replacing calls to `Range.exponential` with `Range.constant` seems like an easy way to avoid issue #237.  For this branch, all the tests pass on my machine, which is using MSBuild 16.6.0.  Since AppVeyor is still using version 16.5.0, I expect all tests to pass for both the current state of `master` as well as for this branch.

I am not confident in my understanding of the difference between `Range.exponential` and `Range.constant`.

Does `Range.constant x y` model the ([discrete](https://en.wikipedia.org/wiki/Discrete_uniform_distribution) or [continuous](https://en.wikipedia.org/wiki/Uniform_distribution_(continuous))) uniform distribution  and `Range.exponential` model either the [(discrete) geometric distribution](https://en.wikipedia.org/wiki/Geometric_distribution) or the [(continuous) exponential distribution](https://en.wikipedia.org/wiki/Exponential_distribution)?

### DIFF
--- a/src/Elmish.WPF.Tests/MergeTests.fs
+++ b/src/Elmish.WPF.Tests/MergeTests.fs
@@ -162,7 +162,7 @@ let ``starting with random items, when merging after a replacement, should conta
 [<Fact>]
 let ``starting with random items, when merging after swapping two adjacent items, should contain the merged items and never call create and call update exactly once for each item`` () =
   Property.check <| property {
-    let! list1 = Gen.guid |> Gen.list (Range.exponential 2 50)
+    let! list1 = Gen.guid |> Gen.list (Range.constant 2 50)
     let! firstSwapIndex = (0, list1.Length - 2) ||> Range.constant |> Gen.int
 
     let observableCollection = ObservableCollection<_> list1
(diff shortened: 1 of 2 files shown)


cmeeren:
This is better asked in the [Hedgehog repo](https://github.com/hedgehogqa/fsharp-hedgehog). `Range.constant` is, as the name implies, constant (i.e. any test in the run, whether the first or last test, has an equal chance of picking any value from the range, whereas for `Range.exponential`, early tests in the run will pick values closer to the range's "zero" the later tests will pick values from the whole range, and the "amount of values" available to be picked from increases exponentially with the test repeat number. In other words, compared to `Range.linear`, `Range.exponential` will pick more "low" values. This increases performance if used as the size for lists etc., because there's less data to generate.

cmeeren:
If the switch from `Range.exponential` to `Range.constant` fixes a MissingMethodException, then that should probably be reported as a bug in the Hedgehog repo.

bender2k14:
I have reported the issue to Hedgehog [here](https://github.com/hedgehogqa/fsharp-hedgehog/issues/199).  In the meantime, let's go with the workaround in this PR.

> This increases performance if used as the size for lists etc., because there's less data to generate.

With this change to use `Range.constant`, I can run all our tests on my machine in less than 22 seconds.  So if it really is just a performance consideration, then I think we can also stick with this workaround forever.

Issue #236, opened 2020-06-15
Trees done right


bender2k14:
This is a spin off of https://github.com/elmish/Elmish.WPF/issues/143#issuecomment-644371235.

That issue is about adding lazy behavior to the `SubModelSeq` binding (i.e. what the `OneWaySeq` binding calls `itemEquals`).  Even after that lazy behavior is implemented, there might still be an issue for trees.

To gain a better understanding of the potential issue, I will modify the `SubModelSeq` sample, which involves trees, to use  _explicit_ trees.  By "explicit tree", I mean a tree in which each node contains direct references to its children.  To achieve this, I expect to add an extra layer in the XAML to map from the tree node containing the counter to the counter itself....much like how a `SubModel` binding looks in the XAML.

I briefly searched for how to create a tree in ELM, but I only found independent solutions to correctly and efficiently handle the case of trees, so this makes me think that ELM doesn't directly support trees.

ScottHutchinson:
And googling for "elm trees" yields more results botanical than software. ;o)

cmeeren:
> Even after that lazy behavior is implemented, there might still be an issue for trees.

Could you elaborate?

bender2k14:
Yes, it is still relevant.

First, https://github.com/elmish/Elmish.WPF/issues/143#issuecomment-641532812 is about the `SubModelSeq` sample being slow when the tree becomes large.  It is slow even when the tree is flat.  To see this, test on the branch [testing/large_TreeView_slow](https://github.com/bender2k14/Elmish.WPF/tree/testing/large_TreeView_slow).  It is slow because of code in the sample.

Second, I hastily implemented the "lazy"-style behavior as described in #143.  Then I modified the sample to use the lazy behavior with the `refEq` function.  See the branch [testing/large_TreeView_slow_optimize_Elmish.WPF](https://github.com/bender2k14/Elmish.WPF/tree/testing/large_TreeView_slow_optimize_Elmish.WPF).  The resulting code is fast, but the behavior is wrong.  The short circuit of update is incorrect because it ignores changes in the children.  In order to see a change in a child, there has to be a change in its parent, grandparent, and so on all the way up to a top-level counter.  See this GIF for the behavior in the simplest case of a child counter with top-level parent counter.

![2020-08-13_22-45-44_248](https://user-images.githubusercontent.com/34664007/90212253-80a15880-ddb8-11ea-9813-a4cbe8e3678e.gif)

[...]

PR #241, opened 2020-07-22
SubModelSeq update logic now more functional

bender2k14:
This PR only changes code in the `SubModelSeq` sample.

At a high level, the changes are to use more higher-order functions in order to be both more succinct and have more reusable code.

My intention is for the behavior to be unchanged.  I also expect the performance to be the same.  These changes are _not_ trying to address #236.

I tested this, and I think everything still works the same.

@ScottHutchinson, I think you will be interested in these changes since [you are using a `SubModelSeq` binding](https://github.com/elmish/Elmish.WPF/issues/143#issuecomment-641532812).

### DIFF
--- a/src/Samples/SubModelSeq/Program.fs
+++ b/src/Samples/SubModelSeq/Program.fs
@@ -4,6 +4,21 @@ open System
 open Elmish
 open Elmish.WPF
 
+
+module Func =
+
+  let flip f b a = f a b
+
+  let flatten f a = f a a
+
+  let applyIf p f a =
+    if p a then f a else a
+
+
+let map get set f a =
+  a |> get |> f |> Func.flip set a
+
+
 module List =
 
   let swap i j =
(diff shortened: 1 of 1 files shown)


src/Samples/SubModelSeq/Program.fs
```diff
@@ -4,6 +4,24 @@ open System
...
 
+
+let mapIf p f a =
+  if p a then f a else a
+
+
+module Func =
+
+  let flip<'a, 'b, 'c> (f: 'a -> 'b -> 'c) =
+    fun b a -> f a b
```
cmeeren, line 15: This can be simplified to

```suggestion
  let flip f a b = f b a
```

but given that it's only used a single place, I suggest just removing it and rewriting the relevant code.

src/Samples/SubModelSeq/Program.fs
```diff
@@ -4,6 +4,24 @@ open System
...
+
+  let dedup (f: 'a -> 'a -> 'b) =
+    fun a -> f a a
```
cmeeren, line 18: I find this confusing, and it's only used once. I suggest removing and rewriting relevant code.

src/Samples/SubModelSeq/Program.fs
```diff
@@ -13,177 +31,199 @@ module List =
...
+    module Value =
+      let get m = m.Value
+      let set v m = { m with Value = v }
+      let map f = f |> map get set
+      let mapIf p f = f |> map |> mapIf p
+      let mapIfId id f = f |> mapIf (fun m -> m.Id = id)
+  
+    let create v id =
+      { Id = id
+        Value = v }
```
cmeeren, line 61: I don't think this function adds anything. It's just as easy to create the type without it.

```suggestion
```

src/Samples/SubModelSeq/Program.fs
```diff
@@ -13,177 +31,199 @@ module List =
...
+  module DummyRoot =
+    let get m = m.DummyRoot
+    let set v m = { m with DummyRoot = v }
+    let map f = f |> map get set
 
-  let asDummyRoot ns : Tree.Node<Counter> =
-    { Data = Counter.create() // Placeholder data to satisfy type system. User never sees this.
-      Children = ns }
+  let createIdentifiableCounter () =
+    () |> Guid.NewGuid |> Identifiable.create Counter.init
```
cmeeren, line 156: I don't like this kind of "pipe-o-mania".

```suggestion
    { Id = Guid.NewGuid ()
      Value = Counter.init }
```

src/Samples/SubModelSeq/Program.fs
```diff
@@ -13,177 +31,199 @@ module List =
...
+      DummyRoot =
+        createIdentifiableCounter () // Placeholder data to satisfy type system. User never sees this.
+        |> RoseTree.asLeaf
+        |> RoseTree.addChild (createIdentifiableCounter ()) }
+
+  let hasId id n = n.Data.Id = id
+
+  let addChildCounter nId =
+    nId
+    |> Identifiable.create Counter.init
```
cmeeren, line 169:
```suggestion
    { Id = nId
      Value = Counter.init }
```

src/Samples/SubModelSeq/Program.fs
```diff
@@ -13,177 +31,199 @@ module List =
         | a when a = j -> i
         | a -> a)
 
+  let swapWithNext i = swap i (i + 1)
+  let swapWithPrev i = swap i (i - 1)
+
+  let cons a ma = a :: ma
```
cmeeren, line 37:
```suggestion
  let cons x xs = x :: xs
```

src/Samples/SubModelSeq/Program.fs
```diff
@@ -13,177 +31,199 @@ module List =
...
 
+        
 [<AutoOpen>]
-module Domain =
+module IdentifiableTypes =
```
cmeeren, line 41:
```suggestion
module Identifiable =
```

src/Samples/SubModelSeq/Program.fs
```diff
@@ -13,177 +31,199 @@ module List =
...
+
+  type Identifiable<'a> =
+    { Id: Guid
+      Value: 'a }
```
cmeeren, line 45: Is there a real benefit to this type? Why not just have each counter have an ID? (I have no IDE at the moment, only GH, so I may be missing something.)

src/Samples/SubModelSeq/Program.fs
```diff
@@ -13,177 +31,199 @@ module List =
...
+    module Id =
+      let get m = m.Id
+      let set v m = { m with Id = v }
+      let map x = x |> map get set
+    module Value =
+      let get m = m.Value
+      let set v m = { m with Value = v }
+      let map f = f |> map get set
+      let mapIf p f = f |> map |> mapIf p
+      let mapIfId id f = f |> mapIf (fun m -> m.Id = id)
```
cmeeren, line 57: Going by the built-in `Result.map` and `Result.mapError` (as opposed to `Result.Ok.map` and `Result.Error.map`), I suggest removing the inner `Id` and `Value` modules and renaming `Id.get` to `getId`. The ID `set` and `map` can be removed since they are unused (this is a sample, not a library). The value function names are fine.

```suggestion
  module Identifiable =
  
    let getId m = m.Id
    
    let get m = m.Value
    
    let set v m = { m with Value = v }
    
    let map f = f |> map get set
    
    let mapIf p f = f |> map |> mapIf p
    
    let mapIfId id f = f |> mapIf (fun m -> m.Id = id)
```

src/Samples/SubModelSeq/Program.fs
```diff
@@ -13,177 +31,199 @@ module List =
...
+  
+    let create v id =
+      { Id = id
+        Value = v }
 
-  type CounterId = CounterId of Guid
+
+[<AutoOpen>]
+module CounterTypes =
```
cmeeren, line 65:
```suggestion
module Counter =
```

src/Samples/SubModelSeq/Program.fs
```diff
@@ -13,177 +31,199 @@ module List =
...
-    let setStepSize step c =
-      { c with StepSize = step }
 
-    let canReset c =
-      c.CounterValue <> 0 || c.StepSize <> 1
 
-    let reset c =
-      { create () with Id = c.Id }
+[<AutoOpen>]
+module TreeTypes =
```
cmeeren, line 95:
```suggestion
module RoseTree =
```

src/Samples/SubModelSeq/Program.fs
```diff
@@ -13,177 +31,199 @@ module List =
...
+    module Data =
+      let get m = m.Data
+      let set v m = { Data = v; Children = m.Children }
+      let rec map f n =
+        { Data = n.Data |> f
+          Children = n.Children |> (f |> map |> List.map) }
+    module Children =
+      let get m = m.Children
+      let set v m = { m with Children = v }
+      let map f = map get set f
```
cmeeren, line 113: I suggest a similar change here as with `identifiable.Id/Value`. Rename and remove any unused functions.

src/Samples/SubModelSeq/Program.fs
bender2k14: I prefer to keep it as a separate function.  Haskell considers it so important that [it is include in their prelude](http://zvon.org/other/haskell/Outputprelude/flip_f.html).  I wish F# thought as highly about this function.

If the arguments appear in this order `let flip f b a = f a b`, then the inferred types match the argument names.  I have changed to that syntax and will leave this conversation for you to resolve or continue.

src/Samples/SubModelSeq/Program.fs
bender2k14: Like your suggestion for `Func.flip`, I have simplified the syntax to `let dedup f a = f a a`.

Unlike `Func.flip`, I am not in principle against inlining this function.  However, I find the one call site no clearer after inlining this, and I think this function and its use would be easier to understand if it had a better name.  I just asked [here](https://twitter.com/bender2k14/status/1286112662003363841) if anyone is aware of such a name.  I also admit that the call site might have other ways to improve clarity.

src/Samples/SubModelSeq/Program.fs
bender2k14: Sure.  I made this change.

src/Samples/SubModelSeq/Program.fs
bender2k14: Sure.  I made this change.

src/Samples/SubModelSeq/Program.fs
bender2k14: Sure.  I made this change.

src/Samples/SubModelSeq/Program.fs
bender2k14: I prefer `a` over `x` since it matches the inferred type assigned by Visual Studio's IntelliSense and visible in the tooltips.

If the type were `Counter list`, then I would name it `counters`.  But when the (inferred) type is `'a list`, then I don't find it helpful to emphasize that the containing type is `list` by pluralizing the initial single letter with an `s` after it.  We are in the `List` module, so I think it is unnecessary to further emphasize that we are working with a `list`.

Instead, the letter `m` there is short for "monad", which I find is a more helpful reminder, especially in F# where we have to implement functions like functor's `map` for each functor because we lack higher kinded types.  There are many other monads that also have a `cons` function, and the current implementation is not inherently unique to `list` (....it is only syntactically unique).

Another advantage of the `ma` notation is that it looks like the generic type definition `M<'a>`.  If we were in some computational expression, then it is very natural to write `let! a = ma`.  This notation also scales in that it is also natural to write `mma` for an instance of type `M<M<'a>>` as well as `let! ma = mma` followed by `let! a = ma`.

I think it would be reasonable to consider a different letter than `m` since this arguments would also apply to (say) functors.  However, I don't like look/feel of `fa`, and this is also a bit confusing since `f` is most often used to identify a function.  I think using `c`, short for "container" would be reasonable.  In that case, the name would be `ca`.  I think that look good.

src/Samples/SubModelSeq/Program.fs
bender2k14: Sure

src/Samples/SubModelSeq/Program.fs
bender2k14: Sure

src/Samples/SubModelSeq/Program.fs
bender2k14: Sure

src/Samples/SubModelSeq/Program.fs
bender2k14: I picked up this naming convention from [`language-ext`](https://github.com/louthy/language-ext).  I just did a search for `ma` (including matching case and whole word).  There are 29758 matching lines.  For `mma` there are 22 matching lines.

src/Samples/SubModelSeq/Program.fs
bender2k14: A single `Guid` is what's left after extracting all the single `Counter` fields to a type that is identical to the one in the `SingleCounter` sample.

src/Samples/SubModelSeq/Program.fs
bender2k14: Sure

src/Samples/SubModelSeq/Program.fs
bender2k14: Sure.  I think all functions are used.

src/Samples/SubModelSeq/Program.fs
cmeeren: > Haskell considers it so important that [it is include in their prelude](http://zvon.org/other/haskell/Outputprelude/flip_f.html). I wish F# thought as highly about this function.

I agree, though I always use FSharpPlus which adds it, so it's a moot point for me.

In any case the reason I suggested removing it here was to simplify the sample (less to think about), but I'm fine with keeping it, especially with the rewritten syntax.

src/Samples/SubModelSeq/Program.fs
cmeeren: >  I also admit that the call site might have other ways to improve clarity.

I'll leave this unresolved until we have investigated that. The call side mixes piping and composition, which I find confusing, so it may be better to rewrite it. We'll see.

src/Samples/SubModelSeq/Program.fs
cmeeren: > I prefer `a` over `x` since it matches the inferred type assigned by Visual Studio's IntelliSense and visible in the tooltips.

I don't find `x: 'a` to be a problem, but I'm fine with keeping `a`.

I don't find the monad references of the `m` prefix helpful. In F#, it's common to append `s` to signal that a variable is a collection, e.g. `x` and `xs`. I can't find this explicitly mentioned in official MS guides, but [this location](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#formatting-lists-and-arrays) names it like that, and other well-known resources also do that (e.g F# for fun and profit, if I remember correctly).

In any case, I won't loose any sleep over this. I'll leave it to you to find a good name (or keep the current one) and resolve this conversation.

src/Samples/SubModelSeq/Program.fs
cmeeren: Fair enough. It's a balance between more reusable code (a general `Counter` has no ID) and simplicity (`Identifiable<_>` and its functions add some mental overhead).

src/Samples/SubModelSeq/Program.fs
cmeeren: Actually, I'll just resolve this. If we modify the call side (to be reviewed separately) then we'll remove it anyway if it's no longer used.

src/Samples/SubModelSeq/Program.fs
```diff
@@ -13,177 +29,192 @@ module List =
...
-      { create () with Id = c.Id }
 
+  module RoseTree =
 
-module Tree =
+    let getData m = m.Data
+    let setData v m = { Data = v; Children = m.Children }
+    let rec mapData f n =
+      { Data = n.Data |> f
+        Children = n.Children |> (f |> mapData |> List.map) }
```
cmeeren, line 99:
```suggestion
    let rec mapData f t =
      { Data = n.Data |> f
        Children = t.Children |> (f |> mapData |> List.map) }
```

src/Samples/SubModelSeq/Program.fs
```diff
@@ -13,177 +29,192 @@ module List =
...
+      { Data = a
+        Children = [] }
 
-  let rec map f n =
-    let nn = n |> f
-    { nn with Children = nn.Children |> List.map (map f) }
+    let addChild a = a |> asLeaf |> List.cons |> mapChildren
+    
+    let dataOfChildren n =
+      n.Children |> List.map getData
```
cmeeren, line 115:
```suggestion
    let dataOfChildren t =
      t.Children |> List.map getData
```

src/Samples/SubModelSeq/Program.fs
```diff
@@ -13,177 +29,192 @@ module List =
...
+    
+    let rec preorderFlatten n =
+      n :: List.collect preorderFlatten n.Children
```
cmeeren, line 118:
```suggestion
    let rec preorderFlatten t =
      t :: List.collect preorderFlatten n.Children
```

cmeeren, line 118: Why `preorder`? I don't understand the name. Is this depth-first? Could `depthFirstFlatten` be a better name?

src/Samples/SubModelSeq/Program.fs
```diff
@@ -13,177 +29,192 @@ module List =
...
+    |> Func.dedup
+    |> RoseTree.mapChildren
+
+  let update = function
+    | ToggleGlobalState -> not |> SomeGlobalState.map
+    | CounterMsg (nId, msg) -> msg |> Counter.update |> Identifiable.mapIfId nId |> RoseTree.mapData |> DummyRoot.map
+    | AddCounter pId -> () |> Guid.NewGuid |> addChildCounter |> mapIf (hasId pId) |> RoseTree.map |> DummyRoot.map
+    | Remove nId -> nId |> hasId >> not |> List.filter |> RoseTree.mapChildren |> RoseTree.map |> DummyRoot.map
+    | MoveUp nId -> nId |> swapCounters List.swapWithPrev |> RoseTree.map |> DummyRoot.map
+    | MoveDown nId -> nId |> swapCounters List.swapWithNext |> RoseTree.map |> DummyRoot.map
```
cmeeren, line 178: I find the point-free style more difficult to read here than the normal "explicit model" style. Unless you strongly disagree, could you revert to "explicit model" style? Then you can also remove a lot of the helper functions, since you can do `{ m with ...}` (I'll leave the naming of `model`/`m` to you). I think this makes the sample easier to understand. I can certainly see and appreciate the aesthetic qualities of consistent point-free style as demonstrated here, but many find it confusing.

Also, note that I'm not a fan of using pipes wherever pipes can be used. For example, I prefer `SomeGlobalState.map not` instead of `not |> SomeGlobalState.map`, and, particularly, `() |> Guid.NewGuid` should be `Guid.NewGuid ()`.

I'll continue reviewing this PR after this is fixed.

src/Samples/SubModelSeq/Program.fs
bender2k14: Yes, this is a depth-first search.  There are at least [six orderings](https://en.wikipedia.org/wiki/Depth-first_search#Vertex_orderings) when doing a DFS, and preorder is one of them.  In that sense, the name is helpful to understand the implementation.

On the other hand, I just checked, and as one might guess, the two call sites in this sample don't depend on the order (...so any DFS or any BFS would do).  For the sake of the sample then, we could simplify the name further to just `flatten`.

Want to simply call it `flatten`?

src/Samples/SubModelSeq/Program.fs
bender2k14: > Also, note that I'm not a fan of using pipes wherever pipes can be used. For example, I prefer `SomeGlobalState.map not` instead of `not |> SomeGlobalState.map`, and, particularly, `() |> Guid.NewGuid` should be `Guid.NewGuid ()`.

Sure.  I don't mind these changes.  In these small cases, my preference varies.  When writing code though, I strongly prefer to always using pipes.  I find the syntactic consistency helpful when I am in that mindset.  Then when I come back to the code just to read it (as you are), I also dislike some of the piping and remove it accordingly.

> I find the point-free style more difficult to read here than the normal "explicit model" style. Unless you strongly disagree, could you revert to "explicit model" style? Then you can also remove a lot of the helper functions, since you can do `{ m with ...}` (I'll leave the naming of `model`/`m` to you). I think this makes the sample easier to understand. I can certainly see and appreciate the aesthetic qualities of consistent point-free style as demonstrated here, but many find it confusing.

Sure.  I switched to the "explicit model" style and removed the resulting unused helper functions.  I made this change via a particularly minimal syntactic change.  Specifically, I change `msg |> f |> g |> DummyRoot.map` into `{ m with DummyRoot = (msg |> f |> g) m.DummyRoot`.  I am not trying to be difficult or obtuse with this change.  I don't mind considering further alternative ways to express this.

src/Samples/SubModelSeq/Program.fs
cmeeren: Yes, personally I would prefer that since it's irrelevant.

bender2k14: Done

src/Samples/SubModelSeq/Program.fs
cmeeren: Finally home again with access to good tooling. After more carefully reviewing this part of the PR, I actually find I prefer the way it was. I find it elegant. The only things I'd want changed (from your original suggestion) is what I mentioned above:

> For example, I prefer `SomeGlobalState.map not` instead of `not |> SomeGlobalState.map`, and, particularly, `() |> Guid.NewGuid` should be `Guid.NewGuid ()`.

src/Samples/SubModelSeq/Program.fs
bender2k14: @jacobstanley was kind enough to point out [here](https://twitter.com/jacobstanley/status/1286593647639576576) that this function is `join` aka `flatten` for the monad `'a -> 'b` (in the type parameter `'b`).  I updated the name.

src/Samples/SubModelSeq/Program.fs
```diff
@@ -13,181 +28,195 @@ module List =
...
 
-  let rec counterTreeBindings () : Binding<Model * Counter, Msg> list = [
-    "CounterIdText" |> Binding.oneWay(fun (_, { Id = CounterId id }) -> id)
-
-    "CounterValue" |> Binding.oneWay(fun (_, c) -> c.CounterValue)
-
-    "Increment" |> Binding.cmd(fun (_, c) -> Increment c.Id)
-
-    "Decrement" |> Binding.cmd(fun (_, c) -> Decrement c.Id)
+  let rec counterTreeBindings () : Binding<Model * Identifiable<Counter>, Msg> list = [
```
cmeeren, line 204: Any particular reason you have changed the spacing between the bindings in `counterTreeBindings`? I don't care that much, but I can't see any particular pattern.

src/Samples/SubModelSeq/Program.fs
bender2k14: Yay!  I am so glad that you like it.

What I find particularly elegant is that the type of the data passed from the left into each pipe operator is always* a function with signature `a -> a`.

*maybe not always, but at least most of them

I recently started writing code like this in my project at work.  I have been on journey into functional programming for about three years now.  The "first" or main idea in functional programming might be to avoid mutation.  The second idea might be to use monads to make mutation avoidance easier.  The third idea might be to use higher-order functions to make mutation avoidance easier.  I don't want to mutate, I studied monads for a long time (and I think I understand them well now), I have been using the given higher-order functions (like `List.map` and `List.filter`) for a long time starting with LINQ, but now I think I have just begun to write my own useful higher-order functions.

Back on topic...

I just pushed a commit that uses the mapping functions again.  I also made the other changes you requested.  In case I didn't do exactly what you wanted, I will leave this open for you to comment again or resolve.

src/Samples/SubModelSeq/Program.fs
```diff
@@ -13,181 +28,195 @@ module List =
...
+    | MoveUp nId -> nId |> swapCounters List.swapWithPrev |> RoseTree.map |> mapDummyRoot
+    | MoveDown nId -> nId |> swapCounters List.swapWithNext |> RoseTree.map |> mapDummyRoot
 
   /// Returns all top-level counters.
   let topLevelCounters m =
-    m.DummyRoot |> Tree.dataOfChildren
+    m.DummyRoot |> RoseTree.dataOfChildren
 
   /// Returns all immediate child counters of the specified parent counter ID.
   let childrenCountersOf pid m =
```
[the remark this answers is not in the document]
bender2k14: Done

src/Samples/SubModelSeq/Program.fs
bender2k14: While maintaining the order, I remove the blank lines between the bindings specific to a single counter, which are the ones named `CounterValue`, `Increment`, `Decrement`, `StepSize`, and `Reset`.

When I started making the changes in this branch, my goal was not merely to simplify the code in the `SubModelSeq` sample.  My goal was to extract those five bindings.  So my first first change might have been to remove those blank lines.  All the other changes were either to reduce the "noise" to help me focus on my goal or to do something useful if I didn't know what else to do to achieve my goal.

In the end, removing those lines is not important for this PR.  If you want, I can put them back.

cmeeren: No, it's fine as it is. It's moot point once #244 is merged.

cmeeren:
I'm happy with this. Feel free to merge. Thanks for your work!

bender2k14:
I locally squashed and rebased onto `master`.  I will complete the PR after the build succeeds.

src/Samples/SubModelSeq/Program.fs
bender2k14: After all that, I just pushed 51ea9f210244c17e0d54773e46cc824cc7457d03 to `master`, which changes the names to `head` and `tail`.

I still like the other naming style, but I think these names are better in this case.

Thanks for suggesting that these names could be improved :)

Issue #243, opened 2020-07-22
OnCloseRequested as a function that accepts the current model


bender2k14:
In all of the ways to create a `SubModelWin` binding, there is an optional argument of type `'msg` called `onCloseRequested`.  Here is one example.

https://github.com/elmish/Elmish.WPF/blob/783eaa42bcbd261960147cfa33ddb6f6beda9613/src/Elmish.WPF/Binding.fs#L1443

Eventually the data provided by as that argument is stored in a field called `OnCloseRequested` with type `'msg voption`

https://github.com/elmish/Elmish.WPF/blob/783eaa42bcbd261960147cfa33ddb6f6beda9613/src/Elmish.WPF/Binding.fs#L100

I would like to improve the expressiveness of the `SubModelWin` binding by changing the type of the `OnCloseRequested` field from
- from `'msg voption`
- to `'model -> 'msg voption`.

The change I want does not require a change to the public API.  However, we could expose this increased flexibility with additional overloads of `subModelWin` that have type `'model -> 'msg` for the optional argument named `onCloseRequested`.

cmeeren:
Is that needed? You have the model when you handle the specified message in the `update` function. I can't see a way to have it both places.

bender2k14:
The good reason is so I can implement `mapMsgWithModel`.  After adding this `'model` parameter to [`OnCloseRequested`](https://github.com/bender2k14/Elmish.WPF/commit/bf68353ee28c66cc2cf19797f38cd6f7ba74c528) and [`ToMsg`](https://github.com/bender2k14/Elmish.WPF/commit/3892323d834332cf50e516ceba7ec8e525fc272d) in those commits, I was able to add `mapMsgWithModel` in [this commit](https://github.com/elmish/Elmish.WPF/pull/244/commits/c9a9b16af6bbd2b04eca6c67b64c8251b66b46cc).

If the `'model` is now available in two places and we only want to have one, then my current feeling would be that the way I just added is the one to keep.

bender2k14:
Oh, I think my previous commit isn't so correct.  Let me try again.

I think the best argument for adding this `'model` argument is to avoid race conditions.  When `onCloseRequested` should be executed, the `'model` instance I am suggesting be provided to it might be a different than the one provided to `update`.

[...]

Issue #242, opened 2020-07-22
More composable SubModelSeq


bender2k14:
I want the `SubModelSeq` sample to have more composable code.  In particular, I want the counter bindings to be defined with the rest of the counter code and then for those bindings to be "mapped" to higher-level bindings at the `App` level.

I am trying to achieve this in [this branch](https://github.com/bender2k14/Elmish.WPF/tree/improve_SubModelSeq_Composability).  Although I am trying to write really good code, I am not suggesting that necessarily merge in any of those commits as is.  The primary purpose of this branch is as a proof of concept.  I first want to get it working and then consider how to internally structure the code and what the external API should be.

Below, the function `Bindings.mapMsgWithModel` doesn't exist, but I have almost created it.  To get this far, ~the two important changes are~ one important change is removing the `wrapDispatch` feature (which I originally [requested](https://github.com/elmish/Elmish.WPF/issues/114), [contributed](https://github.com/elmish/Elmish.WPF/pull/118), and later [regretted](https://github.com/elmish/Elmish.WPF/pull/212)) ~and changed the type of the binding argument `OnCloseRequested` from (essentially) `'msg` to `'model -> 'msg`.  I am going to (independently) suggest this second change in a separate issue.~  (Edited: I did add issue #243 about changing the type of `OnCloseRequested`, but I realized that it is not important because it doesn't require a change to the public API.)

Tomorrow, I plan to continue trying to create `Bindings.mapMsgWithModel`.  A remaining difficulty is the `toMsg` binding arguments.  I currently don't expect to encounter any other difficulties.

Now to be more specific, I will compare what we have now to what I want.  For what we have now, I use the code in PR #241.

# Current Code

The counter code is...
```F#
type Counter =
  { Count: int
    StepSize: int }

type CounterMsg =
  | Increment
  | Decrement
  | SetStepSize of int
  | Reset

module Counter =

  let init =
    { Count = 0
      StepSize = 1 }
  
  let canReset = (<>) init
  
  let update msg m =
    match msg with
    | Increment -> { m with Count = m.Count + m.StepSize }
    | Decrement -> { m with Count = m.Count - m.StepSize }
    | SetStepSize x -> { m with StepSize = x }
    | Reset -> init
```

...which lacks bindings while the recursive bindings are

```F#
let rec counterBindings () : Binding<Model * Identifiable<Counter>, Msg> list = [
  "CounterIdText" |> Binding.oneWay(fun (_, c) -> c.Id)

[...]

PR #244, opened 2020-07-22
DRAFT: Improve sub model seq composability

bender2k14:
This is a repost as a draft PR of issue #242 since the ability to include specific comments in the corresponding branch is likely useful.  The original comment from that issue now follows in full.  The branch linked at the start of the second paragraph is now the branch in this PR.

---

I want the `SubModelSeq` sample to have more composable code.  In particular, I want the counter bindings to be defined with the rest of the counter code and then for those bindings to be "mapped" to higher-level bindings at the `App` level.

I am trying to achieve this in [this branch](https://github.com/bender2k14/Elmish.WPF/tree/improve_SubModelSeq_Composability).  Although I am trying to write really good code, I am not suggesting that necessarily merge in any of those commits as is.  The primary purpose of this branch is as a proof of concept.  I first want to get it working and then consider how to internally structure the code and what the external API should be.

Below, the function `Bindings.mapMsgWithModel` doesn't exist, but I have almost created it.  To get this far, ~the two important changes are~ one important change is removing the `wrapDispatch` feature (which I originally [requested](https://github.com/elmish/Elmish.WPF/issues/114), [contributed](https://github.com/elmish/Elmish.WPF/pull/118), and later [regretted](https://github.com/elmish/Elmish.WPF/pull/212)) ~and changed the type of the binding argument `OnCloseRequested` from (essentially) `'msg` to `'model -> 'msg`.  I am going to (independently) suggest this second change in a separate issue.~  (Edited: I did add issue #243 about changing the type of `OnCloseRequested`, but I realized that it is not important because it doesn't require a change to the public API.)

Tomorrow, I plan to continue trying to create `Bindings.mapMsgWithModel`.  A remaining difficulty is the `toMsg` binding arguments.  I currently don't expect to encounter any other difficulties.

Now to be more specific, I will compare what we have now to what I want.  For what we have now, I use the code in PR #241.

# Current Code

The counter code is...
```F#
type Counter =
  { Count: int
    StepSize: int }

type CounterMsg =
  | Increment
  | Decrement
  | SetStepSize of int
  | Reset

module Counter =

  let init =
    { Count = 0
      StepSize = 1 }
  
  let canReset = (<>) init
  
  let update msg m =
    match msg with
    | Increment -> { m with Count = m.Count + m.StepSize }
    | Decrement -> { m with Count = m.Count - m.StepSize }
    | SetStepSize x -> { m with StepSize = x }
    | Reset -> init
```

...which lacks bindings while the recursive bindings are

```F#
let rec counterBindings () : Binding<Model * Identifiable<Counter>, Msg> list = [
  "CounterIdText" |> Binding.oneWay(fun (_, c) -> c.Id)

  "CounterValue" |> Binding.oneWay(fun (_, c) -> c.Value.Count)
  "Increment" |> Binding.cmd(fun (_, c) -> CounterMsg (c.Id, Increment))
  "Decrement" |> Binding.cmd(fun (_, c) -> CounterMsg (c.Id, Decrement))
  "StepSize" |> Binding.twoWay(
    (fun (_, c) -> float c.Value.StepSize),
    (fun v (_, c) -> CounterMsg (c.Id, SetStepSize (int v))))
  "Reset" |> Binding.cmdIf(
    (fun (_, c) -> CounterMsg (c.Id, Reset)),
    (fun (_, c) -> Counter.canReset c.Value))

  "Remove" |> Binding.cmd(fun (_, c) -> Remove c.Id)

  "AddChild" |> Binding.cmd(fun (_, c) -> AddCounter c.Id)

  "MoveUp" |> Binding.cmdIf(
    (fun (_, c) -> MoveUp c.Id),
    (fun (m, c) -> m |> childrenCountersOfParentOf c.Id |> List.tryHead <> Some c))

  "MoveDown" |> Binding.cmdIf(
    (fun (_, c) -> MoveDown c.Id),
    (fun (m, c) -> m |> childrenCountersOfParentOf c.Id |> List.tryLast <> Some c))

  "GlobalState" |> Binding.oneWay(fun (m, _) -> m.SomeGlobalState)

  "ChildCounters" |> Binding.subModelSeq(
    (fun (m, c) -> m |> childrenCountersOf c.Id),
    (fun ((m, _), childCounter) -> (m, childCounter)),
    (fun (_, c) -> c.Id),
    snd,
    counterBindings)
]
```

# Desired Code

Instead, I want the counter bindings to be defined with the rest of the Counter code like they are in the `SingleCounter` sample, which would look like...

```F#
type Counter =
  { Count: int
    StepSize: int }

type CounterMsg =
  | Increment
  | Decrement
  | SetStepSize of int
  | Reset

module Counter =

  let init =
    { Count = 0
      StepSize = 1 }
  
  let canReset = (<>) init
  
  let update msg m =
    match msg with
    | Increment -> { m with Count = m.Count + m.StepSize }
    | Decrement -> { m with Count = m.Count - m.StepSize }
    | SetStepSize x -> { m with StepSize = x }
    | Reset -> init

  let bindings () : Binding<Counter, CounterMsg> list = [
    "CounterValue" |> Binding.oneWay (fun m -> m.Count)
    "Increment" |> Binding.cmd Increment
    "Decrement" |> Binding.cmd Decrement
    "StepSize" |> Binding.twoWay(
      (fun m -> float m.StepSize),
      int >> SetStepSize)
    "Reset" |> Binding.cmdIf(Reset, canReset)
]
```

...and then the recursive bindings would be defined something like this
```F#
let rec counterBindings () : Binding<Model * Identifiable<Counter>, Msg> list =
  Counter.bindings
  |> Bindings.mapModel (fun (_, c) -> c.Value)
  |> Bindings.mapMsgWithModel (fun (_, c) msg -> CounterMsg (c.Id, msg))
  |> List.append
    [
      "CounterIdText" |> Binding.oneWay(fun (_, c) -> c.Id)
    
      "Remove" |> Binding.cmd(fun (_, c) -> Remove c.Id)
    
      "AddChild" |> Binding.cmd(fun (_, c) -> AddCounter c.Id)
    
      "MoveUp" |> Binding.cmdIf(
        (fun (_, c) -> MoveUp c.Id),
        (fun (m, c) -> m |> childrenCountersOfParentOf c.Id |> List.tryHead <> Some c))
    
      "MoveDown" |> Binding.cmdIf(
        (fun (_, c) -> MoveDown c.Id),
        (fun (m, c) -> m |> childrenCountersOfParentOf c.Id |> List.tryLast <> Some c))
    
      "GlobalState" |> Binding.oneWay(fun (m, _) -> m.SomeGlobalState)
    
      "ChildCounters" |> Binding.subModelSeq(
        (fun (m, c) -> m |> childrenCountersOf c.Id),
        (fun ((m, _), childCounter) -> (m, childCounter)),
        (fun (_, c) -> c.Id),
        snd,
        counterBindings)
    ]
```

### DIFF
--- a/src/Elmish.WPF.Tests/BindingTests.fs
+++ b/src/Elmish.WPF.Tests/BindingTests.fs
@@ -2119,7 +2119,7 @@ module cmdIf =
     let ``sets the correct binding name`` () =
       Property.check <| property {
         let! bindingName = GenX.auto<string>
-        let binding = bindingName |> Binding.cmdIf(fail, fail, id)
+        let binding = bindingName |> Binding.cmdIf(fail, fail)
         test <@ binding.Name = bindingName @>
       }
 
@@ -2476,7 +2476,7 @@ module cmdParamIf =
     let ``final autoRequery equals original uiBoundCmdParam`` () =
       Property.check <| property {
         let! uiBoundCmdParam = GenX.auto<bool>
-        let d = Binding.cmdParamIf(fail, fail, uiBoundCmdParam = uiBoundCmdParam, wrapDispatch = id) |> getCmdParamData
+        let d = Binding.cmdParamIf(fail, fail, uiBoundCmdParam = uiBoundCmdParam) |> getCmdParamData
         test <@ d.AutoRequery = uiBoundCmdParam @>
       }
 
@@ -3775,18 +3775,18 @@ module sorting =
         let! b = GenX.auto<bool>
         let! vo = GenX.auto<obj voption>
         let data =
-          [ SubModelSelectedItemData { Get = fail; Set = fail2; SubModelSeqBindingName = s; WrapDispatch = fail }
+          [ SubModelSelectedItemData { Get = fail; Set = fail2; SubModelSeqBindingName = s }
             OneWayData { Get = fail }
             OneWayLazyData { Get = fail; Map = fail; Equals = fail2 }
             OneWaySeqLazyData { Get = fail; Map = fail; Equals = fail2; GetId = fail; ItemEquals = fail2 }
-            TwoWayData { Get = fail; Set = fail2; WrapDispatch = fail }
-            TwoWayValidateData { Get = fail; Set = fail2; Validate = fail; WrapDispatch = fail }
-            CmdData { Exec = fail; CanExec = fail; WrapDispatch = fail }
-            CmdParamData { Exec = fail2; CanExec = fail2; AutoRequery = b; WrapDispatch = fail }
+            TwoWayData { Get = fail; Set = fail2 }
+            TwoWayValidateData { Get = fail; Set = fail2; Validate = fail }
+            CmdData { Exec = fail; CanExec = fail }
+            CmdParamData { Exec = fail2; CanExec = fail2; AutoRequery = b }
             SubModelData { GetModel = fail; GetBindings = fail; ToMsg = fail; Sticky = b }
             SubModelWinData { GetState = fail; GetBindings = fail; ToMsg = fail; GetWindow = fail2; IsModal = b; OnCloseRequested = vo }
             SubModelSeqData { GetModels = fail; GetId = fail; GetBindings = fail; ToMsg = fail }
-            SubModelSelectedItemData { Get = fail; Set = fail2; SubModelSeqBindingName = s; WrapDispatch = fail }
+            SubModelSelectedItemData { Get = fail; Set = fail2; SubModelSeqBindingName = s }
           ]
         let sorted = data |> List.sortWith BindingData.subModelSelectedItemLast
         match sorted with
--- a/src/Elmish.WPF.Tests/BindingTests.fs
+++ b/src/Elmish.WPF.Tests/BindingTests.fs
@@ -2984,9 +2984,10 @@ module subModel =
     [<Fact>]
     let ``final toMsg simply unboxes`` () =
       Property.check <| property {
+        let! m = GenX.auto<int>
         let! x = GenX.auto<int>
         let d = Binding.subModel(fail, fail) |> getSubModelData
-        test <@ d.ToMsg (box x) = x @>
+        test <@ d.ToMsg m (box x) = x @>
       }
 
 
@@ -3024,12 +3025,13 @@ module subModel =
     [<Fact>]
     let ``final toMsg returns value from original toMsg`` () =
       Property.check <| property {
+        let! m = GenX.auto<int>
         let! x = GenX.auto<int>
 
         let toMsg = string
         let d = Binding.subModel(fail, toMsg, fail) |> getSubModelData
 
-        test <@ d.ToMsg (box x) = toMsg x @>
+        test <@ d.ToMsg m (box x) = toMsg x @>
       }
 
 
@@ -3067,12 +3069,13 @@ module subModel =
     [<Fact>]
     let ``final toMsg returns value from original toMsg`` () =
       Property.check <| property {
+        let! m = GenX.auto<int>
         let! x = GenX.auto<int>
 
         let toMsg = string
         let d = Binding.subModel(fail, fail, toMsg, fail) |> getSubModelData
 
-        test <@ d.ToMsg (box x) = toMsg x @>
+        test <@ d.ToMsg m (box x) = toMsg x @>
       }
 
 
@@ -3123,9 +3126,10 @@ module subModelOpt =
     [<Fact>]
     let ``final toMsg simply unboxes`` () =
       Property.check <| property {
+        let! m = GenX.auto<int>
         let! x = GenX.auto<int>
         let d = Binding.subModelOpt((fail: _ -> _ voption), fail) |> getSubModelData
-        test <@ d.ToMsg (box x) = x @>
+        test <@ d.ToMsg m (box x) = x @>
       }
 
 
@@ -3182,9 +3186,10 @@ module subModelOpt =
     [<Fact>]
     let ``final toMsg simply unboxes`` () =
       Property.check <| property {
+        let! m = GenX.auto<int>
         let! x = GenX.auto<int>
         let d = Binding.subModelOpt((fail: _ -> _ option), fail) |> getSubModelData
-        test <@ d.ToMsg (box x) = x @>
+        test <@ d.ToMsg m (box x) = x @>
       }
 
 
@@ -3240,12 +3245,13 @@ module subModelOpt =
     [<Fact>]
     let ``final toMsg returns value from original toMsg`` () =
       Property.check <| property {
+        let! m = GenX.auto<int>
         let! x = GenX.auto<int>
 
         let toMsg = string
         let d = Binding.subModelOpt((fail: _ -> _ voption), toMsg, fail) |> getSubModelData
 
-        test <@ d.ToMsg (box x) = toMsg x @>
+        test <@ d.ToMsg m (box x) = toMsg x @>
       }
 
 
@@ -3302,12 +3308,13 @@ module subModelOpt =
     [<Fact>]
     let ``final toMsg returns value from original toMsg`` () =
       Property.check <| property {
+        let! m = GenX.auto<int>
         let! x = GenX.auto<int>
 
         let toMsg = string
         let d = Binding.subModelOpt((fail: _ -> _ option), toMsg, fail) |> getSubModelData
 
-        test <@ d.ToMsg (box x) = toMsg x @>
+        test <@ d.ToMsg m (box x) = toMsg x @>
       }
 
 
@@ -3364,12 +3371,13 @@ module subModelOpt =
     [<Fact>]
     let ``final toMsg returns value from original toMsg`` () =
       Property.check <| property {
+        let! m = GenX.auto<int>
         let! x = GenX.auto<int>
 
         let toMsg = string
         let d = Binding.subModelOpt((fail: _ -> _ voption), fail, toMsg, fail) |> getSubModelData
 
-        test <@ d.ToMsg (box x) = toMsg x @>
+        test <@ d.ToMsg m (box x) = toMsg x @>
       }
 
 
@@ -3427,12 +3435,13 @@ module subModelOpt =
     [<Fact>]
     let ``final toMsg returns value from original toMsg`` () =
       Property.check <| property {
+        let! m = GenX.auto<int>
         let! x = GenX.auto<int>
 
         let toMsg = string
         let d = Binding.subModelOpt((fail: _ -> _ option), fail, toMsg, fail) |> getSubModelData
 
-        test <@ d.ToMsg (box x) = toMsg x @>
+        test <@ d.ToMsg m (box x) = toMsg x @>
       }
 
 
@@ -3492,10 +3501,11 @@ module subModelSeq =
     [<Fact>]
     let ``final toMsg extracts and unboxes the second tuple element`` () =
       Property.check <| property {
+        let! m = GenX.auto<int>
         let! x = GenX.auto<int>
         let! y = GenX.auto<string>
         let d = Binding.subModelSeq(fail, fail, fail) |> getSubModelSeqData
-        test <@ d.ToMsg (box x, box y) |> unbox = y @>
+        test <@ d.ToMsg m (box x, box y) |> unbox = y @>
       }
 
 
@@ -3535,11 +3545,12 @@ module subModelSeq =
     [<Fact>]
     let ``final toMsg returns the value of original toMsg`` () =
       Property.check <| property {
+        let! m = GenX.auto<int>
         let! id = GenX.auto<int>
         let! msg = GenX.auto<string>
         let toMsg (id: int, msg: string) = msg.Length + id
         let d = Binding.subModelSeq(fail, fail, toMsg, fail) |> getSubModelSeqData
-        test <@ d.ToMsg (box id, box msg) |> unbox = toMsg (id, msg) @>
+        test <@ d.ToMsg m (box id, box msg) |> unbox = toMsg (id, msg) @>
       }
 
 
@@ -3581,11 +3592,12 @@ module subModelSeq =
       [<Fact>]
       let ``final toMsg returns the value of original toMsg`` () =
         Property.check <| property {
+          let! m = GenX.auto<int>
           let! id = GenX.auto<int>
           let! msg = GenX.auto<string>
           let toMsg (id: int, msg: string) = msg.Length + id
           let d = Binding.subModelSeq(fail, fail, fail, toMsg, fail) |> getSubModelSeqData
-          test <@ d.ToMsg (box id, box msg) |> unbox = toMsg (id, msg) @>
+          test <@ d.ToMsg m (box id, box msg) |> unbox = toMsg (id, msg) @>
         }
 
 
--- a/src/Elmish.WPF.Tests/BindingTests.fs
+++ b/src/Elmish.WPF.Tests/BindingTests.fs
@@ -3773,7 +3773,6 @@ module sorting =
       Property.check <| property {
         let! s = GenX.auto<string>
         let! b = GenX.auto<bool>
-        let! vo = GenX.auto<obj voption>
         let data =
           [ SubModelSelectedItemData { Get = fail; Set = fail2; SubModelSeqBindingName = s }
             OneWayData { Get = fail }
@@ -3784,7 +3783,7 @@ module sorting =
             CmdData { Exec = fail; CanExec = fail }
             CmdParamData { Exec = fail2; CanExec = fail2; AutoRequery = b }
             SubModelData { GetModel = fail; GetBindings = fail; ToMsg = fail; Sticky = b }
-            SubModelWinData { GetState = fail; GetBindings = fail; ToMsg = fail; GetWindow = fail2; IsModal = b; OnCloseRequested = vo }
+            SubModelWinData { GetState = fail; GetBindings = fail; ToMsg = fail; GetWindow = fail2; IsModal = b; OnCloseRequested = fail }
             SubModelSeqData { GetModels = fail; GetId = fail; GetBindings = fail; ToMsg = fail }
             SubModelSelectedItemData { Get = fail; Set = fail2; SubModelSeqBindingName = s }
           ]
--- a/src/Elmish.WPF.Tests/ViewModelTests.fs
+++ b/src/Elmish.WPF.Tests/ViewModelTests.fs
@@ -187,7 +187,7 @@ module Helpers =
     name |> createBinding (SubModelData {
       GetModel = getModel >> ValueOption.map box
       GetBindings = fun () -> bindings |> List.map boxBinding
-      ToMsg = unbox<'subMsg> >> toMsg
+      ToMsg = fun _ -> unbox<'subMsg> >> toMsg
       Sticky = sticky
     })
 
@@ -202,7 +202,7 @@ module Helpers =
       GetModels = getModels >> Seq.map box
       GetId = unbox<'subModel> >> getId >> box
       GetBindings = fun () -> bindings |> List.map boxBinding
-      ToMsg = fun (id, msg) -> toMsg (unbox<'id> id, unbox<'subMsg> msg)
+      ToMsg = fun _ (id, msg) -> toMsg (unbox<'id> id, unbox<'subMsg> msg)
     })
 
 
--- a/src/Elmish.WPF.Tests/ViewModelTests.fs
+++ b/src/Elmish.WPF.Tests/ViewModelTests.fs
@@ -141,7 +141,6 @@ module Helpers =
     name |> createBinding (TwoWayData {
       Get = get >> box
       Set = unbox<'a> >> set
-      WrapDispatch = id
     })
 
 
@@ -154,7 +153,6 @@ module Helpers =
       Get = get >> box
       Set = unbox<'a> >> set
       Validate = validate
-      WrapDispatch = id
     })
 
 
@@ -165,7 +163,6 @@ module Helpers =
     name |> createBinding (CmdData {
       Exec = exec
       CanExec = canExec
-      WrapDispatch = id
     })
 
 
@@ -178,7 +175,6 @@ module Helpers =
       Exec = unbox >> exec
       CanExec = unbox >> canExec
       AutoRequery = autoRequery
-      WrapDispatch = id
     })
 
 
@@ -219,7 +215,6 @@ module Helpers =
       Get = get >> ValueOption.map box
       Set = ValueOption.map unbox<'id> >> set
       SubModelSeqBindingName = subModelSeqBindingName
-      WrapDispatch = id
     })
 
 
--- a/src/Elmish.WPF/Binding.fs
+++ b/src/Elmish.WPF/Binding.fs
@@ -1,4 +1,4 @@
-namespace Elmish.WPF
+﻿namespace Elmish.WPF
 
 open System.Windows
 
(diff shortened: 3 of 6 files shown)


bender2k14:
Got it! :D

See the current state of the branch in this PR.  Try the `SubModelSeq` sample.  I also tested it.  I think everything still works.

In the `SubModelSeq` sample, see how the code in `CounterTypes` is identical to the code in `SingleCounter` (except the type names are different).  This is possible because...
- `BindingData<'model, 'msg>` is a contravariant functor in `'model`, and I had previously implemented the corresponding `map` function under the name `mapModel`.
- `BindingData<'model, 'msg>` is a (once again) a covariant functor in `'msg` after I removed the `wrapDispatch` feature in [this commit](https://github.com/bender2k14/Elmish.WPF/commit/806586da709f0bb9b02c6e4dae17720b03174cb1), so I was able to implement the corresponding `map` function under the name `mapMsg`.
- I implemented a function `mapMsgWithModel` that is stronger than `mapMsg` in that it accepts a mapping function that not only takes in the source `'msg` but also the current `'model`.  I am not aware of any bugs in my branch, but if one exists, I think it is mostly likely to be in the commits that change [`OnCloseRequested`](https://github.com/bender2k14/Elmish.WPF/commit/bf68353ee28c66cc2cf19797f38cd6f7ba74c528) and [`ToMsg`](https://github.com/bender2k14/Elmish.WPF/commit/3892323d834332cf50e516ceba7ec8e525fc272d) to accept the current model.
- I made a bunch of internal types like `BindingData<'model, 'msg>` public that were previously internal in [this commit.](https://github.com/bender2k14/Elmish.WPF/commit/577e05bdc9b5afb2bd45860a35ad29734757b842).  That was just the easiest way to publicly expose `mapModel` and `mapMsgWithModel`.  If we still want to restrict the user, I think we can make the "constructors" of these types private.

Since `mapMsgWithModel` is more expressive than `mapWithMsg` but has a worse name, an idea is to cheat a bit in the names by
- not exposing the current function called `mapMsg` and then
- renaming `mapMsgWithModel` to `mapMsg`.

In summary, the two API changes I made were
- removing the `wrapDispatch` feature in [this commit](https://github.com/bender2k14/Elmish.WPF/commit/806586da709f0bb9b02c6e4dae17720b03174cb1) and
- making types like `BindingData<'model, 'msg>` public that were previously internal in [this commit](https://github.com/bender2k14/Elmish.WPF/commit/577e05bdc9b5afb2bd45860a35ad29734757b842).

What do you think @cmeeren?

bender2k14:
@ScottHutchinson, just like PR #241, I think you will be interested in these further improves to the `SubModelSeq` binding.

cmeeren:
I'm confused on several fronts here, but first: Currently you're appending a lot of bindings where many overlap the source bindings (`Counter.bindings`). `CounterValue`, `Increment`, etc. Is that bug?

bender2k14:
> Currently you're appending a lot of bindings where many overlap the source bindings (`Counter.bindings`). For example `CounterValue`, `Increment`, etc. Is that bug?

Yes, that was a mistake.  I pushed a fix to remove those duplicate bindings shortly (about 90 minutes) before you posted that comment.

As a side note, I am surprised that these duplicate bindings didn't cause a problem.  I am going to investigate that furthre.

bender2k14:
> As a side note, I am surprised that these duplicate bindings didn't cause a problem. I am going to investigate that further.

I forgot what happens when there are bindings with the same name.  We log about the duplicate and continue using the first binding.

https://github.com/elmish/Elmish.WPF/blob/c20e323c466140b14acaff9d6c531f31e5bacd46/src/Elmish.WPF/ViewModel.fs#L617

The sample still works because...
```F#
Counter.bindings ()
|> Bindings.mapModel (fun (_, c) -> c.Value)
|> Bindings.mapMsgWithModel (fun (_, c) msg -> CounterMsg (c.Id, msg))
```

...and..
```F#
[ "CounterValue" |> Binding.oneWay(fun (_, c) -> c.Value.Count)
  "Increment" |> Binding.cmd(fun (_, c) -> CounterMsg (c.Id, Increment))
  "Decrement" |> Binding.cmd(fun (_, c) -> CounterMsg (c.Id, Decrement))
  "StepSize" |> Binding.twoWay(
    (fun (_, c) -> float c.Value.StepSize),
    (fun v (_, c) -> CounterMsg (c.Id, SetStepSize (int v))))
  "Reset" |> Binding.cmdIf(
    (fun (_, c) -> CounterMsg (c.Id, Reset)),
    (fun (_, c) -> Counter.canReset c.Value)) ]
```

...are identical, which is the entire goal of this PR.

Here is the some output in the logs that I didn't notice before.

```
Binding name 'CounterValue' is duplicated. Only the first occurance will be used.
Binding name 'Increment' is duplicated. Only the first occurance will be used.
Binding name 'Decrement' is duplicated. Only the first occurance will be used.
Binding name 'StepSize' is duplicated. Only the first occurance will be used.
Binding name 'Reset' is duplicated. Only the first occurance will be used.
```

Once we have better logging (c.f. #105), it won't be as easy to miss noticing these.

cmeeren:
> I pushed a fix to remove those duplicate bindings shortly (about 90 minutes) before you posted that comment.

Ok, I hadn't started looking at the code, I just read your description.

Continuing, then: I'm in favor of anything that improves Elmish.WPF, but I need to understand this better. It may be the sample, or it may be me missing something here (it's been long since I've delved deeply into Elmish.WPF code), but I can't immediately see why the new API makes things significantly better. I get that `mapModel` and `mapMsgwithModel` are general tools that make things more flexible, but they also introduce cognitive overhead. Could you give me some pointers to situations that are significantly improved by these tools? (You don't necessarily have to provide complete code; I'm just trying to understand the core concepts and benefits.)

bender2k14:
> > I pushed a fix to remove those duplicate bindings shortly (about 90 minutes) before you posted that comment.
> 
> Ok, I hadn't started looking at the code, I just read your description.

Good catch.  I just removed the "bug" from the code in my initial comment as well.

> Could you give me some pointers to situations that are significantly improved by these tools? (You don't necessarily have to provide complete code; I'm just trying to understand the core concepts and benefits.)

Ah, yes.  I will give you executable code.  This makes me realize that I forgot one thing, which is to "push" the composability into the XAML code.  I will improve the `SubModelSeq` sample further and then report back.

bender2k14:
I just pushed [this commit](https://github.com/elmish/Elmish.WPF/pull/244/commits/cd02495488ee800225a1d12318eaadb668bfb4a0), which extracts the individual counter bindings into their own XAML file.

Now in the `SubModelSeq` sample, all the individual counter code is completely independent (from the recursive tree portion of the code).  As you know @cmeeren, the entry project for these samples is the F# project.  If the entry point was the C# project, then they could demonstrate the design-time capabilities of Elmish.WPF.  Because all the individual counter code is now completely independent, it can be designed (if it didn't already exist), written, and tested (including via the design-time view) independently of the rest of the application.  Before the (binding) code was highly coupled.  This change makes the individual counter bindings completely decoupled from the other bindings in the application.

src/Samples/SubModelSeq.Views/SubModelSeq.Views.csproj
```diff
@@ -5,4 +5,10 @@
     <UseWpf>true</UseWpf>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Page Update="Counter.xaml">
+      <SubType>Designer</SubType>
+    </Page>
+  </ItemGroup>
+
```
cmeeren, line 13: I think this can safely be removed.

cmeeren:
> Because all the individual counter code is now completely independent, it can be designed (if it didn't already exist), written, and tested (including via the design-time view) independently of the rest of the application. Before the (binding) code was highly coupled. This change makes the individual counter bindings completely decoupled from the other bindings in the application.

That's awesome.

Do you have ideas about how to document the new APIs/functionality? Not just how to use it, but when to use it. Is this only useful for recursive bindings?

> If the entry point was the C# project, then they could demonstrate the design-time capabilities of Elmish.WPF.

That would require additional projects for each sample, right? (I know this has been debated before, but I'm unsure if something has changed since then.)

---

Also, I just noticed a typo, not related to this PR but I'm putting it here before I forget - feel free to fix:

> Binding name 'CounterValue' is duplicated. Only the first **occurance** will be used.

Should be "occurrence".

src/Samples/SubModelSeq.Views/SubModelSeq.Views.csproj
bender2k14: Yep.  Good catch.

For simplicity, I want to split off any changes that can be merged into `master` now.   The commit containing this change (which I will revert) is such a change.  I push such a commit to `master` and then rebase the branch in this PR.  I will mention here after that is done.

bender2k14: Done.

bender2k14:
> > Because all the individual counter code is now completely independent, it can be designed (if it didn't already exist), written, and tested (including via the design-time view) independently of the rest of the application. Before the (binding) code was highly coupled. This change makes the individual counter bindings completely decoupled from the other bindings in the application.
> 
> That's awesome.

I know!!  It makes me so happy :D

> Do you have ideas about how to document the new APIs/functionality? Not just how to use it, but when to use it.

On a micro level, `mapModel` and `mapMsg`/`mapMsgWithModel` are great at removing duplicate code.  Look at these five bindings again.
```F#
"CounterValue" |> Binding.oneWay(fun (_, c) -> c.Value.Count)
"Increment" |> Binding.cmd(fun (_, c) -> CounterMsg (c.Id, Increment))
"Decrement" |> Binding.cmd(fun (_, c) -> CounterMsg (c.Id, Decrement))
"StepSize" |> Binding.twoWay(
  (fun (_, c) -> float c.Value.StepSize),
  (fun v (_, c) -> CounterMsg (c.Id, SetStepSize (int v))))
"Reset" |> Binding.cmdIf(
  (fun (_, c) -> CounterMsg (c.Id, Reset)),
  (fun (_, c) -> Counter.canReset c.Value))
```

The function arguments getting data from the model and sending it to WPF all do "this": `fun (_, c) -> c.Value`.  The function arguments creating a message all do "this": `fun (_, c) msg -> CounterMsg (c.Id, msg)`.  The functions `mapModel` and `mapMsgWithModel` are exactly the functions that allow the removal of this duplicate code.

You might recall that I gave another (but admittedly weaker) example in https://github.com/elmish/Elmish.WPF/pull/212#issuecomment-624877706 where I would consider using `mapMsg` to remove some duplicate code.

On a macro level, the only use case I see is the present one.  The items in a `SubModelSeq` binding each have an ID.  When a message "bubbles up" from one of them, it needs to have an ID associated with it so that `update` is able to route the desired message to the correct item.  At the same time, we don't want the bindings specific to that item to have to know about this dependency on an ID.  However, that is the current situation in `master`; all the counter-specific bindings know about this ID.

That counter code doesn't contain any `SubModel` bindings.  Think about how painful it would be to have a `SubModel` binding in the counter.  The code would be so coupled.  This approach is completely unreasonable; it doesn't architecturally scale at all.

On the other hand, this one-two punch of `mapModel` followed by `mapMsgWithModel` nicely solves this problem.  Maybe there is another way to solve this problem, but I am unaware of it.

> Is this only useful for recursive bindings?

Not just recursive bindings.  This is also helpful in a flat list of submodel bindings.

This reminds me that our `SubModelSeq` sample is a bit intense.  I think it would be helpful if there were a sample that demonstrated the `SubModelSeq` binding using a flat list.

On a related note, #236 is about the idea of creating a binding specifically for trees (aka recursive bindings).  @ScottHutchinson has identified that the existing code is very inefficient.  If we created a binding for trees, then we might be able to simplify the current `SubModelSeq` binding.

> > If the entry point was the C# project, then they could demonstrate the design-time capabilities of Elmish.WPF.
> 
> That would require additional projects for each sample, right? (I know this has been debated before, but I'm unsure if something has changed since then.)

No extra projects needed.  One C# project and one F# project is sufficient.  The C# project can be the executable project.  It instantiaties `MainWindow` and then calls the existing entry point in the F# project.  See [this branch](https://github.com/bender2k14/Elmish.WPF/tree/design_time_for_samples) where I just made the this change for the `SingleCounter` sample.  I could make this branch into its own PR and make the same change for all the other samples.

I think this would be a good change.  I think it would be better to show people with executable code how design-time support works than only describe it in the README.  (However, those words in the README are excellent.  I prefer to leave them just as they are.)

> Also, I just noticed a typo, not related to this PR but I'm putting it here before I forget - feel free to fix:
> 
> > Binding name 'CounterValue' is duplicated. Only the first **occurance** will be used.
> 
> Should be "occurrence".

Great catch.  I pushed a fix to `master`.

cmeeren:
> For simplicity, I want to split off any changes that can be merged into `master` now.

Good idea. I'll take a look at this PR when you mark it as "Ready for review". Let me know in other ways before that if there's something in particular you want me to look at.

> On a micro level, ...
> On a macro level, ...

Thanks for the explanations!

> This reminds me that our `SubModelSeq` sample is a bit intense. I think it would be helpful if there were a sample that demonstrated the `SubModelSeq` binding using a flat list.

That would be great. If you want to have a go at that, knock yourself out.

> No extra projects needed. One C# project and one F# project is sufficient. The C# project can be the executable project. It instantiaties `MainWindow` and then calls the existing entry point in the F# project.

I see, so more or less what I suggested in #93 except that the F# projects are merged?

>  See [this branch](https://github.com/bender2k14/Elmish.WPF/tree/design_time_for_samples) where I just made the this change for the `SingleCounter` sample. I could make this branch into its own PR and make the same change for all the other samples.

I think that's a great idea. (Why haven't we done that already?) Feel free to go ahead with this.

bender2k14:
> > No extra projects needed. One C# project and one F# project is sufficient. The C# project can be the executable project. It instantiaties `MainWindow` and then calls the existing entry point in the F# project.
> 
> I see, so more or less what I suggested in #93 except that the F# projects are merged?

Yep.

> > See [this branch](https://github.com/bender2k14/Elmish.WPF/tree/design_time_for_samples) where I just made the this change for the `SingleCounter` sample. I could make this branch into its own PR and make the same change for all the other samples.
> 
> I think that's a great idea. (Why haven't we done that already?) Feel free to go ahead with this.

I create issue #246 for this work.

bender2k14:
> > This reminds me that our `SubModelSeq` sample is a bit intense. I think it would be helpful if there were a sample that demonstrated the `SubModelSeq` binding using a flat list.
> 
> That would be great. If you want to have a go at that, knock yourself out.

I created issue #247 for this work.

bender2k14:
> I'll take a look at this PR when you mark it as "Ready for review". Let me know in a comment before that if there's something in particular you want me to look at.

Yes, there is one thing.  Please consider the implications of removing the `wrapDispatch` feature, which I did in [this commit](https://github.com/elmish/Elmish.WPF/pull/244/commits/85bb6273be3bcb9bf272b98fd6c8194f3655e4d0).  This is a breaking change with no migration path.

I originally [requested](https://github.com/elmish/Elmish.WPF/issues/114), [contributed](https://github.com/elmish/Elmish.WPF/pull/118), and later [regretted](https://github.com/elmish/Elmish.WPF/pull/212) this `wrapDispatch` feature when I realized it prevents exactly the kind of feature I added in this PR (specifically `mapMsg`).  I think the the `wrapDispatch` feature (as implemented) and `mapMsg` feature added in this PR are mutually exclusive, and I think the `mapMsg` feature is more useful.

Even though I originally requested the `wrapDispatch` feature, I am not currently using it in my project at work.  Furthermore, I admit that I was lazy in implementing this `wrapDispatch` feature.  The feature I actually desired and still think would be good is the ability to apply adjust the stream of messages leaving a binding via the high-level concepts of throttling, debouncing, and limiting.  I don't know of any useful way to wrap `dispatch` except by using one of those concepts.  I was lazy in that I decided to only implement this less useful, more raw, but also more expressive `wrapDispatch` feature than directly implementing the concepts of throttling, debouncing, and limiting.

So although there might not be a migration path in theory, _maybe_ a migration path in practice is to add support for the higher-level concepts of throttling, debouncing, and limiting.  I say "maybe" because I will need to verify (by implementing one of them) that these features do not prevent `BindingData<'model, 'msg>` from being a covariant functor in `'msg`.

cmeeren:
If we need to make that breaking change with no migration path to identical functionality, we should announce that in an issue and request feedback, and leave that up for a while. And before that we should look at the existing discussions of the feature and any comments we have received there.

Is it really impossible to wrap dispatch when we have mapMsg? If we can implement throttling/debouncing, wouldn't it also be possible to implement a generic wrapDispatch? (Note, I haven't looked at the code, and cocariance/contravariance currently just makes my head hurt).

bender2k14:
PR #249 is now complete, which added design-time view models to our samples.  I rebased the branch in this PR onto `master`.

In https://github.com/elmish/Elmish.WPF/pull/244#issuecomment-662681917, I said (without that cross out, which I added in the quote):

> ~If the entry point was the C# project, then~ they could demonstrate the design-time capabilities of Elmish.WPF. Because all the individual counter code is now completely independent, it can be designed (if it didn't already exist), written, and tested (including via the design-time view) independently of the rest of the application. Before the (binding) code was highly coupled. This change makes the individual counter bindings completely decoupled from the other bindings in the application.

Now the C# project is the entry point, and we do have design-time view models in use.  After rebasing I added one new commit, which is to hook up the view model for `Counter.xaml`.

bender2k14:
> Is it really impossible to wrap dispatch [_with an arbitrary `wrapDispatch` function_] when we have `mapMsg`?

Yes, it really is impossible.  Here is my argument.

`BindingData<'model, 'msg>` is a DU.  One of its cases is `TwoWayData of TwoWayData<'model, 'msg, obj>`, where
```F#
type internal TwoWayData<'model, 'msg, 'a> = {
  Get: 'model -> 'a
  Set: 'a -> 'model -> 'msg
  WrapDispatch: Dispatch<'msg> -> Dispatch<'msg>
}
```

The type of `WrapDispatch` is `Dispatch<'msg> -> Dispatch<'msg>`.  This type is invariant in `'msg`.  Therefore `TwoWayData<'model, 'msg, obj>` and `BindingData<'model, 'msg>` is also invariant in `'msg`.

On the other hand, the function `mapMsg` is the function that the [Haskell wiki documentation for functor](https://wiki.haskell.org/Functor#Functor_Laws) calls `fmap`.  Although "covariant" doesn't appear on that page, it is about about a covariant functor.  Most functors and monads are covariant, so this adjective is often omitted.  For extra proof that this page really is about a covariant functor, see and contrast the Haskell documentation for [`Data.Functor`](https://hackage.haskell.org/package/base-4.8.2.0/docs/Data-Functor.html) and [`Data.Functor.Covariant`](https://hackage.haskell.org/package/contravariant-1.4/docs/Data-Functor-Contravariant.html).

Wikipedia has an article called [Covariance and contravariance (computer science)](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)).  Overall, I don't like this article very much.  I find it too verbose.  However, the initial section of the [Formal definition subsection](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)#Formal_definition) clearly and succinctly defines covariant, contravariant, bivariant, variant, and invariant.  Another downside of this Wikipedia is that it definitions use the concept of [subtyping](https://en.wikipedia.org/wiki/Subtyping), which is a form of polymorphism, which is very common in OOP.  In contrast, my experience (mostly with F#) is that prefers FP tries to avoid polymorphism (although the Java dependency ZIO claims to both be FP and polymorphic as I read [here](https://degoes.net/articles/zio-history)).

Anyway, the takeaway (as the names suggest) is that it is impossible to be both covariant and invariant at the same time.

bender2k14:
> If we can implement throttling/debouncing, wouldn't it also be possible to implement a generic `wrapDispatch`?

I don't think so.  My idea here is to name the specific types of dispatch wrapping.  Something like
```F#
type DispatchWrapper =
  | Limit of LimitParams
  | Throttle of ThrottleParams
  | Debounce of DebounceParams
```

I would change our current
- `wrapDispatch` of type `Dispatch<'msg> -> Dispatch<'msg>` to
- `dispatchWrappers` of type `DispatchWrapper seq`

...and then change...

```F#
let dispatch' = wrapDispatch dispatch
```

...to something like...
```F#
let dispatch' =
  dispatchWrappers
  |> Seq.map toDispatchWrappingFunction
  |> Seq.fold (>>) id
  <| dispatch
```

This (avoids the previous impossibility and) works because the concepts of throttling, debouncing, and limiting a data stream do not depend on the type of the data in the steam.  Specifically, `mapMsg` (and `mapMsgWithModel`) would leave `dispatchWrappers` unchanged.

cmeeren:
> However, ignore the functor laws for a moment and suppose `F<'a> = ('a -> ()) * (() -> 'a)`. Then given
> 
>     * an instance of type `('a -> ()) * (() -> 'a)` and
> 
>     * a function `f` of type `'a -> 'b`,
> 
> 
> try to create an instance of type `('b -> ()) * (() -> 'b)`. You can't because `('a -> ()) * (() -> 'a)` is invariant* in `'a`.

Thanks. I wrote the following to convince myself of this:

```f#
let mapWrapDispatch
      (map: 'a -> 'b)
      (wrap: ('a -> unit) -> ('a -> unit))
      : (('b -> unit) -> ('b -> unit)) =
  fun dispatch msg ->
    IMPOSSIBLE
```

Since this will be a breaking change, I currently stand by my previous remark:

> If we need to make that breaking change with no migration path to identical functionality, we should announce that in an issue and request feedback, and leave that up for a while. And before that we should look at the existing discussions of the feature and any comments we have received there.

bender2k14:
> we should look at the existing discussions of the feature and any comments we have received there.

Do you mean the `wrapDispatch` feature or the `mapMsg` feature?  I originally thought you meant `wrapDispatch`, but maybe I misunderstood.  If you mean the `wrapDispatch` feature, then...

The issue that created the `wrapDispatch` feature is #114.  Only you and I commented there.

> Furthermore, if we actually need to implement this ourselves, that requires some research, too, to make sure it's performant (both CPU and memory).

You mean the throttling, debouncing, and limiting features, right?  My hope is that we can delay this work.  If no one is using the `wrapDispatch` feature, then we don't even need to provide a partial migration path (right away).

cmeeren:
> Do you mean the `wrapDispatch` feature or the `mapMsg` feature?

`wrapDispatch`.



> You mean the throttling, debouncing, and limiting features, right?

Yep.

> My hope is that we can delay this work. If no one is using the `wrapDispatch` feature, then we don't even need to provide a partial migration path (right away).

Yep, that's fine. We may want to request comments in an issue to be sure (as sure as we can, anyway). Alternatively we can just release v4 and take it from there if we get any comments. People don't *have* to upgrade right away, after all. Perhaps that's the best course of action? It certainly allows us to move faster. What do you think?

bender2k14:
> Yep, that's fine. We may want to request comments in an issue to be sure (as sure as we can, anyway). Alternatively we can just release v4 and take it from there if we get any comments. People don't _have_ to upgrade right away, after all. Perhaps that's the best course of action? It certainly allows us to move faster. What do you think?

Yes, I think moving faster would be ok.  We can do a bit of both though.  I created issue #253, which can be a source of discussion for both the proposed removal of the `warpDispatch` feature as well as any other breaking change that will go into the next major version.

cmeeren:
Great! I'm happy with that approach. Thanks for creating that issue.

bender2k14:
This PR into `master` has been replaced with PR #256 into `v4`.

cmeeren:
Thanks! Did you know you can retarget PRs? The "Edit" button at the top, by the title, allows you to do that.

bender2k14:
Oh, cleary not.  Thanks for letting me know.

cmeeren:
No problem. Note that you of course have to rebase and force push the branch itself, too.

bender2k14:
Done.  Thanks for the reminder.

Why are the previous three comments marked as "off-topic"?  Is that something you did @cmeeren?  I don't see how to do that.

2020-08-09_08-50-21_234_chrome

cmeeren:
Yes, I did that. Three dots -> Hide -> Off-Topic. :) Nothing I do too much of, but it cleans the conversation up a bit for future readers (including our future selves).

PR #245, opened 2020-07-22
Also target .NET Core 3.0 in tests

### LINKED ISSUE #238: Dual target test project
Since the project that becomes a NuGet package (named `Elmish.WPF`) dual targets `netcoreapp3.0` and `net461`, our test project should also dual target these versions.

This is probably an easy fix, but I would feel more confident if issue #237 were fixed first.

bender2k14:
Fixes #238 

Test project now also targets .NET Core 3.0.

I will complete this after I see the automated build succeed.


cmeeren:
Thanks! 👍

Issue #197, opened 2020-01-27
Experience when targeting .NET Core 3.1 with VS 16.4


bender2k14:
It has been a few months now ([back in October](https://github.com/elmish/Elmish.WPF/issues/154)) since we last discussed the difference in the developer experience for WPF in Visual Studio when targeting either .NET Framework or .NET Core.  At the time, there were design-time bugs when targeting .NET Core 3.0.

Is anyone knowledgeable about the quality of the developer experience when using the latest Visual Studio (version 16.4.x) and targeting .NET Core 3.1?  How does it compare to targeting .NET Framework?

BentTranberg:
I have moved all my source to .NET Core 3.1, and continuously update VS to the latest version. I use the Elmish.WPF based design time model all the time, and can't remember anything that should indicate the issue you're reffering to still exists. Either it has been fixed, or else something must have caused it to be considerably less noticeable. I haven't had time to verify it properly, through it's been on my mind constantly to do so. As for Core vs Framework in general, the quality of the developer experience seems to be exactly equal at this time. I use many components from DevExpress, and they too work exactly the same in the Core version. When moving to .Core, I also of course switched to the new project file format, which simplifies quite a few things for me.

BentTranberg:
Spoke too soon. Hit me like a boomerang, and of course right after I say I don't see it. The issue is still there. But it does seem to be less intrusive in my design time design work. Btw runtime design in VS works well too, on Framework and Core.

bender2k14:
> When moving to .Core, I also of course switched to the new project file format, which simplifies quite a few things for me.

I think the new project format can be used when building with `dotnet` even if still targeting .NET Framework.

> Spoke too soon.

Ok.  I will test myself sometime with the expectation that this bug still exists.

Thanks for the report :)

BentTranberg:
> I think the new project format can be used when building with dotnet even if still targeting .NET Framework.

That's right, and I've used that fact to ease migration. First recreate the project file in the new project format while targeting Framework, and work through any problems with that. When done, switch target from Framework to Core, and work through any problems with that. Easy-peasy.

cmeeren:
Is there any reason this issue should stay open?

bender2k14:
This bug involves the WPF design-time in Visual Studio and .NET Core....nothing to do with Elmish.WPF, so we can close.  If anyone finds and shares the official issue tracking for this bug, that would be appreciated.

Issue #154, opened 2019-10-16
Design time model won't display data consistently when using ViewModel.designInstance


BentTranberg:
I'm trying to use ViewModel.designInstance to display a design time model, but for most of the time the designer in VS won't display the design time data, but instead just shows the user control as if no design time model is in use.

This is the repo with the solution : https://github.com/BentTranberg/ExploreElmishWpf
(I'll have to learn how to supply a link to a specific revision.)
And it's the CounterPane model.

I also use an alternative design time model, which works reliably all the time. This code snippet shows both models, and the code comments shows the corresponding XAML.

```
    // d:DataContext="{x:Static vm:CounterPane.designTimeModel}"
    let dtModel = { Count = 3; StepSize = 7 }
    let dtBindings = bindings ()
    let designTimeModel = ViewModel.designInstance dtModel dtBindings

// d:DataContext="{d:DesignInstance vm:CounterPaneDesignTimeModel, IsDesignTimeCreatable=True}"
type CounterPaneDesignTimeModel() =
    member val Count = 4 with get, set
    member val StepSize = 8 with get, set
```

I have found a silly way to trigger the display of the correct data. If I write e.g. a "z" in a very specific position, like this, in the XAML file in VS, then instantly the number 42 is displayed as Counter value.

```
d:DataContext="z{x:Static vm:CounterPane.designTimeModel}"
```
I know 42 is supposedly the answer to the Ultimate Question of Life, the Universe, and Everything, but honestly if somebody is joking with that number it's not me.

Then if I remove that "z", the correct numbers - 3 and 7 - are displayed.

But when I work on, the numbers will soon go away again at some point.

Also, if I alternately paste the two different XAML snippets into the XAML file, the corresponding data will display correctly. But only use of CounterPaneDesignTimeModel is sure to display the numbers permanently while I work on.

I've had this issue in my business project for years, and it's part of the reason why I chose to use classes as design time models.

bender2k14:
I will take a look.

In general though, my impression is that the XAML compiler and the design time are slightly buggy.  When I can't get the design time to display the design time data, one of my troubleshooting steps is to
1. commit all working changes,
2. in a git bash terminal, execute `git clean -fdx` (essentially deleting anything in the repo but not committed), and
3. build.

I recommend trying that when the design time isn't working.

cmeeren:
[...]

PR #249, opened 2020-07-27
added design-time view models for all samples

### LINKED ISSUE #246: Add design-time view models to all samples
This is a spinoff of https://github.com/elmish/Elmish.WPF/pull/244#issuecomment-663030373.

The samples are more instructive if they also show how design-time support works.  It is a relatively minor change to add this support, so the plan is to make this happen.  In particular, the C# project will become the executable project with the F# project as a dependency (which is the opposite of what it is now).

I have one question.  Each sample currently has the following naming convention for its projects:
- F#: `SampleName`
- C#: `SampleName.Views`

Since this change will swap the dependency relationship between the two projects, do we also want to change the names?  Here are some suggestions.

1. Keep them the same
- F#: `SampleName`
- C#: `SampleName.Views`

2. Swap the primary/secondary naming to something like
- F#: `SampleName.Models`
- C#: `SampleName`

3. Have an explanatory suffix on both projects with something like
- F#: `SampleName.Models`
- C#: `SampleName.Views`

bender2k14:
Resolves #246 

This branch adds design-time view models to each sample.

There were a couple that didn't work as I expected.  For example, data isn't showing up for `CounterWithClock.xaml` in the `SubModel` sample.  Not sure why.  I will look into it more.

This branch also includes some minor formatting adjustments near where I made some changes.  I think I will extract those changes, merge them into `master`, and then rebase this branch so that the diff is cleaner.

I am open to feedback now.  I marked the PR as a draft while I address the two things I mentioned above.

### DIFF
--- a/src/Samples/EventBindingsAndBehaviors.Views/EventBindingsAndBehaviors.Views.csproj
+++ b/src/Samples/EventBindingsAndBehaviors.Views/EventBindingsAndBehaviors.Views.csproj
@@ -2,11 +2,16 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.0</TargetFramework>
+    <OutputType>Exe</OutputType>
     <UseWpf>true</UseWpf>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Xaml.Behaviors.Wpf" Version="1.1.3" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\EventBindingsAndBehaviors\EventBindingsAndBehaviors.fsproj" />
   </ItemGroup>
 
 </Project>
--- a/src/Samples/EventBindingsAndBehaviors.Views/Program.cs
+++ b/src/Samples/EventBindingsAndBehaviors.Views/Program.cs
@@ -1,0 +1,11 @@
+﻿using System;
+using Elmish.WPF.Samples.EventBindingsAndBehaviors;
+using static Elmish.WPF.Samples.EventBindingsAndBehaviors.Program;
+
+namespace EventBindingsAndBehaviors.Views {
+  public static class Program {
+    [STAThread]
+    public static void Main() =>
+      main(new MainWindow());
+  }
+}
--- a/src/Samples/FileDialogs.CmdMsg/Program.fs
+++ b/src/Samples/FileDialogs.CmdMsg/Program.fs
@@ -113,17 +113,19 @@ open Core
 open Platform
 
 
+let designVm = ViewModel.designInstance (init () |> fst) (bindings ())
+
+
 let timerTick dispatch =
   let timer = new Timers.Timer(1000.)
   timer.Elapsed.Add (fun _ -> dispatch (SetTime DateTimeOffset.Now))
   timer.Start()
 
 
-[<EntryPoint; STAThread>]
-let main _ =
+let main window =
   Program.mkProgramWpfWithCmdMsg init update bindings toCmd
   |> Program.withSubscription (fun _ -> Cmd.ofSub timerTick)
   |> Program.withConsoleTrace
   |> Program.runWindowWithConfig
     { ElmConfig.Default with LogConsole = true; Measure = true }
-    (MainWindow())
+    window
--- a/src/Samples/FileDialogs/Program.fs
+++ b/src/Samples/FileDialogs/Program.fs
@@ -1,4 +1,4 @@
-module Elmish.WPF.Samples.FileDialogs.Program
+﻿module Elmish.WPF.Samples.FileDialogs.Program
 
 open System
 open System.IO
@@ -12,12 +12,13 @@ type Model =
   { CurrentTime: DateTimeOffset
     Text: string
     StatusMsg: string }
-
+    
+        
 let init () =
   { CurrentTime = DateTimeOffset.Now
     Text = ""
     StatusMsg = "" },
-  Cmd.none
+  []
 
 type Msg =
   | SetTime of DateTimeOffset
@@ -88,17 +89,19 @@ let bindings () : Binding<Model, Msg> list = [
 ]
 
 
+let designVm = ViewModel.designInstance (init () |> fst) (bindings ())
+
+
 let timerTick dispatch =
   let timer = new Timers.Timer(1000.)
   timer.Elapsed.Add (fun _ -> dispatch (SetTime DateTimeOffset.Now))
   timer.Start()
 
 
-[<EntryPoint; STAThread>]
-let main _ =
+let main window =
   Program.mkProgramWpf init update bindings
   |> Program.withSubscription (fun _ -> Cmd.ofSub timerTick)
   |> Program.withConsoleTrace
   |> Program.runWindowWithConfig
     { ElmConfig.Default with LogConsole = true; Measure = true }
-    (MainWindow())
+    window
--- a/src/Samples/NewWindow/Program.fs
+++ b/src/Samples/NewWindow/Program.fs
@@ -1,4 +1,4 @@
-module Elmish.WPF.Samples.NewWindow.Program
+﻿module Elmish.WPF.Samples.NewWindow.Program
 
 open System
 open System.Windows
@@ -112,12 +112,21 @@ module App =
   ]
 
 
-[<EntryPoint; STAThread>]
-let main _ =
-  let createWindow2 () = Window2(Owner = Application.Current.MainWindow)
-  let bindings = App.mainBindings Window1 createWindow2
+let fail _ = failwith "never called"
+let mainDesignVm = ViewModel.designInstance (App.init ()) (App.mainBindings fail fail ())
+let window1DesignVm = ViewModel.designInstance (App.init ()) (App.window1Bindings ())
+let window2DesignVm = ViewModel.designInstance App.initWindow2 (App.window2Bindings ())
+
+
+let main mainWindow (createWindow1: Func<#Window>) (createWindow2: Func<#Window>) =
+  let createWindow1 () = createWindow1.Invoke()
+  let createWindow2 () =
+    let window = createWindow2.Invoke()
+    window.Owner <- mainWindow
+    window
+  let bindings = App.mainBindings createWindow1 createWindow2
   Program.mkSimpleWpf App.init App.update bindings
   |> Program.withConsoleTrace
   |> Program.runWindowWithConfig
     { ElmConfig.Default with LogConsole = true; Measure = true }
-    (MainWindow())
+    mainWindow
--- a/src/Samples/SubModelSeq/Program.fs
+++ b/src/Samples/SubModelSeq/Program.fs
@@ -253,11 +253,11 @@ module Bindings =
     "AddCounter" |> Binding.cmd(fun m -> AddChild m.DummyRoot.Data.Id)
   ]
 
+let mainDesignVm = ViewModel.designInstance (App.init ()) (Bindings.rootBindings ())
 
-[<EntryPoint; STAThread>]
-let main _ =
+let main window =
   Program.mkSimpleWpf App.init App.update Bindings.rootBindings
   |> Program.withConsoleTrace
   |> Program.runWindowWithConfig
     { ElmConfig.Default with LogConsole = true; Measure = true }
-    (MainWindow())
+    window
--- a/README.md
+++ b/README.md
@@ -254,7 +254,7 @@ Then use the following attributes wherever you need a design-time VM:
 
 When targeting .NET Framework, “Project code” must be enabled in the XAML designer for this to work.
 
-When targeting .NET Core 3, the designer is buggy when it comes to design-time view models as used above. If you don’t get the data you expect, try to remove and re-add (e.g. undo) the `d:DataContext` line. That should make it work. You will have to do this as often as needed.
+When targeting .NET Core 3, a bug in the XAML designer causes design-time data to not be displayed through `DataContext` bindings.  See [this issue](https://developercommunity.visualstudio.com/content/problem/1133390/design-time-data-in-datacontext-binding-not-displa.html) for more details.
 
 #### Can I open new windows/dialogs?
 
--- a/src/Elmish.WPF.sln
+++ b/src/Elmish.WPF.sln
@@ -3,13 +3,13 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 16
 VisualStudioVersion = 16.0.28803.352
 MinimumVisualStudioVersion = 10.0.40219.1
-Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "SingleCounter", "Samples\SingleCounter\SingleCounter.fsproj", "{2DBB8062-2843-43A0-B73C-4777A30BD4FF}"
-EndProject
-Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "Elmish.WPF", "Elmish.WPF\Elmish.WPF.fsproj", "{8C6E8D34-7205-4C57-9722-87E30E4FC5CE}"
-EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Samples", "Samples", "{BBAFEB1E-93C0-4C7E-8E0A-026BB05C88EC}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SingleCounter.Views", "Samples\SingleCounter.Views\SingleCounter.Views.csproj", "{55F79BA4-8265-4612-8354-D04F91BF9B03}"
+EndProject
+Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "SingleCounter", "Samples\SingleCounter\SingleCounter.fsproj", "{2DBB8062-2843-43A0-B73C-4777A30BD4FF}"
+EndProject
+Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "Elmish.WPF", "Elmish.WPF\Elmish.WPF.fsproj", "{8C6E8D34-7205-4C57-9722-87E30E4FC5CE}"
 EndProject
 Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "UiBoundCmdParam", "Samples\UiBoundCmdParam\UiBoundCmdParam.fsproj", "{9581A1DE-70A6-4AC2-AF65-FD6E95F3A983}"
 EndProject
--- a/src/Samples/EventBindingsAndBehaviors.Views/MainWindow.xaml
+++ b/src/Samples/EventBindingsAndBehaviors.Views/MainWindow.xaml
@@ -1,12 +1,17 @@
 ﻿<Window x:Class="Elmish.WPF.Samples.EventBindingsAndBehaviors.MainWindow"
         xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
         xmlns:i="http://schemas.microsoft.com/xaml/behaviors"
+        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
         xmlns:local="clr-namespace:Elmish.WPF.Samples.EventBindingsAndBehaviors"
+        xmlns:vm="clr-namespace:Elmish.WPF.Samples.EventBindingsAndBehaviors;assembly=EventBindingsAndBehaviors"
         Title="EventBindings"
         Height="450"
         Width="400"
-        WindowStartupLocation="CenterScreen">
+        WindowStartupLocation="CenterScreen"
+        mc:Ignorable="d"
+        d:DataContext="{x:Static vm:Program.designVm}">
   <StackPanel Margin="15" HorizontalAlignment="Center">
     <TextBlock Text="Switch focus between these two inputs." Margin="0,0,0,10" TextAlignment="Center" />
     <StackPanel Orientation="Horizontal" HorizontalAlignment="Center">
--- a/src/Samples/EventBindingsAndBehaviors/EventBindingsAndBehaviors.fsproj
+++ b/src/Samples/EventBindingsAndBehaviors/EventBindingsAndBehaviors.fsproj
@@ -1,7 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
   
   <PropertyGroup>
-    <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp3.0</TargetFramework>
     <UseWpf>true</UseWpf>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
@@ -13,7 +12,6 @@
   
   <ItemGroup>
     <ProjectReference Include="..\..\Elmish.WPF\Elmish.WPF.fsproj" />
-    <ProjectReference Include="..\EventBindingsAndBehaviors.Views\EventBindingsAndBehaviors.Views.csproj" />
   </ItemGroup>
   
 </Project>
--- a/src/Samples/EventBindingsAndBehaviors/Program.fs
+++ b/src/Samples/EventBindingsAndBehaviors/Program.fs
@@ -1,6 +1,5 @@
 ﻿module Elmish.WPF.Samples.EventBindingsAndBehaviors.Program
 
-open System
 open Elmish
 open Elmish.WPF
 open System.Windows
@@ -67,11 +66,12 @@ let bindings () : Binding<Model, Msg> list = [
   "MousePosition" |> Binding.oneWay (fun m -> sprintf "%dx%d" m.MousePosition.X m.MousePosition.Y)
 ]
 
+let designVm = ViewModel.designInstance (init ()) (bindings ())
 
-[<EntryPoint; STAThread>]
-let main _ =
+
+let main window =
   Program.mkSimpleWpf init update bindings
   |> Program.withConsoleTrace
   |> Program.runWindowWithConfig
      { ElmConfig.Default with LogConsole = true; Measure = true }
-     (MainWindow())
+     window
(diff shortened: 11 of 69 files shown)


cmeeren:
Thanks. I'll have a look at it once the the two things you mentioned are addressed.

bender2k14:
The diff is very clean now.

I tried and again failed to get the the design-time view model working for `CounterWithClock.xaml` in the `SubModel` sample.  Issue #154 was about design-time view models working for .NET Framework but not for .NET Core.  But the current issue exists for me for both targets.

I will try again soon.

src/Samples/FileDialogs.CmdMsg/Program.fs
```diff
@@ -18,11 +18,14 @@ module Core =
     | Save of string
     | Load
 
-
-  let init () =
+
```
cmeeren, line 21: Could you remove these spaces?

src/Samples/FileDialogs.CmdMsg/Program.fs
```diff
@@ -113,17 +116,19 @@ open Core
 open Platform
 
 
+let vm = ViewModel.designInstance (initModel ()) (bindings ())
```
cmeeren, line 119: Instead of factoring out an `initModel`, how do you feel about just using `(init () |> fst)` in the call to `designInstance`? Keeps the rest of the code a bit cleaner, I think.

src/Samples/FileDialogs/Program.fs
```diff
@@ -88,17 +92,19 @@ let bindings () : Binding<Model, Msg> list = [
 ]
 
 
+let vm = ViewModel.designInstance (initModel ()) (bindings ())
```
cmeeren, line 95: See previous comment re. `initModel`.

src/Samples/EventBindingsAndBehaviors.Views/Program.cs
```diff
@@ -0,0 +1,11 @@
...
+using Elmish.WPF.Samples.EventBindingsAndBehaviors;
+using static Elmish.WPF.Samples.EventBindingsAndBehaviors.Program;
+
+namespace EventBindingsAndBehaviors.Views {
+  public static class Program {
+    [STAThread]
+    public static void Main() =>
+      main(new MainWindow());
+  }
+}
```
cmeeren, line 11: I think I prefer `using Elmish.WPF.Samples.EventBindingsAndBehaviors;` and `Program.main(...)`. Is that fine with you?

src/Samples/NewWindow/Program.fs
```diff
@@ -112,12 +112,21 @@ module App =
...
-let main _ =
-  let createWindow2 () = Window2(Owner = Application.Current.MainWindow)
-  let bindings = App.mainBindings Window1 createWindow2
+let fail _ = failwith "never called"
+let mainVm = ViewModel.designInstance (App.init ()) (App.mainBindings fail fail ())
+let window1Vm = ViewModel.designInstance (App.init ()) (App.window1Bindings ())
+let window2Vm = ViewModel.designInstance App.initWindow2 (App.window2Bindings ())
+
+
+let main (mainWindow, (createWindow1: Func<#Window>), (createWindow2: Func<#Window>)) =
```
cmeeren, line 121: AFAIK you don't have to pass a tuple here; If you pass them as normal curried F# args, then the call from C# will still be the same.

src/Samples/NewWindow/Program.fs
```diff
@@ -112,12 +112,21 @@ module App =
...
+  let createWindow1 () = createWindow1.Invoke()
+  let createWindow2 () =
+    let window = createWindow2.Invoke()
+    window.Owner <- mainWindow
+    window
```
cmeeren, line 126: An alternative is to use type parameters here:

```suggestion
let main<'Win1, 'Win2 when 'Win1: (new: unit -> 'Win1) and 'Win2: (new: unit -> 'Win2) and 'Win1 :> Window and 'Win2 :> Window> mainWindow =
  let createWindow1 () = new 'Win1()
  let createWindow2 () =
    let window = new 'Win2()
    window.Owner <- mainWindow
    window
```

(Modification to C# code also needed, of course.)

The difference is that the type parameter approach guarantees that this code controls the construction, as opposed to the C# code being able to pass a function that does other stuff, too. It's slightly more verbose, though.

Use the approach you prefer.

src/Samples/SubModelSeq/Program.fs
```diff
@@ -248,11 +248,11 @@ module Bindings =
     "AddCounter" |> Binding.cmd(fun m -> AddChild m.DummyRoot.Data.Id)
   ]
 
+let mainVm = ViewModel.designInstance (App.init ()) (Bindings.rootBindings ())
```
cmeeren, line 251: Any reason this is called `mainVm` and not `vm` like the others?

src/Samples/EventBindingsAndBehaviors.Views/EventBindingsAndBehaviors.Views.csproj
```diff
@@ -2,11 +2,16 @@
...
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Xaml.Behaviors.Wpf" Version="1.1.3" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\EventBindingsAndBehaviors\EventBindingsAndBehaviors.fsproj" />
+  </ItemGroup>
+
 </Project>
```
cmeeren, line 17: Not a comment for this file per se, but is there any reason there's no design-time VM for this sample?

src/Samples/FileDialogs.CmdMsg/Program.fs
cmeeren, line 119: How do you feel about renaming this (and others) to `designVm`, just to be entirely clear?

bender2k14: Oh, sure.  I didn't even consider that.

src/Samples/FileDialogs/Program.fs
bender2k14: Done

src/Samples/EventBindingsAndBehaviors.Views/Program.cs
bender2k14: The compiler isn't fine with it 😛 

In your code, `Program` resolves to the type on line 6 (i.e. to the `static class` named `Program` in `Program.cs` instead of the "`static class`"/`module` named `Program` in `Program.fs`).

I am have the `using static` line in my application at work, but I also name the two program files `ProgramCSharp.cs` and `ProgramFSharp.fs` to improve clarity.  If we made that change first, then we could write `ProgramFSharp.main(...)` as you suggested.

I am fine with whatever you prefer here.

cmeeren: Let's keep the current approach.

src/Samples/NewWindow/Program.fs
bender2k14: Yes, you are correct.  Thanks for the reminder.

src/Samples/NewWindow/Program.fs
bender2k14: This is interesting.  I forgot about this approach.  I do like that it avoids having to deal with `Func<>` in F#, but the additional verbosity without any need for the extra safety (since it is trivial to check that it is essentially constructors being passed, and it is also acceptable if the C# code sent a more complex function anyway) makes me prefer passing in `Func<>`s.  I will resolve without change then since you deferred to me, but if you change your mind though, then I am also fine with the type parameter solution.

src/Samples/FileDialogs.CmdMsg/Program.fs
bender2k14: Sure.  Good suggestion.

src/Samples/SubModelSeq/Program.fs
bender2k14: My intention was to use the name `vm` when there is a single XAML file because in that case there is no ambiguity.  This sample has two XAML files, which are named `MainWindow` and `Counter`, and this is the view model for `MainWindow`.  The fact that is it not easy to create a view model for the XAML file `Counter` (via bindings that are also used to define the parent bindings) is precisely what draft PR #244 is about.

src/Samples/EventBindingsAndBehaviors.Views/EventBindingsAndBehaviors.Views.csproj
bender2k14: Oh, I mistakenly thought it didn't display any data from the model.  I just added it now.

src/Samples/FileDialogs.CmdMsg/Program.fs
bender2k14: Done

src/Samples/NewWindow/Program.fs
cmeeren: I'm fine with the current approach.

cmeeren:
As far as GitHub shows me, all the conversaions are now resolved. That just leaves the problematic `CounterWithClock.xaml` view model, right?

bender2k14:
Correct.  There is more than one with that same thigh.  I don't remember all of them, but one of them is the main window in the same project.

bender2k14:
After an exceedingly long time debugging.  I figured out why `DataContext` bindings were not working at design-time.

The main issue is .NET Core.  The samples target .NET Core 3.1.  Switching to .NET Framework 4.7.2 (which I tried almost immediately when initially debugging this problem) _and also executing [`git clean -fdx`](https://tysonwilliams.coding.blog/2020-07-11_systematic_cleaning)_ fixes the problem.

I recommend that we have the samples target .NET Framework 4.7.2 and expand the existing warning in the README about the issues with design-time bindings working when targeting .NET Core.

On this topic of the other issue with bindings when targeting .NET Core, I would like to understand it better.  It was originally reported in issue #154 and then discussed again in issue #197.  @BentTranberg or @cmeeren, can you test this again when using the latest versions of .NET Core and Visual Studio?  I only half followed along before, so I am not sure how to reproduce this bug.  I tried several things, so maybe that issue is fixd, but it could also be me not testing for it correctly.

I will also either find an existing GitHub issue or create one (but don't have the time now).

cmeeren:
Sorry, I don't quite follow you. What specifically is it you want us to test?

Also, what exactly is not working for `netcoreapp3.1` but working for `net472`?

bender2k14:
> > On this topic of the other issue with bindings when targeting .NET Core, I would like to understand it better. It was originally reported in issue #154 and then discussed again in issue #197. @BentTranberg or @cmeeren, can you test this again when using the latest versions of .NET Core and Visual Studio? I only half followed along before, so I am not sure how to reproduce this bug. I tried several things, so maybe that issue is fixd, but it could also be me not testing for it correctly.
> 
> What specifically is it you want us to test?

Does the bug described in issue #154 still exist when using the latest version of .NET Core and Visual Studio?

bender2k14:
> what exactly is not working for `netcoreapp3.1` but working for `net472`?

Consider the `SubModel` sample.  Follow these steps.

## .NET Core
1. Execute `git clean -fdx` at the root of the repository.
2. Check out the branch [`design_time_VMs_for_samples`](https://github.com/bender2k14/Elmish.WPF/tree/design_time_VMs_for_samples), which is the branch for this PR and currently **targets .NET Core 3.1**.
3. Build the code.
4. Open `CounterWithClock.xaml`
5. Observe that design-time data within `DataContext` bindings are **not displayed**.

2020-08-01_09-26-10_220_devenv

## .NET Framework
1. Execute `git clean -fdx` at the root of the repository.
2. Check out the branch [`design_time_VMs_for_samples_.Net_Framework`](https://github.com/bender2k14/Elmish.WPF/tree/design_time_VMs_for_samples_.Net_Framework), which **targets .NET Framework 4.7.2**.
3. Build the code.
4. Open `CounterWithClock.xaml`
5. Observe that design-time data within `DataContext` bindings are **displayed**.

2020-08-01_09-29-46_221_devenv

BentTranberg:
> Does the bug described in issue #154 still exist when using the latest version of .NET Core and Visual Studio?

Yes, bug still there. Tested it just now with very latest public versions of VS and .NET Core SDK. And Windows 10 updated.

I believe @cmeeren suggested long time ago that this is a problem in WPF in .NET Core, and I can certainly believe that. I never got the time to make a minimal repro and post it as an issue in the dotnet/wpf repo before being laid off, so it'll have to wait one or two more months if I'm to do that when I'm back at work. I'm thinking that if this can be reproduced using C# rather than F# for the design time model (I haven't even thought about whether it can be accomplished technically), then maybe that would help attract more attention in that repo. On the other hand I like the idea of promoting F#.

cmeeren:
> > what exactly is not working for `netcoreapp3.1` but working for `net472`?
> 
> Consider the `SubModel` sample. Follow these steps.

Since the others samples/files work, what is different for `CounterWithClock.xaml`?

bender2k14:
> > > what exactly is not working for `netcoreapp3.1` but working for `net472`?
> > 
> > Consider the `SubModel` sample. Follow these steps.
> 
> Since the others samples/files work, what is different for `CounterWithClock.xaml`?

The issue is a binding to `DataContext`

bender2k14:
> The difference is a binding to `DataContext`

I will post an issue soon to the [.NET WPF repo](https://github.com/dotnet/wpf).  It will be a bit easier to see the issue in the MWE that I provide there than via the `SubModel` sample and the `CounterWithClock.xaml` file.

cmeeren:
Thanks, I see. For good measure, could you see if it works if add a `d:DataContext` binding parallell to the existing `DataContext` binding? E.g.

```xaml
<local:Counter DataContext="{Binding Counter}" d:DataContext="{Binding Counter}" HorizontalAlignment="Center" />
```

Or perhaps mess a little more around with the `d:DataContext` binding if it doesn't work. Just throwing stones at the wall here to see what sticks.

> I recommend that we have the samples target .NET Framework 4.7.2 and expand the existing warning in the README about the issues with design-time bindings working when targeting .NET Core.

I'm on board with improving the readme, but I'm a bit hesitant to make the samples target the now more or less legacy .NET Framework. I'm comfortable with having design-time VMs not work properly in all samples and chalk that up to an upstream WPF/tooling bug for now.

> I will post an issue soon to the [.NET WPF repo](https://github.com/dotnet/wpf). It will be a bit easier to see the issue in the MWE that I provide there than via the `SubModel` sample and the `CounterWithClock.xaml` file.

Thanks a lot! Regarding the C#/F# remarks by @BentTranberg:

> I'm thinking that if this can be reproduced using C# rather than F# for the design time model (I haven't even thought about whether it can be accomplished technically), then maybe that would help attract more attention in that repo. On the other hand I like the idea of promoting F#.

I'm thinking that if it's reproducible in C#, that should be very clearly stated to get it prioritized higher. As long as that's clear, I guess it won't matter particularly much whether it's demonstrated in C# or F#.

cmeeren:
As an aside, I *think* there are other ways than `d:DataContext` to get design-time data in WPF, but it's been a long time since I investigated. There may exist another method that works.

BentTranberg:
There's a risk I may have mislead you, so I just want to make it clear that what I tested was my old sample with `d:DataContext="{x:Static model:CounterPane.designTimeModel}"` and the design time support from Elmish.WPF. I see that the binding in your samples are not like that, so I'm going to test these samples now.

BentTranberg:
Allright, I can confirm that **CounterWithClock.xaml** shows design time data in branch **design_time_VMs_for_samples_.Net_Framework**, but not in branch **design_time_VMs_for_samples**. Just as shown in the screenshots above.

bender2k14:
> For good measure, could you see if it works if add a `d:DataContext` binding parallel to the existing `DataContext` binding? E.g.
> 
> ```
> <local:Counter DataContext="{Binding Counter}" d:DataContext="{Binding Counter}" HorizontalAlignment="Center" />
> ```
> 
> Or perhaps mess a little more around with the `d:DataContext` binding if it doesn't work. I'm just throwing stones at the wall here to see what sticks.

Sure.  I tired your specific suggestion, but the behavior was the same.  I didn't try anything else.

Here is [my bug report](https://developercommunity.visualstudio.com/content/problem/1133390/design-time-data-in-datacontext-binding-not-displa.html).  The dotnet/wpf repo said (via the pre-populated text in a new issue) that the correct place to submit this bug report is when the Visual Studio Developer community, so that is what I did.

bender2k14:
> I'm on board with improving the readme, but I'm a bit hesitant to make the samples target the now more or less legacy .NET Framework. I'm comfortable with having design-time VMs not work properly in all samples and chalk that up to an upstream WPF/tooling bug for now.

Ok, I am fine with that.

I would prefer to reproduce the bug from #154, submit another bug report, and then link to both bug reports from our README.

bender2k14:
> I never got the time to make a minimal repro and post it as an issue in the dotnet/wpf repo before being laid off, so it'll have to wait one or two more months if I'm to do that when I'm back at work. I'm thinking that if this can be reproduced using C# rather than F# for the design time model (I haven't even thought about whether it can be accomplished technically), then maybe that would help attract more attention in that repo. On the other hand I like the idea of promoting F#.

I am willing to create a minimal repro using only C# and post the issue with Visual Studio's Developer Community.  @BentTranberg, do you mind helping me reproduce the problem?  Is there a branch that you can share with me?  I don't mind if it is not minimal.  I can simplify it after I can reproduce it.  However, I remember that you wanted to not work on Elmish.WPF while temporarily laid off to avoid a potential conflict of interest, so please turn me down if you would prefer not to help right now.

BentTranberg:
I would absolutely post the issue directly in the dotnet/wpf repo rather than the VS Dev Community. The latter is for the millions of students, "MS customers", tinkerers and others that post flames and insist MS fix their problem. For me it's a waste of time. The repo is for people like you, who know darn well what they're doing, and typically somehow engage in fixing the problem after reporting it. I'm not that familiar with dotnet/wpf, but I've used dotnet/fsharp a lot. I even got a [thanks](https://github.com/dotnet/fsharp/issues/3510#issuecomment-325718054) from Don Syme after filing one issue report, so I sure feel welcome there as long as I engage intelligently.

My lay-off most probably ends in mid to late August, or else in September. I'll wait until then engaging in any work (also for other reasons, I'm quite busy until then), but at least I can point you to my files with the model that fails when on .NET Core 3.x

The repo is https://github.com/BentTranberg/ExploreElmishWpf

The interesting files are CounterDemo.fs, CounterDemo.xaml and CounterDemo.xaml.fs

In CounterDemo.fs there are some comments about the problem. You probably remember that we discussed the issue long ago, and I described how I used an alternative way of getting a design time model to work. The comments explain this.

I haven't followed the ongoing development you've been so busy with, so I don't know what relevance my old source has these days.

cmeeren:
> I would prefer to reproduce the bug from #154, submit another bug report, and then link to both bug reports from our README.

That would be great. Regarding #154, the readme already contains instructions based on that issue (see below), but a link won't hurt.

> When targeting .NET Core 3, the designer is buggy when it comes to design-time view models as used above. If you don’t get the data you expect, try to remove and re-add (e.g. undo) the `d:DataContext` line. That should make it work. You will have to do this as often as needed.

bender2k14:
> > Here is [my bug report](https://developercommunity.visualstudio.com/content/problem/1133390/design-time-data-in-datacontext-binding-not-displa.html). The dotnet/wpf repo said (via the pre-populated text in a new issue) that the correct place to submit this bug report is with the Visual Studio Developer community, so that is what I did.
>
> I would absolutely post the issue directly in the dotnet/wpf repo rather than the VS Dev Community.

Here is the pre-populated text I mentioned.

> Is this bug related specifically to tooling in Visual Studio (e.g. XAML Designer, Code editing, etc...)? If yes, please file the issue via the instructions [here](https://docs.microsoft.com/visualstudio/ide/how-to-report-a-problem-with-visual-studio?view=vs-2019).

Continuing to quote you @BentTranberg.

> The latter is for the millions of students, "MS customers", tinkerers and others that post flames and insist MS fix their problem. For me it's a waste of time. The repo is for people like you, who know darn well what they're doing, and typically somehow engage in fixing the problem after reporting it.

I feel the same way.  I think of Visual Studio Developer Community as the place where issues go to die.  I would much rather create an issue on GitHub.  However, I think the instruction not to post an issue with the XAML designer is clear and doing so risks angering the people best able to fix the problem.

bender2k14:
> > @BentTranberg, do you mind helping me reproduce the problem? Is there a branch that you can share with me?
>
> The repo is https://github.com/BentTranberg/ExploreElmishWpf
> 
> The interesting files are CounterDemo.fs, CounterDemo.xaml and CounterDemo.xaml.fs
> 
> In CounterDemo.fs there are some comments about the problem. You probably remember that we discussed the issue long ago, and I described how I used an alternative way of getting a design time model to work. The comments explain this.
> 
> I haven't followed the ongoing development you've been so busy with, so I don't know what relevance my old source has these days.

Great.  I have reproduced the problem now.  That was so easy.  I don't know why I had trouble following along before.  There are no issues with the ongoing development.  (Also, I was a bit surprised when you said to look at `CounterDemo.xaml.fs`, but now I see that was a simple typo.  You meant `CounterDemo.xaml.cs`.)

BentTranberg:
> You meant CounterDemo.xaml.cs

Oops! Sorry. Must be getting old. Wait, let med check the calendar ... yup, I'll be 60 in around two weeks. Yikes!

Corrected it in my post above, so as not to waste anybody's time later.

> ... as the place where issues go to die.

I feel the same way. But agree with you it's better to follow the instructions.

Just curious: Why is it working when I use that design time model that doesn't use Elmish.WPF? As shown as an alternative in CounterDemo.fs in ExploreElmishWpf. Is there some way this can be exploited to work around the problem?

bender2k14:
@BentTranberg, I changed `CounterPane` to `CounterDemo` in
```XAML
d:DataContext="{x:Static model:CounterPane.designTimeModel}"
```

...and now XAML designer and design-time data bindings are working perfectly for me.

Does this change fix your problem?

BentTranberg:
> Does this change fix your problem?

Yes. What? Darn, it seems I messed up at some point in the past, and never noticed because I did not expect things to work anyway, exactly because of this bug in .NET Core. Thanks for catching that.

So now I opened that production solution I've been working on until the virus hit, and indeed the design time models that all use `d:DataContext="{x:Static model:` and Elmish.WPF design time support are now working without the hiccups that required fiddling with the XAML to get the visuals back online. I can also see the visual designs appears much faster than they used to.

That's great. But then what about the problem uncovered in branch **design_time_VMs_for_samples**? What's that really about?

cmeeren:
> So now I opened that production solution I've been working on until the virus hit, and indeed the design time models that all use `d:DataContext="{x:Static model:` and Elmish.WPF design time support are now working *without the hiccups that required fiddling with the XAML to get the visuals back online*.

(Emphasis mine)

Are you targeting .NET Core 3? If yes, are you saying that this section of the readme doesn't apply and that things work as intended?

> When targeting .NET Core 3, the designer is buggy when it comes to design-time view models as used above. If you don’t get the data you expect, try to remove and re-add (e.g. undo) the `d:DataContext` line. That should make it work. You will have to do this as often as needed.

bender2k14:
> > Does this change fix your problem?
> 
> Yes. What? Darn, it seems I messed up at some point in the past, and never noticed because I did not expect things to work anyway, exactly because of this bug in .NET Core. Thanks for catching that.
> 
> So now I opened that production solution I've been working on until the virus hit, and indeed the design time models that all use `d:DataContext="{x:Static model:` and Elmish.WPF design time support are now working without the hiccups that required fiddling with the XAML to get the visuals back online. I can also see the visual designs appears much faster than they used to.

I went back to analyze and test [this commit](https://github.com/BentTranberg/ExploreElmishWpf/commit/ca01009ee58a8fcfd9b36240a292c348f7035395) in [your ExploreElmishWpf repo](https://github.com/BentTranberg/ExploreElmishWpf).  I think (based on timestamps) that that commit was the head of your `master` branch when you posted #154.  You didn't have the typo I just found in that commit.  I also don't have any issues with that commit.  Since you also don't have any problems with the head of your `master` branch, my guess is that the bug that you experienced and @cmeeren reproduced in #154 is now fixed.

> That's great. But then what about the problem uncovered in branch **design_time_VMs_for_samples**? What's that really about?

Yes.  This is definitely a bug with the XAML designer when targeting .NET Core.

> Are you targeting .NET Core 3? If yes, are you saying that this section of the readme doesn't apply and that things work as intended?

That is what I think @BentTranberg is saying.  I am thinking we can remove the existing statement from the README and add a more specific statement that links to the issue I created on Visual Studio Developer Community.

BentTranberg:
> Are you targeting .NET Core 3? If yes, are you saying that this section of the readme doesn't apply and that things work as intended?

Yes, this is on .NET Core 3.1. Indeed it seems to have been fixed at some point, but I have no idea when or where. I toyed around for a while trying to provoke the problem, without success.

cmeeren:
> Yes, this is on .NET Core 3.1. Indeed it seems to have been fixed at some point, but I have no idea when or where. I toyed around for a while trying to provoke the problem, without success.

Thanks. Did you try the things I mentioned as (potentially) problematic in https://github.com/elmish/Elmish.WPF/issues/154#issuecomment-542803048?

BentTranberg:
> Did you try the things I mentioned as (potentially) problematic in #154 (comment)?

Yes. But you're right - we should be cautious. This problem didn't behave in a totally consistent manner. I want to toy around a lot more than I did before I feel perfectly confident that the problem is gone, or at least doesn't appear frequently enough to be noticed or bothersome.

I'm out of time now, but can do some more testing tomorrow afternoon.

bender2k14:
In the meantime, I pushed a commit to update the README.

I will test all the samples one more time now.

bender2k14:
All the samples and design-time bindings worked (except for the [known issue](https://developercommunity.visualstudio.com/content/problem/1133390/design-time-data-in-datacontext-binding-not-displa.html)).

I pushed one more commit that makes `SimpleCounter.Views` the default startup project.

When you are satisfied @cmeeren, I will squash, rebase, force push, wait for the build to succeed, and then complete the PR.

cmeeren:
> Yes. But you're right - we should be cautious. This problem didn't behave in a totally consistent manner. I want to toy around a lot more than I did before I feel perfectly confident that the problem is gone, or at least doesn't appear frequently enough to be noticed or bothersome.

> In the meantime, I pushed a commit to update the README.

I think @BentTranberg meant we should wait with that. But it's fine, let's keep it. 😊

cmeeren:
IIRC I already reviewed this PR. I don't see any more commits here, so feel free to complete it.

BentTranberg:
> I want to toy around a lot more than I did before I feel perfectly confident that the problem is gone

I finally got the time to do that, and the problem never surfaced again. Great.

PR #255, opened 2020-08-07
Log using Microsoft.Extensions.Logging.Abstractions; implement WpfProgram

### LINKED ISSUE #105: Potential logging improvements

cmeeren:
Fixes #105

Logging implemented using `Microsoft.Extensions.Logging.Abstractions`.

I also had to create a separate Program type for Elmish.WPF. I think that has been mentioned/discussed previously, can't remember where.

Log levels can be discussed, but I think they are uncontroversial - most of what we log fits with `Trace` semantics (the most verbose level of logging), and other than that there are just a few `Warning`s.

image

Client usage is very simple and I'd say well-known in .NET. The samples are now using Serilog.

At a minimum, if users want to see logs from Elmish.WPF using Serilog, they have to install `Serilog.Extensions.Logging` (and a chosen sink, e.g. `Serilog.Sinks.Console`). The Serilog logger is initialized as normal (e.g. the static logger):

```f#
Log.Logger <- LoggerConfiguration().WriteTo.Console().CreateLogger()
```

Then users simply use `SerilogLoggerFactory` in the call to `WpfProgram.withLogger`:

```f#
WpfProgram.mkSimple (fun () -> init) update bindings
  |> WpfProgram.withLogger (new SerilogLoggerFactory())
  |> WpfProgram.runWindow window
```

Currently we log to four categories which the user can control. To see everything (as in the samples), users must do something similar to this:

```f#
LoggerConfiguration()
  .MinimumLevel.Override("Elmish.Messages", Events.LogEventLevel.Verbose)
  .MinimumLevel.Override("Elmish.State", Events.LogEventLevel.Verbose)
  .MinimumLevel.Override("Elmish.WPF.Bindings", Events.LogEventLevel.Verbose)
  .MinimumLevel.Override("Elmish.WPF.BindingPerformance", Events.LogEventLevel.Verbose)
  .WriteTo.Console()
```

(If they really just want it all, they can use `.MinimumLevel.Verbose()` instead of the four lines above, but I chose to use those four lines in the samples to clearly show how to opt into specific kinds of logs.)

I really like this way of logging and find it a good fit for Elmish.WPF. IMHO the dependency on `Microsoft.Extensions.Logging.Abstractions` is entirely unproblematic. The users have control over logging without having to write a lot of custom boilerplate.

Heck, even the "lazy" behavior of binding measure logs is preserved this way. If `Trace` (`Verbose` in Serilog) is not enabled for the `Elmish.WPF.BindingPerformance` category, then no measurements are performed.

We can of course adjust the categories if we want.


bender2k14:
Wow, you were done so quickly!  Nice work. :)

Ok.  The pragmatist in me says we should go with your suggestion.  It perfectly supports my desired use case and you are in favor of it.  So I am now in favor of it as well.

I will review and leave some comments as I find the time.

We are going to start having branches like the one here and the one in PR #244 conflict.  How would you prefer to manage the difference between v3 and v4 (if at all).  Here are some ideas that come to mind.
1. `master` remains v3 for now and we make a v4 branch
2. `master` becomes v4 and make a v3 branch
3. `master` becomes v4 and we make a v3 branch if needed

I vote for 1.  That way, we can easily use the existing devops to release a new v3 version if needed (though unlikely I think).  Also, we can merge into `master` changes that are related to our v4 work but can be merged in right away because they don't depend on it.

bender2k14:
> ```fsharp
> LoggerConfiguration()
>   .MinimumLevel.Override("Elmish.Messages", Events.LogEventLevel.Verbose)
>   .MinimumLevel.Override("Elmish.State", Events.LogEventLevel.Verbose)
>   .MinimumLevel.Override("Elmish.WPF.Bindings", Events.LogEventLevel.Verbose)
>   .MinimumLevel.Override("Elmish.WPF.BindingPerformance", Events.LogEventLevel.Verbose)
>   .WriteTo.Console()
> ```

What do you think about extracting those strings to reduce the magic to something like this?  Then auto-complete would help the user more quickly type these from scratch and get them correct.
> ```fsharp
> LoggerConfiguration()
>   .MinimumLevel.Override(LoggerName.Elmish.Messages, Events.LogEventLevel.Verbose)
>   .MinimumLevel.Override(LoggerName.Elmish.State, Events.LogEventLevel.Verbose)
>   .MinimumLevel.Override(LoggerName.Elmish.WPF.Bindings, Events.LogEventLevel.Verbose)
>   .MinimumLevel.Override(LoggerName.Elmish.WPF.BindingPerformance, Events.LogEventLevel.Verbose)
>   .WriteTo.Console()
> ```

cmeeren:
> I vote for 1.

Yes, let's go with 1 👍 

> What do you think about extracting those strings to reduce the magic

For this use case I have never seen anyone defining constants (doesn't mean no-one is doing it, though). But it can't hurt. IMHO we should primarily document the normal string version and also mention that the strings are accessible in the `LoggerName` module. Perhaps we should call it `LogCategory` or `LoggerCategory`, since ["category" is used in the MS library](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerfactory.createlogger?view=dotnet-plat-ext-3.1).

cmeeren:
Oh wait, I'm not sure it's a good idea with log category constants anyway. The categories are hierarchical, so one can use `"Elmish"` to enable anything starting with `"Elmish"`, or enable `Elmish.WPF` and then disable `"Elmish.WPF.BindingPerformance"`. This is possible despite `"Elmish"` or `"Elmish.WPF"` not being used directly as logger categories.

I think perhaps it's best if we just document the logger categories and have users use strings, as they probably are using everywhere else, too.

cmeeren:
For your review: I have not spent that much effort on log levels or logger categories. I think it can work as it is now, but I'm certainly open to other suggestions.

src/Elmish.WPF/ViewModel.fs
```diff
@@ -6,69 +6,14 @@ open System.Collections.Generic
...
 open System.Windows
+open Microsoft.Extensions.Logging
 
 open Elmish
 
 
 [<AutoOpen>]
 module internal ViewModelHelpers =
 
-  let historicalMerge
```
bender2k14, line 16: I just removed this from `master` and rebased `v4`

cmeeren: Thanks!

src/Elmish.WPF/WpfProgram.fs
```diff
@@ -0,0 +1,130 @@
...
+    Program.mkProgram init update (fun _ _ -> bindings ())
+    |> fromElmishProgram
+
+
+  /// Starts the Elmish dispatch loop, setting the bindings as the DataContext
+  /// for the specified FrameworkElement. Non-blocking. This is a low-level function;
+  /// for normal usage, see runWindow and runWindowWithConfig.
+  let startElmishLoop
+      (element: FrameworkElement)
+      (program:  WpfProgram<'model, 'msg>) =
```
bender2k14, line 51: Nit: There is an extra space in this line

bender2k14:
> Log levels can be discussed, but I think they are uncontroversial - most of what we log fits with `Trace` semantics (the most verbose level of logging), and other than that there are just a few `Warning`s.

I would prefer changing all of the `Warning`s to `Error`s.  I think a `Warning` is something that is not on the happy path, cannot be avoided, is recoverable, and is not because of a programmer mistake.  Examples include failed authentication, failed validation, and no internet connection.  I think an `Error` is something the results from a programmer mistake and could mean the program will not work correctly.  None of these will happen if the program was implemeted correctly.

```F#
log.LogWarning("subModelSelectedItem binding referenced binding '{SubModelSeqBindingName}', but no compatible binding was found with that name", d.SubModelSeqBindingName)
log.LogWarning("Binding name '{BindingName}' is duplicated. Only the first occurrence will be used.", b.Name)
log.LogWarning("[{BindingNameChain}] Attempted to close window, but did not find window reference", winPropChain)
log.LogWarning("[{BindingNameChain}] Attempted to hide window, but did not find window reference", winPropChain)
log.LogWarning("[{BindingNameChain}] Attempted to show existing hidden window, but did not find window reference", winPropChain)
log.LogWarning("[{BindingNameChain}] TryGetMember FAILED: Property {BindingName} doesn't exist", propNameChain, binder.Name)
log.LogWarning("[{BindingNameChain}] TrySetMember FAILED: Property {BindingName} doesn't exist", propNameChain, binder.Name)
log.LogWarning("[{BindingNameChain}] TrySetMember FAILED: Binding {BindingName} is read-only", propNameChain, binder.Name)
```

src/Elmish.WPF/WpfProgram.fs
```diff
@@ -0,0 +1,130 @@
...
+    |> fromElmishProgram
+
+
+  /// Creates a WpfProgram that uses commands
+  let mkProgram
+      (init: unit -> 'model * Cmd<'msg>)
+      (update: 'msg  -> 'model -> 'model * Cmd<'msg>)
+      (bindings: unit -> Binding<'model, 'msg> list) =
+    Program.mkProgram init update (fun _ _ -> bindings ())
+    |> fromElmishProgram
```
bender2k14, line 43: By immediately calling `fromElmishProgram`, there is no chance for the user to call `Program` mapping functions from the `Elmish` module.  Is it too much of a burden to have the user call this function?

cmeeren: I have attempted to add all relevant `Program` mappings to `WpfProgram`, but I was thinking we could add `mapElmishProgram` function as a fallback, so users can do anything they want. What do you think?

bender2k14: For example, I want to call `Elmish.Program.withErrorHandler`.  For this specific case, an alternative is to add another logging category and pipe this data into the logging abstraction.

Another function worth considering is `Elmish.Program.withSubscription`.  Its [current implementation is equivalent to returning the command from `init`](https://github.com/elmish/elmish/issues/183#issuecomment-526826760), but [its behavior might change in v4 of Elmish](https://github.com/elmish/elmish/issues/194#issue-516190007).

A user might also have a reason to call `withTermination`, though I don't expect to myself.

cmeeren: But the reason I haven't added it yet, is that I see only a slight benefit, and several drawbacks. The drawbacks stem from the fact that there is no current `Program` functions that are relevant for the user, and several that may be misleading:

The following are only intended for library authors:
* `map`
* `mapErrorHandler` (**Note to self:** This should be logged in this PR!)
* `setState`
* `view`
* `withSetState`
* `withSyncDispatch`

The following are already present in `WpfProgram` (some adjusted for Elmish.WPF):
* `mkSimple`
* `mkProgram`
* `withSubscription`

The following should not be used in Elmish.WPF:
* `run`
* `runWith`
* `withErrorHandler` (would break logging set up using `WpfProgram.withLogger`)
* `withConsoleTrace` (superfluous since it's controlled using `WpfProgram.withLogger`)
* `withTrace` (would break logging set up using `WpfProgram.withLogger`)

In other words, giving Elmish.WPF users the ability to map `Program` would at best be useless and at worst cause problems. In a sense, the `Program` state as used in `Elmish.WPF` is an internal implementation detail and contains invariants that must be guarded.

The only benefit in adding `mapElmishProgram` is that if new `Program` functions are added to Elmish in the future, then users can use them without Elmish.WPF needing to be updated. Given the nature of Elmish, I do not think it's very likely that there'll be frequent additions of `Program` functions that are useful for Elmish.WPF users, and if they are added, I have no problem simply adding wrappers to `WpfProgram`, like I have done with `withSubscription`.

In case you disagree with my listing above and think that some functions may be useful to Elmish.WPF users; sure, let's discuss that separately. It doesn't change the fundamental reasoning behind my position; we can just add (possibly adapted) wrappers for the useful functionality.

src/Elmish.WPF/ViewModel.fs
```diff
@@ -273,11 +274,15 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
...
-        propNameChain: string)
+        bindingPerformanceLogThresholdMs: int,
+        propNameChain: string,
+        loggerFactory: ILoggerFactory)
       as this =
   inherit DynamicObject()
 
+  let log = loggerFactory.CreateLogger("Elmish.WPF.Bindings")
+  let logBindingPerformance = loggerFactory.CreateLogger("Elmish.WPF.BindingPerformance")
+
```
bender2k14, line 285: I think the category names can be simplified to `Elmish.WPF` and `Elmish.WPF.Performance`.  The concept of "bindings" is implied since bindings for WPF are the value added by Elmish.WPF over Elmish....which is why there isn't a category called `Elmish.WPF.SomethingOtherThanBindings`.

src/Elmish.WPF/WpfProgram.fs
cmeeren: > For example, I want to call `Elmish.Program.withErrorHandler`. For this specific case, an alternative is to add another logging category and pipe this data into the logging abstraction.

I just pushed a commit doing that.

> Another function worth considering is `Elmish.Program.withSubscription`. Its [current implementation is equivalent to returning the command from `init`](https://github.com/elmish/elmish/issues/183#issuecomment-526826760), but [its behavior might change in v4 of Elmish](https://github.com/elmish/elmish/issues/194#issue-516190007).

That was already present: `WpfProgram.withSubscription`

> A user might also have a reason to call `withTermination`, though I don't expect to myself.

I can't see that function on my end.

src/Elmish.WPF/ViewModel.fs
cmeeren: With my latest commit, there's `Elmish.WPF.Update` for `update` stuff: Errors, messages and updated models. Does that change your position on this?

In any case, I like `Elmish.WPF.Performance`, I'll change that.

cmeeren:
> I would prefer changing all of the `Warning`s to `Error`s.

I agree with your reasoning and will make the change.

src/Elmish.WPF/WpfProgram.fs
```diff
@@ -0,0 +1,133 @@
...
+
+    let logMsgAndModel (msg: 'msg) (model: 'model) = 
+      program.LoggerFactory.CreateLogger("Elmish.WPF.Update").LogTrace("New message: {Message}", msg)
+      program.LoggerFactory.CreateLogger("Elmish.WPF.Update").LogTrace("Updated state:\n{Model}", model)
+
+    let logError (msg: string, ex: exn) =
+      program.LoggerFactory.CreateLogger("Elmish.WPF.Update").LogError(ex, msg)
+
+    program.ElmishProgram
+    |> Program.withTrace logMsgAndModel
```
bender2k14, line 75: `logMsgAndModel` only logs at the `Trace` level.  Can you optimize this so that `Program.withTrace` is only called when `ILogger.IsEnabled(LogLevel.Trace)` is `true`?

src/Elmish.WPF/WpfProgram.fs
```diff
@@ -0,0 +1,133 @@
...
+
+    let uiDispatch (innerDispatch: Dispatch<'msg>) : Dispatch<'msg> =
+      fun msg -> element.Dispatcher.Invoke(fun () -> innerDispatch msg)
+
+    let logMsgAndModel (msg: 'msg) (model: 'model) = 
+      program.LoggerFactory.CreateLogger("Elmish.WPF.Update").LogTrace("New message: {Message}", msg)
+      program.LoggerFactory.CreateLogger("Elmish.WPF.Update").LogTrace("Updated state:\n{Model}", model)
+
+    let logError (msg: string, ex: exn) =
+      program.LoggerFactory.CreateLogger("Elmish.WPF.Update").LogError(ex, msg)
```
bender2k14, line 72: Can you optimize this by first creating the logger and then returning a function that uses it?

src/Elmish.WPF/WpfProgram.fs
cmeeren: AFAIK there's no need for that, as most log frameworks already optimize the heck out of just that. As long as you're not doing important work and just passing arguments, that is. That's why I check `IsEnabled` for the performance logger, since if it's not enabled, we shouldn't measure.

src/Elmish.WPF/WpfProgram.fs
cmeeren: I tried, but the logs didn't show up. I think there may be a race condition. In any case, AFAIK `CreateLogger` is very cheap and intended to be called frequently. I would guess it's cached at the framework level for most log frameworks. Let me know if this is incorrect.

src/Elmish.WPF/WpfProgram.fs
bender2k14: `withTermination` was added in [this commit](https://github.com/elmish/elmish/commit/1fe85cf1f1b28e1070b3dd661a0e2f591bb7eb2e).  I don't think it is released yet.

bender2k14: I am not convinced that routing Elmish logging through Elmish.WPF use of Microsoft's logging abstraction is a good idea.  Currently the user directly interacts with Elmish logging via functions like `withErrorHandler`, `withConsoleTrace` and `withTrace`.  This is very clear.  The proposed alternative is to route these log events through Elmish.WPF and its use of Microsoft's logging abstraction under the category name `Elmish.WPF.Update`.  Even if a better name is selected for this category, this approach is less clear.

> `withConsoleTrace` (superfluous since it's controlled using `WpfProgram.withLogger`)

I wish it were that simple, but as I pointed out in https://github.com/elmish/elmish/issues/196, calling `withConsoleTrace` is not the same as calling `withTrace` and writing to the console.

src/Elmish.WPF/WpfProgram.fs
cmeeren: Hm, seems to work now. I'll make the change. Perhaps I'm mixing this with another thing I came across earlier.

cmeeren: Yep, definitely mixing things (I know what I mixed it up with now). Pushed a change where all loggers are created only once.

src/Elmish.WPF/ViewModel.fs
bender2k14: The events sent to the new category `Elmish.WPF.Update` are about an update by Elmish.  When I think about the concept of update for Elmish.WPF, I think of [`ViewModel<...>.UpdateModel(...)`](https://github.com/elmish/Elmish.WPF/blob/0ae868df7d590046899bd8b582bca174426481d2/src/Elmish.WPF/ViewModel.fs#L783).  I find the use of this category confusing.

src/Elmish.WPF/WpfProgram.fs
cmeeren: > Even if a better name is selected for this category, this approach is less clear.

I don't see why. With this PR, users of Elmish.WPF use a single standard interface to set up all logging. As an Elmish.WPF user, I would expect `WpfProgram.withLogger` to be a one-stop shop for setting up logging. Having to manually wire up "extra" loggers for the Elmish update loop adds boilerplate without utility (given that users can configure logging levels per category we define). Another point is that the fact that Elmish.WPF uses Elmish under the hood could be made an implementation detail. Not all .NET MVU frameworks use Elmish, e.g. [Fabulous](https://github.com/fsprojects/Fabulous).

src/Elmish.WPF/ViewModel.fs
cmeeren: > When I think about the concept of update for Elmish.WPF, I think of [`ViewModel<...>.UpdateModel(...)`](https://github.com/elmish/Elmish.WPF/blob/0ae868df7d590046899bd8b582bca174426481d2/src/Elmish.WPF/ViewModel.fs#L783). I find the use of this category confusing.

`Elmish.WPF.Update` is a public concept. `UpdateModel` is an internal concept that users to not know about.

Still, I welcome other suggestions for category names.

src/Elmish.WPF/WpfProgram.fs
bender2k14: > ...the fact that Elmish.WPF uses Elmish under the hood could be made an implementation detail.

Oh, that seems right.  I will think more on this.

bender2k14: Yes, I am convinced.  It is reasonable to consider that the use of Elmish by Elmish.WPF is an implement detail.  (Although we can't completely hide it because of the NuGet package dependency.)

I am fine with calling `fromElmishProgram` right away, converting and exposing `Elmish.Program` functions as `Elmish.WPF.WpfProgram` functions, and not exposing a `mapElmishProgram` function.

src/Elmish.WPF/WpfProgram.fs
```diff
@@ -0,0 +1,137 @@
+﻿namespace Elmish.WPF
+
+open System.Windows
+open Microsoft.Extensions.Logging
+open Microsoft.Extensions.Logging.Abstractions
+open Elmish
+
+
+type WpfProgram<'model, 'msg> =
```
bender2k14, line 9: Since we are not going to expose `Elmish.Program`, can we rename this to `Elmish.Wpf.Program`?

src/Elmish.WPF/WpfProgram.fs
```diff
@@ -0,0 +1,137 @@
...
+    |> fromElmishProgram
+
+
+  /// Creates a WpfProgram that uses commands
+  let mkProgram
+      (init: unit -> 'model * Cmd<'msg>)
+      (update: 'msg  -> 'model -> 'model * Cmd<'msg>)
+      (bindings: unit -> Binding<'model, 'msg> list) =
+    Program.mkProgram init update (fun _ _ -> bindings ())
+    |> fromElmishProgram
```
bender2k14, line 43: Now that Elmish.WPF has its own `Program` type, can store the `bindings` there (and get them there by passing them into `fromElmishProgram`) instead passing them through the Elmish program in this obscure way?

src/Elmish.WPF/ViewModel.fs
bender2k14: With my new perspective of Elmish as an implementation detail and our hiding of `Elmish.Program`, I like these category names.

src/Elmish.WPF/WpfProgram.fs
bender2k14: I think the warning against premature optimization is bases on these three things.
1. The optimized code is harder to understand.
2. The optimized code is slower than the unoptimized code.
3. The optimized code is faster than the optimized code, but the time spent optimizing wasn't worth the modest speed improvement.

I don't think any of these apply in this case.
1. The optimized code is still easy to read in this case.  (I will try to figure out how to suggest a commit through GitHub.)
2. The optimized code is strictly faster (unless the compiler optimizes the function calls away, in which case, the efficiency is the same).
3. The implementation time is minimal, we are are library, so it is hard for us to tell how much extreme some use case could get.

src/Elmish.WPF/WpfProgram.fs
```diff
@@ -0,0 +1,137 @@
...
+          lastModel <- Some vm
+      | Some vm ->
+          vm.UpdateModel model
+
+    let uiDispatch (innerDispatch: Dispatch<'msg>) : Dispatch<'msg> =
+      fun msg -> element.Dispatcher.Invoke(fun () -> innerDispatch msg)
+
+    let logMsgAndModel (msg: 'msg) (model: 'model) = 
+      updateLogger.LogTrace("New message: {Message}", msg)
+      updateLogger.LogTrace("Updated state:\n{Model}", model)
```
bender2k14, line 73: I think these two should be combined into one call to `LogTrace`.

src/Elmish.WPF/WpfProgram.fs
```diff
@@ -0,0 +1,137 @@
...
+
+    let logError (msg: string, ex: exn) =
+      updateLogger.LogError(ex, msg)
+
+    program.ElmishProgram
+    |> Program.withTrace logMsgAndModel
```
bender2k14, line 79:
```suggestion
    let maybeCallProgramWithTrace program =
      if updateLogger.IsEnabled(LogLevel.Trace) then
        program |> Program.withTrace logMsgAndModel
      else
        program

    program.ElmishProgram
    |> maybeCallProgramWithTrace
```

bender2k14: Then I would inline `logMsgAndModel`.

src/Elmish.WPF/WpfProgram.fs
bender2k14: I figured out how to give "commitable" change suggestions and gave the suggestion I mentioned above [here](https://github.com/elmish/Elmish.WPF/pull/255/files?file-filters%5B%5D=.fs&file-filters%5B%5D=.fsproj&hide-deleted-files=true#r467651476).

src/Elmish.WPF/WpfProgram.fs
cmeeren: I agree in principle, but unfortunately that can make the user experience worse. Consider e.g. the FileDialogs sample, where we have to `open Elmish` (in addition to `open Elmish.WPF`) because the sample uses commands. When users then type `Program.` they will see a lot of noise, because none of the `Elmish.Program` functions are relevant, only the `Elmish.WPF.Program` functions. Also it's sensitive to the order; if they `open Elmish` *after* `Open Elmish.WPF` (not that I know of any heretics who would commit such sacrilege 😜), they won't see our `mkProgram`, `withSubscription`, etc. but instead those from Elmish.

src/Elmish.WPF/WpfProgram.fs
cmeeren: Not sure I understand, but we need a "view" to be able to create a `Program`. We could defer the `Program` creation until `startElmishLoop`, but then we also need to keep track of all `Program` modifications, such as `withSubscription`. I'm not convinced that is more clear than simply "extending" the existing `Program` like we have done.

src/Elmish.WPF/WpfProgram.fs
cmeeren: Oh, I see. I read your suggestion incorrectly and thought you meant checking the level in `ViewModel.fs` before every (Trace) log statement. I'll make the change.

src/Elmish.WPF/WpfProgram.fs
cmeeren: I don't categorically disagree, but may I ask why? What is the benefit of having both the message and updated state as one log statement instead of two?

A reason to keep it this way is that both the message and state may be multiline strings, so it may be easier to find what you want in logs if it's two separate entries. Also, if users want to log only messages or only state even if we use the same category, they may add logging filters to exclude one or the other. (Though if we think users will need that, we should use different log categories.)

But I could certainly be convinced the other way.

src/Elmish.WPF/WpfProgram.fs
bender2k14: Yes, great point.  How much to do you want to hide Elmish as an implementation detail?  Elmish.WPF could also expose its own `Cmd` type and map it to the `Cmd` type in Elmish.

cmeeren: I don't see the point of doing that. I suggest letting users use the Elmish interfaces for that type.

src/Elmish.WPF/WpfProgram.fs
bender2k14: > Oh, I see. I read your suggestion incorrectly and thought you meant checking the level in ViewModel.fs before every (Trace) log statement. I'll make the change.

I would also prefer to optimize those `Trace` log statements in `ViewModel.fs`.  Not by wrapping each on in an `if` with a predicate that gets evaluated each time (say) WPF calls `TryGetMember` but just like the suggestion above, in such a way that the predicate gets evaluated (conceptually) once for each execution of the program.

However, this change isn't very important and it shouldn't be made in this PR anyway.

src/Elmish.WPF/WpfProgram.fs
bender2k14: Ok then.  I have fine this the name `WpfProgram`.

src/Elmish.WPF/WpfProgram.fs
cmeeren: See e.g. the [Serilog docs](https://github.com/serilog/serilog/wiki/Writing-Log-Events#level-detection) which say:

> In most cases, applications should write events without checking the active logging level. Level checking is extremely cheap and the overhead of calling disabled logger methods very low.
> 
> In *rare, performance-sensitive cases*, the recommended pattern for level detection is to

I'm not at all convinced the logging we're doing in ViewModel is such a *rare, performance-sensitive case*. But if you can point to some popular logging framework where normal logging statements are *not* extremely cheap, I can certainly bend the other way.

src/Elmish.WPF/WpfProgram.fs
bender2k14: I specifically mean the change in [this commit](https://github.com/bender2k14/Elmish.WPF/commit/febe5bcd44fe277ca8ebcc6b49a206608e251962) that I just made.  This code makes it easier to understand where the bindings come from.  The current code adds this unnecessary layer of direction via the `View` function.

src/Elmish.WPF/WpfProgram.fs
bender2k14: > In rare, performance-sensitive cases, the recommended pattern for level detection is to store the results of level detection in a field, for example:
```C#
// in some class
readonly bool _isDebug = Log.IsEnabled(LogEventLevel.Debug);

// in some method in that class
if (_isDebug) Log.Debug("Someone is stuck debugging...");
```

I never want to do that.  I never want my code guard each potential call to (say) `Log.Debug` with such an `if` statement.

Instead, I want to optimize like I suggested in the initial comment in this thread and you implemented as
```F#
|> if updateLogger.IsEnabled LogLevel.Trace then Program.withTrace logMsgAndModel else id
```

This `if` statement is only executed once each time the entire application is started.  That is what I am referring to when I said
> 2. The optimized code is strictly faster (unless the compiler optimizes the function calls away, in which case, the efficiency is the same).

We pay the cost of evaluating that `if` statement once.  In exchange, we do highly optimized behavior in each case.  In particular, we never even try to log when logging at that log level is suppressed.

The performance logging code also uses the same highly efficient optimization (and you have maintained it in this branch).
```F#
let measure name callName f =
  if not <| logPerformance.IsEnabled(LogLevel.Trace) then f
  else
    fun x -> // measure, do work, and log
```

That `if` statement is executed once per `ViewModel<...>` instance (during initialization).  It could be improved to also execute just once per execution of the entire application, but this change is also a low priority and not appropriate for this PR.

bender2k14: I wish there was a name for this approach to logging.  One perspective is that it is runtime optimization related to logging.  The other is that it is a very good separation of concerns.  For example, it is very pleasing to read the `ViewModel<...>` code and _not see_ (possibly suppressed) performance logging calls in all the binding cases when (say) `get` and `set` are called.

Normally my experience is that code that is easy to read and well architectured has good performance but not the best.  That is ok though because the best performance isn't needed.

In the case of logging, my claim is that this well-architectured code also has the best performance.

src/Elmish.WPF/WpfProgram.fs
bender2k14: As I said in https://github.com/elmish/Elmish.WPF/issues/105#issuecomment-524628014
> 1. [Emitting multiple log events for the same application event] makes the developer read the log history instead of providing all the relevant information in a single log event. ...
> 2. Those two log event could be written to their sink arbitrarily far apart if the application is concurrently doing other things that produce log events.

In fact, I just learned in the last week about [the features of Log and Diagnostic context in Serilog](https://benfoster.io/blog/serilog-best-practices/#log-and-diagnostic-context).  The basic idea is emitting a single log event for some application event instead of emitting more than one log event and correlating them with a correlation ID.

src/Elmish.WPF/WpfProgram.fs
cmeeren: Oh, I see. It's about the `let bindings = Program.view ...` line where we extract the bindings again. Sure, I'll change it. Thanks!

src/Elmish.WPF/WpfProgram.fs
bender2k14: > A reason to keep it this way is that both the message and state may be multiline strings, so it may be easier to find what you want in logs if it's two separate entries. Also, if users want to log only messages or only state even if we use the same category, they may add logging filters to exclude one or the other. (Though if we think users will need that, we should use different log categories.)

If you want to make it easier to ready, then add more new lines.  Maybe something like
```F#
updateLogger.LogTrace("Created new model from a message.\nMessage: {Message}\nModel: {Model}", msg, model)
```

src/Elmish.WPF/WpfProgram.fs
cmeeren: Maybe I'm tired, but I'm having a hard time following exactly what you mean. But let's not spend more time on that right now, since it's not relevant for this PR.

bender2k14: Sure.  (It is probably me not clearly explaining myself though.)

src/Elmish.WPF/WpfProgram.fs
cmeeren: You've got me convinced. Just pushed a change that merges these.

src/Elmish.WPF/WpfProgram.fs
bender2k14: How about removing this space?

src/Samples/EventBindingsAndBehaviors/Program.fs
```diff
@@ -68,10 +69,16 @@ let bindings () : Binding<Model, Msg> list = [
...
+  Log.Logger <- 
+    LoggerConfiguration()
+      .MinimumLevel.Override("Elmish.WPF.Update", Events.LogEventLevel.Verbose)
+      .MinimumLevel.Override("Elmish.WPF.Bindings", Events.LogEventLevel.Verbose)
+      .MinimumLevel.Override("Elmish.WPF.Performance", Events.LogEventLevel.Verbose)
+      .WriteTo.Console()
+      .CreateLogger()
+
+  WpfProgram.mkSimple init update bindings
+  |> WpfProgram.withLogger (new SerilogLoggerFactory())
```
bender2k14, line 83: Instead of implicitly suggesting to users that they use Serilog's static logger, can we change all the samples to use a local logger.

```suggestion
  let logger =
    LoggerConfiguration()
      .MinimumLevel.Override("Elmish.WPF.Update", Events.LogEventLevel.Verbose)
      .MinimumLevel.Override("Elmish.WPF.Bindings", Events.LogEventLevel.Verbose)
      .MinimumLevel.Override("Elmish.WPF.Performance", Events.LogEventLevel.Verbose)
      .WriteTo.Console()
      .CreateLogger()

  WpfProgram.mkSimple init update bindings
  |> WpfProgram.withLogger (new SerilogLoggerFactory(logger))
```

README.md
```diff
@@ -296,3 +296,29 @@ let save text =
...
+WpfProgram.mkSimple init update bindings
+|> WpfProgram.withLogger yourLoggerFactory
+|> WpfProgram.runWindow window
+```
+
+For example, in Serilog, you need to install Serilog.Extensions.Logging and instantiate `SerilogLoggerFactory`. The samples demonstrate this.
+
+Elmish.WPF logs to these categories:
+
+* `Elmish.WPF.Update`: Logs exceptions (Error level) and messages/models (Trace/Verbose level) during `update`.
```
bender2k14, line 314: Since `Microsoft.Extensions.Logging` calls it `Trace`, can we stick to that naming?

```suggestion
* `Elmish.WPF.Update`: Logs exceptions (Error level) and messages/models (Trace level) during `update`.
```

README.md
```diff
@@ -296,3 +296,29 @@ let save text =
...
+* `Elmish.WPF.Bindings`: Logs events related to bindings. Some logging is done at the Error level (e.g. developer errors such as duplicated binding names, using non-existent bindings in XAML, etc.), but otherwise it’s generally just Trace/Verbose for when you really want to see everything that’s happening (triggering `PropertyChanged`, WPF getting/setting bindings, etc.)
```
bender2k14, line 315: Same here.

```suggestion
* `Elmish.WPF.Bindings`: Logs events related to bindings. Some logging is done at the Error level (e.g. developer errors such as duplicated binding names, using non-existent bindings in XAML, etc.), but otherwise it’s generally just Trace for when you really want to see everything that’s happening (triggering `PropertyChanged`, WPF getting/setting bindings, etc.)
```

README.md
```diff
@@ -296,3 +296,29 @@ let save text =
...
+* `Elmish.WPF.Performance`: Logs the performance of the functions you pass when creating bindings (`get`, `set`, `map`, `equals`, etc.) at the Trace/Verbose level. Use `WpfProgram.withPerformanceLogThreshold` to set the minimum duration to log.
```
bender2k14, line 316: Same here.

```suggestion
* `Elmish.WPF.Performance`: Logs the performance of the functions you pass when creating bindings (`get`, `set`, `map`, `equals`, etc.) at the Trace level. Use `WpfProgram.withPerformanceLogThreshold` to set the minimum duration to log.
```

bender2k14:
I just closely reviewed all the file changes.  I left some minor change suggestions.  Otherwise, everything else looks good.

src/Samples/EventBindingsAndBehaviors/Program.fs
cmeeren: Is this based on some documented best practice for Serilog? Personally I always use the static logger. I find it a piece of global mutable state that simplifies rather than complicates. I have never found a benefit to explicitly passing loggers around as arguments; it's just noise IMHO. (If I was doing OO I'd pass them using DI, but I'm not, so I don't.)

README.md
cmeeren: I think it's fine since our samples use Serilog and it's called `Verbose` there. Each sample has no references to `Trace` and three references to `Verbose`. But if you feel strongly about it, we can change it.

src/Samples/EventBindingsAndBehaviors/Program.fs
bender2k14: It is based on the principle of avoiding (static and) mutability.  I also always set the static logger, and I always log to it, but I don't always directly log to it.

For example, when I upgrade to Elmish.WPF 4.0 in my application at work, my logging will look more like this than the way the samples do now.
```F#
Log.Logger <-
  LoggerConfiguration()
    .WriteTo.Console()
    .CreateLogger()

let elmishWpfLogger =
  LoggerConfiguration()
    .MinimumLevel.Override("Elmish.WPF.Update", Events.LogEventLevel.Verbose)
    .MinimumLevel.Override("Elmish.WPF.Bindings", Events.LogEventLevel.Verbose)
    .MinimumLevel.Override("Elmish.WPF.Performance", Events.LogEventLevel.Verbose)
    .WriteTo.Logger(Log.Logger)
    .CreateLogger()

WpfProgram.mkSimple (fun () -> init) update bindings
|> WpfProgram.withLogger (new SerilogLoggerFactory(elmishWpfLogger))
|> WpfProgram.runWindow window
```

This code does a great job helping to demystify the magic category strings.

In contrast, consider the task of starting with the logging code in the sample and refactoring to what I have written above.  And suppose that code exists along side code that also passes Serilog loggers to Entity Framework Core and ASP.NET Core.  It is basically impossible to statically verify if the transformation is behavior preserving.  The names suggest they are, but just think about how many times you were completely confident that some change was behavior preserving but you ended up being wrong.  I know this has happened to me countless times.  I am getting better, but I am still not perfect.

Since I can't statically analyze the code, I next try to "statically analyze" the person that wrote it.  Did they not know that there was a simpler way to achieve the same behavior (and with the same amount of code)?  Did they know this but didn't do it out of laziness?  Or did they do it on purpose because there is some non-obvious reason that makes the approach better.  This is where a comment should be.  It is a comment that communicates "why".  Self-documenting code is a good thing, but it only communicates "how", not "why".

If you are not convinced though, we can keep code how it is.

README.md
bender2k14: I don't feel strongly, so whatever you want is fine.

src/Samples/EventBindingsAndBehaviors/Program.fs
cmeeren: I'm not sure I follow all parts of your reasoning, but I don't really have a strong opinion here, so I made the change. :)

README.md
cmeeren: Let's keep it like this, then.

cmeeren:
Everything ready, then?

By the way, if you ever want to squash all commits in a PR before merging, you can do it using the dropdown on the Merge button.

bender2k14:
Yep, ready.

I am such a perfectionist.  Sometimes I want more than just one commit, and I am curious to see what the commit message is when you do it now.

cmeeren:
You can choose the message yourself. By deafult it seems to set the first commit message line to the PR title, and it also includes every commit message on the subsequent lines (I removed those).

PR #259, opened 2020-08-13
Make explicit the currently implicit type constraints

### LINKED ISSUE #139: Documentation improvements for 'id
This issue spun out of #131.  We should improve the documentation of `'id` to say how equality is computed.

Currently, equality is computed explicitly using F#'s [`(=)`](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/symbol-and-operator-reference/arithmetic-operators#summary-of-binary-comparison-operators) and implicitly using .NET's [`EqualityComparer<>.Default`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.equalitycomparer-1.default?view=netframework-4.8).

I expect only `EqualityComparer<>.Default` will still be in use by the time #137 is closed, so I recommend we wait until that happens to make any change here.

### LINKED ISSUE #151: Implement the strategy design pattern for VmBinding

bender2k14:
Resolves #139
Resolves #151 
...or will by the time I am done.

This PR is currently a draft.  I just achieved the first change to the API, so it is a good time to share my progress.

The change to the API is that the `'a` type parameter in the `Binding.oneWay` method now includes the equality constraint.

Technically I did "manually" add the equality constraint to `'a` (elsewhere), but only because not adding it where I did would cause a compile error.  So I think the goal I had of having such type constraints included via inference is still achieved in this case in spirit.


bender2k14:
Now `'id` for `OneWaySeq` includes the equality constraint.

bender2k14:
I am finished.  This PR is ready for review.

The typing is much stronger now.  In particular, equality type constraints were added to six binding types.  Also, the definition of `SubModelSelectedItemBinding<...>` no longer uses `obj` for any type parameters.  I did not foresee that improvement.

The only change to the API is the added type constraints.  Any method on `Binding` that creates one of those six binding records now includes the equality constraint as well.

I just finished coding (and need to go now), so I didn't test the samples much.  I will do that later today.  The tests pass for me and the build server.  Overall, this PR is a bit "bigger"/risker, but I was very careful to make my changes slowly and in a good order.

As always, naming can probably be improved.  I am always open to suggestions.

I left several TODO comments along the way.  They are all minor.  I was not planning on doing them in this PR, but I could if that would be better.

cmeeren:
Thanks! I'll try to review it during the weekend. I suppose all the samples still run?

bender2k14:
> I suppose all the samples still run?

Yes.  I just tested all of them, and everything worked correctly.  Thanks for the reminder.

cmeeren:
I have done a very brief review (all I can prioritize right now) and it looks good. Merge when you're ready.

PR #254, opened 2020-08-06
switched to obtaining WPF dependencies via UseWpf=true in the test project

bender2k14:
When I build Elmish.WPF with Visual Studio, the official list of errors and warnings is empty.  However, the build output includes
```
20>------ Build started: Project: Elmish.WPF.Tests, Configuration: Debug Any CPU ------
20>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(2081,5): warning MSB3245: Could not resolve this reference. Could not locate the assembly "PresentationCore". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors.
20>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(2081,5): warning MSB3245: Could not resolve this reference. Could not locate the assembly "PresentationFramework". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors.
20>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(2081,5): warning MSB3245: Could not resolve this reference. Could not locate the assembly "WindowsBase". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors.
20>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(2081,5): warning MSB3243: No way to resolve conflict between "PresentationCore, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" and "PresentationCore". Choosing "PresentationCore, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" arbitrarily.
20>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(2081,5): warning MSB3243: No way to resolve conflict between "PresentationFramework, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" and "PresentationFramework". Choosing "PresentationFramework, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" arbitrarily.
[...]

PR #260, opened 2020-08-13
removed direct references to WPF DLLs

bender2k14:
This PR is similar to PR #254

This PR removes direct references to the WPF DLLs in the project file of the main project since these dependencies are also being picked up via `<UseWpf>true</UseWpf>`.

I will complete this PR after the automated build finishes successfully.


cmeeren:
Thanks a lot!

PR #262, opened 2020-08-15
changed some intermediate uses of List to Seq in merge

bender2k14:
This is a minor change.  I am making the PR into `v4` instead of `master` to avoid a future conflict since this change is in code that code moved.

I think I used `List` before because I knew that `[]` is an empty list but didn't know an equivalent expression for `Seq`.  Now I know about `Seq.empty` (and `List.empty`).

The purpose for making a PR of this minor change instead of just pushing it straight to `master` is a get some more eyes on it.  This is technically a small optimization in code that is already designed to be very efficient.

The diff seems a bit more complicated to me that the actual change, which is
- moved `Seq.toList` to the end,
- changed `List.collect` to `Seq.collect`,
- changed `[]` to `Seq.empty`, and
- changed `List.singleton` to `Seq.singleton`.

### DIFF
--- a/src/Elmish.WPF/Binding.fs
+++ b/src/Elmish.WPF/Binding.fs
@@ -136,14 +136,14 @@ module internal BindingLogic =
 
     let moves =
       additions
-      |> Seq.toList
-      |> List.collect (fun (Kvp (id, (sIdx, s))) ->
+      |> Seq.collect (fun (Kvp (id, (sIdx, s))) ->
         match removals.TryGetValue id with
-        | (false, _) -> []
+        | (false, _) -> Seq.empty
         | (true, (tIdx, t)) ->
             removals.Remove id |> ignore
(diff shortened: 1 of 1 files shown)


cmeeren:
> This is technically a small optimization in code that is already designed to be very efficient.

Since `Seq` is lazy, it has some overhead. I suggest that we run benchmarks on this to ensure that it is actually more performant. Would you be able to look into that?

bender2k14:
> Would you be able to look into that?

Will do

bender2k14:
Wow.  I just noticed that the build for this PR failed because of a failing test.  I will look into this soon.

bender2k14:
Yes.  This exact change is no good.  The data structure `additions` (which has type `Dictionary<_,_>`) is being both iterated over and mutated.

However, something better is probably possible.  I wrote down my thoughts in issue #275.

Issue #268, opened 2020-08-22
TabControl: Difficulty with DataTemplate?


awaynemd:
I'm a newbie. I do not know if this is my ignorance or an issue but would really appreciate some help .   The error from Visual Studio is:

> System.Windows.Data Error: 40 : BindingExpression path error: 'Name' property not found on 'object' ''ContactDetails' (HashCode=-728639187)'. BindingExpression:Path=Name; DataItem='ContactDetails' (HashCode=-728639187); target element is 'TabItem' (Name=''); target property is 'Header' (type 'Object')

Problem: How do I get WPF to see the  "Name" property on ContactDetails?

All suggestions are appreciated.

My WPF XAML is:

```xaml
<TabControl Grid.Row="1" ItemsSource="{Binding Details}">
            <TabControl.ItemContainerStyle>
                <Style TargetType="{x:Type TabItem}">
                    <Setter Property="Header" Value="{Binding Name}" />
                </Style>
            </TabControl.ItemContainerStyle>
            <TabControl.Resources>
                <DataTemplate DataType="{x:Type vm:ContactDetail}">
                    <vi:ContactDetailsView />
                </DataTemplate>
                <DataTemplate DataType="{x:Type vm:Internet}">
                    <vi:InternetView/>
                </DataTemplate>
                <DataTemplate DataType="{x:Type vm:PhoneNumber}">
                    <vi:PhoneNumbersView/>
                </DataTemplate>
                <DataTemplate DataType="{x:Type vm:Address}">
                    <vi:AddressesView/>
                </DataTemplate>
            </TabControl.Resources>
        </TabControl>
```

My Elmish.WPF code is:

```f#
namespace Models

open Elmish.WPF
open Elmish
open System
open System.Windows

type ContactDetail = { Name: string; Content: string; Text: string }
type Internet      = { Name: string; Content: string; Text: string }
type PhoneNumber   = { Name: string; Content: string; Text: string }
type Address       = { Name: string; Content: string; Text: string }

    module FrontOffice =    
        type Details =
            | ContactDetails of ContactDetail 
            | Internet       of Internet      
            | PhoneNumbers   of PhoneNumber   
            | Addresses      of Address       
    
        let contactDetail  : ContactDetail = { Name="Contact Detail"; Content="Content for Contact Detail"; Text="here is the contact detail text" }    
        let internet       : Internet = { Name="Internet"; Content="Content for Internet"; Text="here is the internet text" }
[...]

PR #273, opened 2020-08-29
My initial files -- The TabControl fails.

awaynemd:
@bender2k14  Here are the files you requested.  I'm happy to supply more information if needed. The problem is with how Elmish.WPF is doing the binding (I think?). I can either get "Devices" to correctly resolve the "Name" for the TabItem Header OR the DataTemplate x;Type  -- but NOT BOTH.  How can this be fixed?

Thanks for any help.


awaynemd:
@bender2k14 Need help with TabControl and DataTemplate resolution.

bender2k14:
Let's continue the discussion in issue #268.

PR #282, opened 2020-10-07
removed explicit dependency on WPF from sample F# project

TysonMN:
After we completed PR #249, it seems that we could have removed an explicit dependency on WPF from the sample F# projects.  The `SingleCounter` sample still works after making this change.

I tried making the same change to the sample C# projects and the `Elmish.WPF` project, but then the code doesn't compile.

How about I make this change to all the sample F# projects?

### DIFF
--- a/src/Samples/SingleCounter/SingleCounter.fsproj
+++ b/src/Samples/SingleCounter/SingleCounter.fsproj
@@ -1,8 +1,7 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.0</TargetFramework>
-    <UseWpf>true</UseWpf>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
   


cmeeren:
I don't think you can do that to all F# sample projects, because IIRC some of them use the WPF `Window` type. I'm split on whether we should remove it from the others; the safest thing is to have them there, so that it's easy for users to use WPF types in the F# projects when adapting the samples. How about this?

```diff
 <TargetFramework>netcoreapp3.0</TargetFramework>
+<!-- It is safe to remove <UseWpf> from the F# project file if it does not
+     use any WPF types and the C# project the entry point. -->
 <UseWpf>true</UseWpf>
 <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```

cmeeren:
Also, we should probably target .NET Core 3.1 now since 3.0 is out of support.

TysonMN:
> Also, we should probably target .NET Core 3.1 now since 3.0 is out of support.

Done via fe96ce4b0f149f6d16cc8d52ba0b6180259a8227

TysonMN:
Quoting https://github.com/elmish/Elmish.WPF/pull/282#issuecomment-705136308
> How about this?

I would prefer to leave it as is than add such a comment.  I think it would add some confusion for no real gain.

The only (admittedly minor) gain I see with removing the explicit WPF dependency is that it isn't necessary, so the `fsproj` file is slightly shorter/simpler.  Maybe it is necessary in some cases though.  I really only tried this one so far.

cmeeren:
I'm open to all three approaches (do nothing, remove, add comment). Feel free to do whatever you think is best.

PR #306, opened 2020-12-05
Adds SubModelSeq.List sample

TysonMN:
This PR adds a sample called SubModelSeq.List.  I tested it and it works.  There might be some unused code still to remove.


cmeeren (commented):
Some immediate thoughts just from looking at it on GH:

* In general, it would be good to clarify the scope/intention of the sample.
* Considering we want this sample to be fairly simple(?), could we remove the `Identifiable` wrapper? Is it actually helping in a significant way?
* I suggest moving all counter-related stuff to the `Counter` module, including the bindings. The counter is the "child component", so IMHO all its stuff should be kept together. Now of course, this immediately causes problems since the bindings rely on the root `Model`. So perhaps `Counter` isn't a truly separate child component after all? Perhaps we should remove `GlobalState` from this sample, to keep it simple and focus on the translator pattern if that's the intention?
* By the same reasoning, perhaps we could check for other things to simplify? Personally I find the `FuncOption` stuff and `swapCounters` to be a bit confusing.

TysonMN:
> * I suggest moving all counter-related stuff to the `Counter` module, including the bindings. The counter is the "child component", so IMHO all its stuff should be kept together. Now of course, this immediately causes problems since the bindings rely on the root `Model`. So perhaps `Counter` isn't a truly separate child component after all?

This is trivial to do in `v4` (and is already done for the `SubModelSeq` sample), but it is difficult to do in `master` (which lacks `mapModel`, `mapMsg`, and `mapMsgWithModel`).  It is possible to copy-paste the bindings _exactly_ as they are and put them in the `Counter` module.  However, the model type still needs to remain `_ * Identifiable<Counter>` and the message type needs to remain `InOutMsg<CounterMsg, _>`.

To see what I mean, I made that change in [this commit](https://github.com/elmish/Elmish.WPF/pull/306/commits/28bfccea934cad467aa097a068fa7e5e3e209b48).

> * Perhaps we should remove `GlobalState` from this sample, to keep it simple and focus on the translator pattern if that's the intention?

Sure.  I made that change in [this commit](https://github.com/elmish/Elmish.WPF/pull/306/commits/e259c39cf4729c9bbced036f407a9c9f15c53073).

> * By the same reasoning, perhaps we could check for other things to simplify? Personally I find the `FuncOption` stuff and `swapCounters` to be a bit confusing.

The easiest thing way to address that is to remove the ability to swap.  The would leave removal of the counter as the only out message.  All of these changes are easy to undo (or really, to redo) so I made that change in [this commit](https://github.com/elmish/Elmish.WPF/pull/306/commits/a67cfaf0ffc9e6291f99bfa234e72e48ccd6b892).

> ...could we remove the `Identifiable` wrapper? Is it actually helping in a significant way?

There are two things here.  First, the main benefit of an identifier is that it makes it easy to see what is changing when swapping counters.  However, if we can't swap counters, then this doesn't matter at all.  Second, assuming that we want an ID, it helps to keep the counter code identical to the `SingleCounter` sample.  This is more helpful in `v4` than it is in `master`.

I removed the identifiable wrapper in [this commit](https://github.com/elmish/Elmish.WPF/pull/306/commits/0156b17e19f0c27900bf3e6f13f9796a9f088bc5).

-------

After all those changes, I simplified the remaining code in [this commit](https://github.com/elmish/Elmish.WPF/pull/306/commits/140dc6462c70d58591fedaf33b18e9f6cce091fa).

src/Samples/SubModelSeq.List/Program.fs
```diff
@@ -0,0 +1,121 @@
...
+
+  type Model = Counter list
+
+  type Msg =
+    | CounterMsg of int * CounterMsg
+    | AddCounter
+    | Remove of int
+
+  type OutMsg =
+    | OutRemove
```
cmeeren, line 76: Instead of prefixing `OutMsg` cases with `Out`, how about removing the case prefix and using `[<RequireQualifiedAccess>]` on the DU?

src/Samples/SubModelSeq.List/Program.fs
```diff
@@ -0,0 +1,121 @@
...
+
+
+module Bindings =
+
+  open App
+
+  let counterBindings () : Binding<Model * int * Counter, InOutMsg<CounterMsg, OutMsg>> list =
+    [ "Remove" |> Binding.cmd(OutRemove |> OutMsg)
+    ] @ (Counter.bindings ())
+
```
cmeeren, line 98: To stick to what seems like the "canonical" translator pattern, I suggest that `Remove` should be part of `Counter.bindings`, and that these bindings accept the `InMsg` and `OutMsg` functions/constructors as parameters (either as individual parameters, a normal record, or an anonymous record, feel free to experiment).

src/Samples/SubModelSeq.List/Program.fs
```diff
@@ -0,0 +1,121 @@
...
+        (fun v _ -> v |> int |> SetStepSize |> InMsg))
+      "Reset" |> Binding.cmdIf(
+        Reset |> InMsg,
+        (fun (_, _, c) -> canReset c))
+    ]
+
+
+module App =
+
+  type Model = Counter list
```
cmeeren, line 68: I suggest having the main `Model` be a record with a single field `Counters`. When I learned Elmish, I was always annoyed at the countless (har-har) counter tutorials where the model was just an `int`; it made the rest of the code artificially simple and a bit harder to understand.

cmeeren (commented):
Thanks. I find it simpler now.

Regarding `Identifiable`: In my code at work, I have many domain entities that have an `Id` property of some type. I have also defined the following functions to help me use it without having to wrap everything in `Identifiable` or similar:

```f#
let inline getId x =
  (^a: (member Id: _) x)

module List =

  let inline mapById id f list =
    list |> List.map (fun x -> if getId x = id then f x else x)
```

You could of course also define a similar `removeById` function, or anything else you fancy.

That way, just by consistently making my domain entities contain a field named `Id`, I can treat them using this generic utility code while still keeping my domain models/logic clear of explicit wrappers. SRTP should be used carefully, and in my experience, this is one area that where it works well without adding notable complexity.

For this sample, barring `Identifiable<_>`, I see three options:
* Keep the PR as it is now: Don't have counter IDs; instead use indexes as an implicit identifier.
* Have an `Id` property in the `Counter` type and use the SRTP functions above to treat them generically.
* Don't treat them generically; remove the generic `List` functions currently in the PR and just remove the counters directly, non-generically (either inline or by creating `Counter`-specific functions).

I really like the SRTP approach for this, but if you think the third option is better here for simplicity, that's fine. I'm also open to keeping it like it is.

src/Samples/SubModelSeq.List/Program.fs
TysonMN: Sure.  Done.

src/Samples/SubModelSeq.List/Program.fs
TysonMN: Sure.  Done.

src/Samples/SubModelSeq.List/Program.fs
TysonMN: I don't care so much where we put things in the `master` version of this sample.  We when port this to `v4` though, I think we should remove the duplication present in the current branch by removing `InMsg` in all the places it is currently being used and replace `Counter.bindings ()` with
```fsharp
Counter.bindings () |> Bindings.mapMsg InMsg
```

Then case constructors `InMsg` and `OutMsg` are used right next to each other.

TysonMN:
> For this sample, barring `Identifiable<_>`, I see three options:
> 
> * Keep the PR as it is now: Don't have counter IDs; instead use indexes as an implicit identifier.
> * Have an `Id` property in the `Counter` type and use the SRTP functions above to treat them generically.
> * Don't treat them generically; remove the generic `List` functions currently in the PR and just remove the counters directly, non-generically (either inline or by creating `Counter`-specific functions).
> 
> I really like the SRTP approach for this, but if you think the third option is better here for simplicity, that's fine. I'm also open to keeping it like it is.

I prefer how it is now.

IDs are not essential to this sample (especially now that swapping counters isn't possible).  In my application at work, the places I am using this OutMsg with translator pattern only use `SubModel` bindings...not `SubModelSeq` bindings (when a button on one of my finite number of screens is clicked, which causes a different screen to appear).  To emphsize this, we could switch the sample to one that only uses `SubModel` bindings instead of `SubModelSeq` bindinds.

cmeeren:
> To emphsize this, we could switch the sample to one that only uses `SubModel` bindings instead of `SubModelSeq` bindinds.

Yes, that's possible. It would also remove some complexity and allow us to focus on the `OutMsg` stuff. Perhaps even renaming the sample `OutMsg`.

If we go down that route, we may consider still having a `SubModelSeq.List` sample that focuses exclusively on that and doesn't use `OutMsg`. What do you think?

TysonMN:
> If we go down that route, we may consider still having a `SubModelSeq.List` sample that focuses exclusively on `SubModelSeq` and doesn't use `OutMsg`. What do you think?

Yes, that is very reasonable.  The current `SubModelSeq` sample using a tree is very complicated.  I expect that most uses cases would be much simpler.  I know they are in my application at work.

So consider some potential `SubModelSeq.List` sample that focuses exclusively on `SubModelSeq` and doesn't use `OutMsg`.  In addition to it not using `OutMsg`, I don't want it to be possible to "simplify" anything using `OutMsg` either.  For example, the `SubModelSeq.List` sample in the branch for this PR would satisfy this if it were not possible to remove an item via a button on that item.

Another thing to consider is what the samples might look like when the composable binding API exists.  Still a bit beyond the fog of war for me to see, but I do expect the samples to change some due to the new API.  The existing functionality will still exist, so logically all the current samples could remain, but there will likely be a new ways that better demonstrate the new API.

cmeeren:
> In addition to it not using `OutMsg`, I don't want it to be possible to "simplify" anything using `OutMsg` either. For example, the `SubModelSeq.List` sample in the branch for this PR would satisfy this if it were not possible to remove an item via a button on that item.

Fair point. I'm fine with, uh, removing the removal feature.

>  The existing functionality will still exist, so logically all the current samples could remain, but there will likely be a new ways that better demonstrate the new API.

Yes, I think we should strive to keep the samples up-do-date with respect to best practices.

TysonMN:
I will keep the branch in my fork, but I will close this PR for now.  I expect other improvements to the samples soon, and I don't want this getting in the way.

PR #315, opened 2020-12-30
Support for multiple validation errors #314

### LINKED ISSUE #314: Multiple validation errors
The `INotifyDataErrorInfo` interface supports multiple errors per property.  Elmish.WPF is currently restricting this to a single error (see line 838).

https://github.com/elmish/Elmish.WPF/blob/db20825cc2057b9b464fe56a5752639dd7d9a480/src/Elmish.WPF/ViewModel.fs#L830-L839

The branch [demo/multiple_validation_errors](https://github.com/TysonMN/Elmish.WPF/tree/demo/multiple_validation_errors) contains a proof of concept that shows how to show multiple validation errors in WPF.  See the screenshot below.  The only hack there is the margin I added to the `Submit` button so that it wasn't covered by the additional validation errors.

Would anyone like this feature?

I think this feature can be added while remaining backwards compatible, so I think it is fine to exclude this feature from `v4`.

TysonMN:
Resolves #314

This PR adds support for multiple validation errors.

Here are some interesting points about this implementation.
- Other than adding to the API support for a list of errors, I left the existing API as is.  Internally, I converted each `Option` and `ValueOption` to a `list`.
- Equality of errors is done by the equality for `string list`.  This means that the order of errors matters.  I think that is reasonable.
- The sample still includes the previous code in which there is only one validation error.  I added another text box that includes multiple validation errors.
- To quickly get this PR working, the `Submit` message in the sample no longer takes in the parsed data.  However, doing so is preferred.  As Mark Seemann [recently said](https://blog.ploeh.dk/2020/12/14/validation-a-solved-problem/), validation is a solve problem via applicative functors.  Unfortunately, F# doesn't include a `Validation<,>` monad.  I have created one in my project at work.  I could add that to the sample but that would also increase the learning curve.  Alternatively, one could just parse the input under the assumption it will succeed (via `int`) because the `canExec` function returned `true`.  This is not a good practice to do everywhere though.
- The red box around the text box containing the password is too long on the right side.  I don't know why and didn't try to fix it either.

### DIFF
--- a/src/Elmish.WPF.Tests/BindingTests.fs
+++ b/src/Elmish.WPF.Tests/BindingTests.fs
@@ -866,7 +866,7 @@ module twoWayValidate =
         let! x = GenX.auto<int>
         let! err = GenX.auto<string>
 
-        let validate x = if x < 0 then ValueSome err else ValueNone
+        let validate x = if x < 0 then [ err ] else []
         let d = Binding.twoWayValidate(fail, fail2, validate) |> getTwoWayValidateData
 
         test <@ d.Validate x |> unbox = validate x @>
@@ -917,10 +917,10 @@ module twoWayValidate =
         let! x = GenX.auto<int>
         let! err = GenX.auto<string>
 
-        let validate x = if x < 0 then Some err else None
+        let validate x = if x < 0 then [ err ] else []
         let d = Binding.twoWayValidate(fail, fail2, validate) |> getTwoWayValidateData
 
-        test <@ d.Validate x |> unbox = (validate x |> ValueOption.ofOption) @>
+        test <@ d.Validate x |> unbox = validate x @>
       }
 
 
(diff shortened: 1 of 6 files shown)


cmeeren:
The CI build fails, could you investigate?

TysonMN:
Ah, thanks for pointing that out.  Will do.

TysonMN:
CI build passes now.  I forgot to update the tests to match the new behavior.  I force pushed the needed changes.

cmeeren:
I pushed a few commits. What do you think?

The rectangle width error was because the error message was wider than the input field. I increased the width of the input field.

TysonMN:
Great improvements.  Looks good to me.

Ready to merge?

cmeeren:
Yep, thanks a lot!

PR #320, opened 2021-01-14
Use App.xaml in samples

### LINKED ISSUE #319: Improve communication about application-level resources
This is a spin off of #318.

To help others avoid the "mistake" I made with application-level resources (as described in #318), let's change the samples to have placeholders for application-level resources and also provide some explanation in the readme and/or tutorial.

cmeeren:
Closes #319


cmeeren:
I started with [this commit](https://github.com/cmeeren/Elmish.WPF/commit/2d79fd829ef1d5f6d91120562b3c465d73ac30a4), but I'm having trouble reducing noise in `App.xaml`. If I remove anything that's now there, from the innermost `<ResourceDictionary />` and outwards, compilation invariably fails in `App.xaml.cs` with the error `The name 'InitializeComponent' does not exist in the current context`.

image

@TysonMN any ideas?

TysonMN:
I don't have that problem on your branch.  After stashing or committing all changes, try closing Visual Studio and [running `git clean -fdx`](https://tysonwilliams.coding.blog/2020-07-11_systematic_cleaning).

cmeeren:
I often do, and it doesn't help this time. Check out the latest commit, which fails for me even after `git clean -fdx`.

TysonMN:
I am now experiencing that problem.  Not sure what changed.  Investigating now.

TysonMN:
`App.g.i.cs` in the `obj` folder has an entry point and is missing `Initialize Component`.  This makes me think that the `<StartupObject>` line isn't being respected.  Try targeting .NET Framework or Core to see if that works.

TysonMN:
I pushed commit https://github.com/elmish/Elmish.WPF/pull/320/commits/a51c734cbb4ba84d5a66633d52aafae1ee80aa89 to your branch.  It compiles for me now.

It seems like `<StartupObject>` only works if `App.Resources` contains merged dictionaries with at least one of which is nonempty.  This has to be a WPF complication bug.

cmeeren:
I see. Do you want to report it, and then we can link to the issue in a comment in `App.xaml`?

TysonMN:
I don't understand.  [This code](https://github.com/TysonMN/WPF_StartupObject_bug) compiles.

TysonMN:
Oh, I wasn't calling `InitializeComponent` in the constructor of `App`.

TysonMN:
Wait, why do you have `InitializeComponent` in the constructor of `App`?  If you create a new WPF application in Visual Studio, then you will see that the template does not include a call to `InitializeComponent` in the constructor of `App`.  This makes me think that `App` should not call `InitializeComponent` in its constructor.

My guess is that you added `App.xaml` to the project by creating a `Window` or `UserControl`, both of which do include a call to `InitializeComponent` in their constructor, and then you renamed the element to `Application` but did not delete the constructor or the call to `InitializeComponent`.

cmeeren:
Yes, you are right. I assumed all code-behind partial classes needed `InitializeComponent`. So I should simply remove `InitializeComponent`? (And even the constructor? What about the entire code-behind file? Even if it's possible, should we keep the file and/or constructor to make it easier to "plug in" functionality here [as I did here](https://github.com/cmeeren/Elmish.WPF/commit/2d79fd829ef1d5f6d91120562b3c465d73ac30a4#diff-b766b26edc4976be892304352cd6a72569b6d3d40240c90bc63e1b67494d2a90)?)

TysonMN:
I just tested with your [`app-xaml-startup` branch](https://github.com/cmeeren/Elmish.WPF/tree/app-xaml-startup).  If `InitializeComponent` is removed from the constructor of `App`, then the themes from Material Design in Xaml are not used.  In that sense, calling `InitializeComponent` is necessary.  However, it seems that the compiler doesn't create that method if it thinks it isn't necessary.  Maybe there is a bug related to exactly which edge cases cause that method to exist or not exist.

I suggest that we find some code that causes `InitializeComponent` to be generated and include a call to it in the constructor of `App`.

cmeeren:
> I suggest that we find some code that causes `InitializeComponent` to be generated and include a call to it in the constructor of `App`.

So, pretty much like https://github.com/elmish/Elmish.WPF/commit/a51c734cbb4ba84d5a66633d52aafae1ee80aa89?

TysonMN:
Yea, that seems to work.

cmeeren:
Great. I'll put that and a comment in all the samples, then.

BTW, IIRC `InitializeComponent` is the call that actually kind of loads/initializes the XAML file in the first place.

cmeeren:
Ok, I have experimented a bit more now. There are several ways to go about this. I found a fairly simple way that works with `StartupUri`, which AFAIK is what you originally were looking for. I think it's good to roll with the defaults you get when you create projects in VS, and this one also avoids the InitializeComponent issue: First of all it compiles with `InitializeComponent` and an empty resource dictionary, and second of all you don't even need `InitializeComponent`; resources work correctly without it.

See the latest commit above this comment (b9693c2). In short:
* Removed `Program.cs`
* Added `StartupUri="MainWindow.xaml"` in `App.xaml`
* Changed `Program.runWindowWithConfig` to `Program.startElmishLoop`
* On the first `Activated` event of `App`, start the Elmish loop with the app's `MainWindow`

In v4, we might consider renaming `Program.startElmishLoop` to `Program.startElmishLoopWithConfig` and adding a new `Program.startElmishLoop` without the config argument.

There are [alternatives to the `Activated` event](https://stackoverflow.com/a/29224248/2978652), but I like it best.
* Using the `Loaded` event of the `MainWindow` also seems fairly clean, but it "feels" better to keep app startup stuff in `App`. Feel free to disagree.
* Not using `StartupUri` is another option, but then we're back to what we have discussed previously.

Thoughts?

cmeeren:
@TysonMN To be clear, I'm waiting for feedback from you before continuing. In particular, I'd like to know if the method from the latest commit works for your use-cases, and if you can see any obvious negatives or potential improvements.

TysonMN:
Yes, sorry for the delay.

> I found a fairly simple way that works with `StartupUri`, which AFAIK is what you originally were looking for.

I wrongly assumed that `StartupUri` was necessary to have application-level resources.

I have been thinking a lot about this and going back and forth on what I think is best.  Here is the comparison as I see it.

### `StartupUri` and `Activated` event
**Advantages**
1. `App.InitializeComponet` is (created and) called if necessary (see `obj\Debug\net5.0-windows\App.g.i.cs`).
2. Most similar to how a traditional WPF application starts.

**Disadvantages**
1. Subscribing to an event for only one raising is a bit hacky.
2. Event subscription and unsubscription feels brittle (compared to the same in F# where subscribing to an event returns an instance of `IDisposable` that unsubscribes when `Dispose` is called).

### Custom entry point
**Advantages**
1. Very clear about how the flow of control starts and progresses.

**Disadvantages**
1. Need for nontraditional `<StartupObject>` entry in the project file.
2. Need an explicit call to `App.InitializeComponent` when application-level resources exist.

I think the approach using `StartupUri` and `Activated` event is better for newcomers because of the two advantages given above.  At the same time, I think I will continue using the approach with a custom entry point for my application at work because one the one advantage given above.

What follows is an extended exposition about why I will continue using the approach with a custom entry point.

I strongly dislike all the "magic" that Microsoft puts into their code.  Especially before Entity Framework Core, it was difficult to provide configuration options because some of the configuration was expected to just exist in the same assembly and it would found at runtime via reflection.  Similarly, ASP.NET requests a "startup" type that doesn't have to implement any interface.  Instead, it is just assumed that there will exist an instance method (or more?...I don't even know) with a certain name and certain inputs, and it will get called via reflection.  (One could argue that the example with ASP.NET is a case of [structural typing](https://en.wikipedia.org/wiki/Structural_type_system) instead of the predominant use of [nominal typing](https://en.wikipedia.org/wiki/Nominal_type_system) in all of .NET, but one would be wrong because the structure of the requested type isn't given either; it is just "magic".)

I do a really good job logging anything that can go wrong in my application at work.  (I will eventually have a long series of blog posts about that.)  This starts by configuring the logging before anything has an opportunity to go wrong.  When using the traditional WPF entry point, not only is it a bit unclear what happens in what order (though I now know that the entry point is in `obj\Debug\net5.0-windows\App.g.i.cs`), but it is impossible to configure logging before the [static](https://github.com/dotnet/wpf/blob/bbd9e44d87b4edd8880f4c09c6db36a2f3e23ba8/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Application.cs#L96) and [instance](https://github.com/dotnet/wpf/blob/bbd9e44d87b4edd8880f4c09c6db36a2f3e23ba8/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Application.cs#L104) constructors of `Application` execute.  I don't know of anything that can go wrong in them or where they log if something does go wrong, but it shouldn't be necessary to answer that questions in order to have a high degree of confidence in your application.  (Also, I looked at the [source code for `Application`](https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Application.cs), but I can't figure out where the `MainWindow` is instantiated.)

Here is a specific reason I want to continue using the custom entry point approach in my application at work.  Constructing the `MainWindow` might fail because of some XAML problem even though the solution built successfully.  I have a test that checks for this (also a future blog post).  However, I sometimes makes a change and run the application before running tests.  In this case, my logging gets configured first, which includes logging to [Seq](https://github.com/serilog/serilog-sinks-seq).  Then when the application begins to crash by breaking in Visual Studio and displaying exception messages, I know that all of this is being logged, so I don't have to pay attention to every detail in the moment.  I can make an assumption and try to fix the problem given that assumption.  If that assumption is wrong, I can review exactly what went wrong in Seq instead of spending time trying to reproduce the problem.  I could still have issues that occur when constructing the `MainWindow` logged by configuring logging in the constructor of `App`.  However, then it would not be so clear that this is happening at the right time.  Most of the time nothing is logged.  Is that because everything is working or because logging is not configured correctly?  This is why TDD wants you to see the test fail _for the right reason_ before making it pass by changing the production code...because we want the test to pass _for the right reason_.

cmeeren:
> **Disadvantages**
>
> 1. Subscribing to an event for only one raising is a bit hacky.
> 2. Event subscription and unsubscription feels brittle (compared to the same in F# where subscribing to an event returns an instance of `IDisposable` that unsubscribes when `Dispose` is called).

Generally I don't agree with this. I do agree somewhat with the "feeling", and that the `IDisposable` method feels "safer" in the sense that unsubscription is guaranteed by language constructs. But I don't have any actual arguments against this pattern as used here. IIRC event handlers are called synchronously, so it's entirely deterministic and not brittle at all. And the control flow isn't particularly complicated in this case. Sure, I wouldn't base my business logic around events being subscribed and unsubscribed willy nilly, but that's not relevant here.

Of course, the above is based on assumptions about how the `Activated` event works. Would I stake my life on it? No way. Would I stake the startup code of a production app on it? Given that 1) [the official documentation](https://docs.microsoft.com/en-us/dotnet/api/system.windows.application.activated) fits our use case, 2) it was suggested by a seemingly well written SO answer, and 3) it's observed to work, then yes. (In any case, this doesn't have anything to do with the fact that it's an event; this goes for any API.)

> **Disadvantages**
> 
>   1. Need for nontraditional `<StartupObject>` entry in the project file.
>   2. Need an explicit call to `App.InitializeComponent` when application-level resources exist.

I think the second one here is a fairly major disadvantage, since according to our observations,
1. it is not possible to call `InitializeComponent` if there are no resources (for certain vague and not entirely logical definitions of "no resources"), and
2. if you forget to call it when it's available, resources won't be loaded.

> I think the approach using `StartupUri` and `Activated` event is better for newcomers because of the two advantages given above.

We agree on this point. I take it that you find it okay to make the samples use this pattern, then? We could of course also have a sample that just demonstrates the explicit entry point variant.

> What follows is an extended exposition about why I will continue using the approach with a custom entry point.

Thanks for the detailed explanation. I can relate to the desire for removing all of the "magic".

TysonMN:
> I take it that you find it okay to make the samples use this pattern, then? We could of course also have a sample that just demonstrates the explicit entry point variant.

Indeed.  One sample using the custom entry point would be nice but not necessary.

cmeeren:
I have yet to update all samples, but I have made some important changes I'd like input on:

* Changes to `Program.fs`: `startElmishLoop` is deprecated and renamed to `runElmishLoopWithConfig`, and I have added `runElmishLoop` as a config-less alternative. I think this is a more graceful way to introduce the breaking change mentioned in https://github.com/elmish/Elmish.WPF/pull/320#issuecomment-761209461. I think `run` is just as good as `start` (and has a parallel in `runWindow`), but I'm open to suggestions.
* Updated the doc comments of several `Program.fs` functions
* Updated the Getting started instructions in the readme
* I cleaned up the sample code (removed the test resource), and I think the sample now represents how I want all samples to be adjusted in this PR.

TysonMN:
I will look closely tomorrow.

I still like the name `startElmishLoop`.

cmeeren:
> I still like the name `startElmishLoop`.

Same here, I just wanted to have that name be "config-less" and rename the existing one to `startElmishLoopWithConfig`. That is a breaking change, so I figured that by having other names, it could be done by way of deprecations. But I'm certainly open to other suggestions, including simply keeping the existing name and then breaking it in v4. It's not a change that will cause much churn for users, so I'm not too concerned about it.

TysonMN:
Sure.  I don't mind breaking changes when, as you said, the fix is straightforward.

I have reviewed now.  This looks good.  I especially like the comments in the project file.  That is a nice touch.

cmeeren:
> I have reviewed now. This looks good.

It is my understanding then you can get behind all the current PR changes to `Program.fs`. Is that correct?

TysonMN:
I still prefer the name `startElmishLoop` instead of `runElmishLoop`, but I am indifferent enough that I don't mind deferring that decision to you.

I am happy with all the other changes.

cmeeren:
> I still prefer the name `startElmishLoop` instead of `runElmishLoop`, but I am indifferent enough that I don't mind deferring that decision to you.

I agree. Perhaps I'll just revert the rename and remove the config-less variant for now, and we can bring it back in v4 with a breaking change. The required change for users would be trivial.

cmeeren:
This is now ready for review.

I have not squashed the commits; I plan to do that in GH when merging.

cmeeren:
Forgot to mention that I have tested all the samples.

README.md
```diff
@@ -112,7 +108,9 @@ See the [SingleCounter](https://github.com/elmish/Elmish.WPF/tree/master/src/Sam
      | SetStepSize x -> { m with StepSize = x }
    ```
 
-6. Define the “view” function using the `Bindings` module. This is the central public API of Elmish.WPF. Normally in Elm/Elmish this function is called `view` and would take a model and a dispatch function (to dispatch new messages to the update loop) and return the UI (e.g. a HTML DOM to be rendered), but in Elmish.WPF this function is in general only run once and simply sets up bindings that XAML-defined views can use. Therefore, let’s call it `bindings` instead of `view`.
+6. Define the “view” function using the `Bindings` module. This is the central public API of Elmish.WPF.
+
+   Normally in Elm/Elmish this function is called `view` and would take a model and a dispatch function (to dispatch new messages to the update loop) and return the UI (e.g. a HTML DOM to be rendered), but in Elmish.WPF this function is in general only run once and simply sets up bindings that XAML-defined views can use. Therefore, let’s call it `bindings` instead of `view`.
```
TysonMN, line 113: > in general only run once

My guess is that you mean it would be run exactly once per call to `startElmishLoop` (or whatever we are calling it now) and `startElmishLoop` is typically called once per process execution.

Is my guess correct?  If so, then I suggest simplifying this language by removing "in general".

```suggestion
   Normally in Elm/Elmish this function is called `view` and would take a model and a dispatch function (to dispatch new messages to the update loop) and return the UI (e.g. a HTML DOM to be rendered), but in Elmish.WPF this function is only run once and simply sets up bindings that XAML-defined views can use. Therefore, let’s call it `bindings` instead of `view`.
```

README.md
```diff
@@ -191,14 +211,18 @@ In short, for WPF apps, a solution based on static XAML views is currently the w
 
 #### Do I have to use the project structure outlined above?
 
-Not at all. The above example, as well as the samples, keep everything in a single project for simplicity (the samples have the XAML definitions in separate projects for technical reasons). For more complex apps, you might want to consider a more clear separation of UI and core logic. An example would be the following structure:
+Not at all. The above example, as well as the samples, keep all non-UI code in a single project for simplicity, and all the XAML in a C# project for better tooling.
+
+An alternative with a clearer separation of UI and core logic can be implemented by splitting the F# project into two projects:
```
TysonMN, line 216: FWIW, I use three projects here:
1. `1_Pure`.  It has no dependencies (other than `FSharp.Core`) and contains types Model, Msg, Cmd (and maybe OutMsg and OutCmd) and functions/data init and update.  Those command types are part of the CmdMsg pattern.
2. `2_Impure`.  It has a dependency on `Elmish` (and others like `Serilog`) and maps my Cmd types to Elmish.Cmd types via a function called `bindCmd`.
3. `3_Bindings`.  It has a dependency  on `Elimish.WPF` (and others like `SkiaSharp`) and defines the bindings and design-time view models.

Then some other projects:
0. `0_Library` containing no dependencies, pure code, and project independent code.  Mostly code to support functional programming, but also some code for mathematics (such at geometery and linear algebra) and improving the .NET or `FSharp.Core` API.
4. `4_Views` containing XAML and a few C# classes.
5. `5_Testing.UnitTests`
6. `6_Testing.IntegrationTests`

src/Elmish.WPF/Program.fs
```diff
@@ -5,9 +5,9 @@ open System.Windows
 open Elmish
 
 
-/// Starts the Elmish dispatch loop, setting the bindings as the DataContext
-/// for the specified FrameworkElement. Non-blocking. This is a low-level function;
-/// for normal usage, see runWindow and runWindowWithConfig.
+/// Starts an Elmish dispatch loop, setting the bindings as the DataContext for the
+/// specified FrameworkElement. Non-blocking. If you have an explicit entry point where
+/// you control app/window instantiation, runWindowWithConfig may be a better option.
```
TysonMN, line 10: I prefer "might be" over "may be" since "may be" somewhat looks and definitely sounds like "maybe".
```suggestion
/// you control app/window instantiation, runWindowWithConfig might be a better option.
```

TysonMN (approved):
Very good!  I left some minor feedback.

README.md
cmeeren: It is not entirely correct. Sub-model bindings can be initialized multiple times (whether `subModel` or `subModelSeq`, recursive or not).

README.md
cmeeren: Sure. My intention is just to "whet their appetite" by suggesting that other project structures are possible, e.g. with the aim of getting a clearer separation of UI and core logic, or with an alternative entry point. I think users, when they are ready for it, are able to make the decision to split the projects up further or make adjustments. But if you feel the readme should be modified, let me know.

README.md
TysonMN: Sure, that makes sense.  I was thinking about the top-level function.

README.md
TysonMN: No changes necessary.  Just sharing some thoughts that I had while reading that part.

TysonMN:
I think this is ready to be merged

cmeeren:
Great, thanks!

Issue #331, opened 2021-02-05
Get the model from event streams


Evangelink:
I have started to implement keyboard shortcuts in my application which I am handling through the dispatch mechanism:

```fsharp
let paste dispatch (obs: IObservable<KeyEventArgs>) =
	obs
    |> Observable.filter (...)
    |> Observable.add (fun _ -> MyMessage |> dispatch)

window.KeyDown |> paste MyDispatch
```

(note that I am using observables to make it easy to test the behavior) and I also have a button for the same command. The button is bound using `Binding.cmdIf` which prevents the action under some context but the shortcut can't do the same.

So far, I am handling the "invalid" case in the update and return `model, []` in case the state is not good but I would prefer to be able to create some kind of observable state of my model which I could plug into my observable streams.

Any suggestion?

TysonMN:
Can you share a link to a GitHub repo that demonstrates what you are currently doing and maybe includes some comments about how you would like it to be different?

Evangelink:
I have created a dummy project to showcase the behavior. Let me know your thoughts.

Here is the link https://github.com/Evangelink/Elmish.Wpf.Experiments

TysonMN:
I understand now.  Thanks for creating a project that I can run.

I think you could achieve your desired behavior given the ability to dispatch a list of messages instead of just a single message.  Of course a list can be empty, so that would give you the ability to dispatch no messages.  See #321.  I also expect the message-mapping feature in `v4` would be useful here.  See #263.

Your subscription to the `KeyDown` event is "global".  Suppose you had two counters but only one visible at a time.  Could the pressing `Ctrl+D` be configured so that only the visible counter gets incremented?  Of course this could be done in the F# code with filtering like you have done.  I am wondering if binding to this event in the XAML would narrow the scope of the subscription.

Evangelink:
In my real app, I am trying to mimic a bookmark feature like in web browsers so I really want the Ctrl + D to work from anywhere within the app and not only when a specific part of the app is focused. Of course, it needs to work only when the page is "active".

Regarding the access to an `IObservable<Model>`, I don't know why I didn't think about using a `Subject<Model>` earlier... See https://github.com/Evangelink/Elmish.Wpf.Experiments/pull/2

> I think you could achieve your desired behavior given the ability to dispatch a list of messages instead of just a single message. Of course a list can be empty, so that would give you the ability to dispatch no messages. See #321.

[...]

PR #336, opened 2021-02-10
Enrich event bindings example with KeyBinding

Evangelink:
Follow-up PR from thread #331

### DIFF
--- a/src/Samples/EventBindingsAndBehaviors/MainWindow.xaml
+++ b/src/Samples/EventBindingsAndBehaviors/MainWindow.xaml
@@ -12,6 +12,9 @@
         WindowStartupLocation="CenterScreen"
         mc:Ignorable="d"
         d:DataContext="{x:Static vm:Program.designVm}">
+  <Window.InputBindings>
+    <KeyBinding Command="{Binding ToggleVisibility}" Gesture="Ctrl+T" />
+  </Window.InputBindings>
   <StackPanel Margin="15" HorizontalAlignment="Center">
     <TextBlock Text="Switch focus between these two inputs." Margin="0,0,0,10" TextAlignment="Center" />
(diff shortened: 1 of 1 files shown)


cmeeren:
Hm, it seems that I have misunderstood. This functionality does not seem to depend on event bindings at all. If that is correct, I don't think it belongs in this sample, which specifically demonstrates event bindings using `Microsoft.Xaml.Behaviors.Wpf`. I'd like to keep it focused for that.

If keyboard shortcuts are this simple and already play nice with bindings/Elmish.WPF, do we need this in a sample at all?

Evangelink:
It's indeed that simple and purely wpf related. I totally forgot about this wpf feature and went with a more complex approach on the model (unneedingly complex).

Feel free to close the PR if you think that does not bring anything.

TysonMN:
> But if keyboard shortcuts are this simple and already play nice with bindings/Elmish.WPF, do we need this in a sample at all?

We don't. The reason we thought it might be useful to include this approach to keyboard shortcuts in this sample is that another way to do keyboard shortcuts is to subscribe to the `KeyDown` event.

cmeeren:
Ok. I suggest a separate (very simple) sample dedicated to keyboard shortcuts, then. What do you think?

The sample can, for example, just display a checkbox that can be toggled with a shortcut like Ctrl-Something, together with (visible, not tooltip) text that describes that it can be toggled with this shortcut.

TysonMN:
I want to capture this key binding knowledge, so I am merging this PR.

Evangelink:
Sorry guys, I got really busy at work lately and a bit un-motivated at home to do some extra coding. I will try to get back to you as soon as possible.

TysonMN:
No problem @Evangelink. We are grateful for any amount of help that your provide. :)

PR #338, opened 2021-02-13
changed SubModel sample to use radio buttons in recommended way #283

### LINKED ISSUE #283: Recommended way to use RadioButton
I just realized something that seems obvious in hindsight.  I think the correct way to use a `RadioButton` with Elmish.WPF is to do a `OneWay` binding to the `IsChecked` property and then use a `Cmd` binding to the `Command` property to handle click events.

I created [this repository](https://github.com/TysonMN/Elmish.WPF_RadioButtons) to demonstrate this approach, which I called `SetViaCommand`, and contrast it with the two other approaches that I have used.

Below is a somewhat rambling explanation of the issue with using `TwoWay` bindings to the `IsChecked` properties.  Then I close with short paragraph and some questions for discussion.

---------------------

For a long time now, I have been using a `TwoWay` binding to the `IsChecked` property of `RadioButton`.  In my model, I store the state of the radio button group as a single instance of a discriminated union: one case for each radio button.  If the use case allows no radio button to be selected, then I make the discriminated union optional in the model.

If the discriminated union is optional, then there is no issue.  But what if you want to model a discriminated union that is not optional?

Suppose the XAML look like this.
```fsharp
<StackPanel Orientation="Vertical">
  <StackPanel Orientation="Horizontal">
    <RadioButton IsChecked="{Binding IsA}" GroupName="DU"/>
    <TextBlock Text="A"/>
  </StackPanel>
  <StackPanel Orientation="Horizontal">
    <RadioButton IsChecked="{Binding IsB}" GroupName="DU"/>
    <TextBlock Text="B"/>
  </StackPanel>
</StackPanel>
```

Now compare these two cases.

**Case 1:**
State: `B` is selected.
User interaction: `A` is clicked.
WPF behavior:
1. Binding for `IsChecked` of `A` is set to `true`
2. Binding for `IsChecked` of `B` is set to `false`

**Case 2:**
State: `A` is selected.
User interaction: `B` is clicked.
WPF behavior:
1. Binding for `IsChecked` of `A` is set to `false`
2. Binding for `IsChecked` of `B` is set to `true`

Suppose the DU is optional.  Then when `false` is set for some binding, I check if the DU is in the case corresponding to the binding.  If so, then I set the optional DU to `None`.  If not, then I leave the DU unchanged.

What if the DU is not optional?  Then when `false` is set for some binding, it should always be a no-op.  I achieved this via an explicit `NoOp` message that returns the model given as input.  As I type this, another idea that comes to mind is to adjust the message that sets the DU to have an optional instance of the DU.  Then if I want the behavior to be a no-op, I could pass in that message containing the value `None`.

----------------------

However, none of these compromises is ideal.  For performance reasons but more importantly for the experience during debugging or when reading logs, it is preferable to not have no-op behavior.  Let me say that another way.  Instead of dispatching a message that when handled by `update` causes the same model to be returned as the one given, it is preferable to not have dispatched a message at all.  That is what my new approach to using `RadioButton`s does.

How obvious is this to others?  Am I the only one who didn't think (until now) of this approach to using `RadioButton`s?  Is there an even better way to use `RadioButton`s that I still haven't thought of?

TysonMN:
Resolves #283

I tested the sample.  It works great.

### DIFF
--- a/src/Samples/SubModel.Core/Program.fs
+++ b/src/Samples/SubModel.Core/Program.fs
@@ -44,29 +44,38 @@ module Counter =
 
 module Clock =
 
+  type TimeType =
+    | Utc
+    | Local
+
   type Model =
     { Time: DateTimeOffset
-      UseUtc: bool }
+      TimeType: TimeType }
 
   let init () =
     { Time = DateTimeOffset.Now
-      UseUtc = false }
+      TimeType = Local }
 
   let getTime m =
-    if m.UseUtc then m.Time.UtcDateTime else m.Time.LocalDateTime
+    match m.TimeType with
+    | Utc -> m.Time.UtcDateTime
+    | Local -> m.Time.LocalDateTime
 
   type Msg =
     | Tick of DateTimeOffset
-    | ToggleUtc
+    | SetTimeType of TimeType
 
   let update msg m =
     match msg with
     | Tick t -> { m with Time = t }
-    | ToggleUtc -> { m with UseUtc = not m.UseUtc }
+    | SetTimeType t -> { m with TimeType = t }
 
   let bindings () : Binding<Model, Msg> list = [
     "Time" |> Binding.oneWay getTime
-    "ToggleUtc" |> Binding.cmd ToggleUtc
+    "IsLocal" |> Binding.oneWay (fun m -> m.TimeType = Local)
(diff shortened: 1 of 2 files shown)


cmeeren:
Thanks!

PR #341, opened 2021-02-14
switched automated builds to GitHub Actions #332

### LINKED ISSUE #332: Switch DevOps to GitHub Action
I would prefer to to have the automated builds run by GitHub.  `fsharp-hedgehog` and `fsharp-hedgehog-experimental` use a certain Action that works very well.

TysonMN:
Resolves #332

Haven't tested yet.  I think the build on this PR will use the new action.


TysonMN:
The build on GitHub didn't happen.  The AppVeyor build started because I haven't touched the webhook but failed because I deleted `appveyor.yml `.

I created a [parallel PR to this one in my fork](https://github.com/TysonMN/Elmish.WPF/pull/4), which doesn't include the AppVeyor webhook.  The GitHub build did happen.  [It failed though](https://github.com/TysonMN/Elmish.WPF/runs/1898750579?check_suite_focus=true).  Troubleshooting that now.

TysonMN:
Ah.  I accidently removed the checkout step.  Making progress now.

TysonMN:
The [parallel PR in my fork](https://github.com/TysonMN/Elmish.WPF/pull/4) now [builds successfully](https://github.com/TysonMN/Elmish.WPF/pull/4/checks?check_run_id=1900168749).

This PR is now ready for review.

After this code is merged into `master`, I will delete the AppVeyor webhook.

cmeeren:
Thanks for the work!

I notice that there is much more CI code now than with AppVeyor. I also think a lot of it can be simplified.

In particular:

* Is it really necessary to have a manual check-out step with GH actions? All CI systems do this automatically, and one would think GH of all CI systems would definitely do this.
* Is it necessary to install the .NET Core SDK? On most CI systems, that's already present.
* You don't need an explicit restore step. `dotnet build` already runs `dotnet restore`. Heck, `dotnet test` runs `dotnet build`, so you don't actually need more than `dotnet test`. Just point it at the solution file, and it should work fine. If you feel strongly about having separate build and test steps, I can accept that. I just don't see the point, because the log will contain all relevant details anyway in case of a failure.
* Is the nuget.config needed? Could it be simplified? It seems to contain a lot of default stuff. For the part of it that can't be simplified, could you perhaps add comments above the various nodes explaining why they have to be there? That would help any maintainer/contributor looking at the repo.
* Can you add an empty line between all steps (and above the first step)? It makes the files it easier to read.
* Can you add an empty line at the end of the files? It removes the ugly icon GH shows (at least in the diffs) when there is no newline at the end of the file.

TysonMN:
> * Is it really necessary to have a manual check-out step with GH actions?

Yes.  If I [remove the checkout step](https://github.com/TysonMN/Elmish.WPF/pull/4/commits/4bc53c89599e022823b18e667a996371c17add4e), then [the restore fails because it can't find the project (actuall, the solution) file](https://github.com/TysonMN/Elmish.WPF/pull/4/checks?check_run_id=1949057290).  I agree with you though; the need for this step also surprised me.

> * Is it necessary to install the .NET Core SDK?

If I [remove the installation of version 3.0.x of the .NET Core SDK](https://github.com/TysonMN/Elmish.WPF/pull/4/commits/48b0163733708a33537b46c8dd87716ca57eded7), then [the tests targeting 3.0 fail](https://github.com/TysonMN/Elmish.WPF/pull/4/checks?check_run_id=1949079431).  The [tests still fail](https://github.com/TysonMN/Elmish.WPF/pull/4/checks?check_run_id=1949102629) even when [the tests target .NET Core 3.1 instead of 3.0](https://github.com/TysonMN/Elmish.WPF/pull/4/commits/4c8f2b8cf6754deabeb72b34d1faca93e78d2b9e).

If I remove the installation of version 5.0.103 of the .NET SDK](https://github.com/TysonMN/Elmish.WPF/pull/4/commits/68239990aa387247ba0c20a139b62ea5ac41cbe8), then [the restore step fails](https://github.com/TysonMN/Elmish.WPF/pull/4/checks?check_run_id=1949130530).

> You don't need an explicit restore step. `dotnet build` already runs `dotnet restore`. Heck, `dotnet test` runs `dotnet build`, so you don't actually need more than `dotnet test`. Just point it at the solution file, and it should work fine. If you feel strongly about having separate build and test steps, I can accept that. I just don't see the point, because the log will contain all relevant details anyway in case of a failure. I like to keep non-essential stuff like CI config as simple as possible.

Correct.  The point is that separating out the steps makes debugging faster when a run fails.  In addition to immediately knowing if the failure is because of restore or build or test, the logs for the uniquely failing step are roughly a third as large as they would be when combined.  We only have to "write this CI config once", but we will be looking at run failures significantly more often.  I want to optimize for that case.

> * Is the nuget.config needed?

Not sure.  The restore steps seems inconsistent when `NuGet.org` doesn't exist.

I [removed that `NuGet.config` in this commit](https://github.com/TysonMN/Elmish.WPF/pull/4/commits/d8174951ad8857f20de2c5c6d9e5d6639ce6b714) and [the corresponding restore was successful](https://github.com/TysonMN/Elmish.WPF/runs/1949153445).  But [in this next commit that squashed that change and the addition of new lines](https://github.com/TysonMN/Elmish.WPF/pull/4/commits/e4f78a22e70ba9c35a6e6bb6c1265540925444fb), [the restore failed](https://github.com/TysonMN/Elmish.WPF/runs/1949181312).  My guess is that the second restore failed because the `NuGet.org` was removed in the previous commit but the restore for the first commit passed because of cached NuGet packages.

I tried [removing `NuGet.config` again](https://github.com/TysonMN/Elmish.WPF/pull/4/commits/2d05beb9f31e80b0287bb92a98b412900bebaffc).  This time, [the restore succeeded the first time](https://github.com/TysonMN/Elmish.WPF/runs/1949221887) and [the "second" time](https://github.com/TysonMN/Elmish.WPF/runs/1949249761).

Adding this `NuGet.config` solved a similar restore failure on a side project of mine, so it seemed obvious to me include it after the restore failed in the way it did.  I tried to find some mention of there no longer being a default NuGet package source.  The best I could find is [this issue about Visual Studio](https://developercommunity.visualstudio.com/content/problem/506006/visual-studio-2019-preview-does-not-have-a-nuget-p.html).

It would make me feel better if we kept `NuGet.config`.

> Could [nuget.config] be simplified?

(FWIW...)
Not sure.  I created this file by hand about 1.5 years ago when I started my current project.  At the time, I read through all the configuration options and decided that I wanted the file to look like that.  I have now forgotten what everything means.  I can read up on it again if you want me to.

As I take a brief look at it again, it seems one setting was duplicated.  I have removed that now.

> * Can you add an empty line between all steps (and above the first step)?
> * Can you add an empty line at the end of the files?

Done and done in a force pushed commit.

NuGet.config
```diff
@@ -0,0 +1,17 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <solution>
+    <add key="disableSourceControlIntegration" value="True" />
```
cmeeren, line 4: I don't understand this line. According to [the reference](https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file#solution-section), it indicates

> whether to ignore the packages folder when working with source control.

I don't understand this. NuGet has nothing to do with VC (Git in our case). We use a `.gitignore` for that. I have probably misunderstood this setting, but I don't understand why it would be needed.

NuGet.config
```diff
@@ -0,0 +1,17 @@
...
+  </solution>
+  <packageRestore>
+    <add key="enabled" value="True" />
+    <add key="automatic" value="True" />
+  </packageRestore>
```
cmeeren, line 9: I don't think these are needed, particularly not when building with `dotnet build`. Seems like legacy stuff. All my VS and CI stuff have always worked without these.

NuGet.config
```diff
@@ -0,0 +1,17 @@
...
+  <packageSources>
+    <add key="Official source v3" value="https://api.nuget.org/v3/index.json" />
+  </packageSources>
```
cmeeren, line 12: I have never needed this. The docs for [dotnet restore](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-restore#specify-feeds) says:

> To restore the dependencies, NuGet needs the feeds where the packages are located. Feeds are usually provided via the nuget.config configuration file. A default configuration file is provided when the .NET SDK is installed. To specify additional feeds, do one of the following:

So I don't think this is needed.

NuGet.config
```diff
@@ -0,0 +1,17 @@
...
+  <packageManagement>
+    <add key="format" value="0" />
+    <add key="disabled" value="False" />
+  </packageManagement>
```
cmeeren, line 16: These are legacy and not relevant for SDK-style projects.

cmeeren (commented):
> We only have to "write this CI config once", but we will be looking at run failures significantly more often.

I see your point, but have a different perspective. We may have to modify it down the line, which requires understanding it, and we may have to read it even more often when debugging of the build. In short, the best parts of the CI config, like with all code, are the parts you never have to write. Therefore, a more complicated CI config (more steps, and each step being more complicated, e.g. `--no-restore` and `--no-build`) has costs. This should be offset by some tangible benefits. I agree that having a smaller log for a failing `dotnet build` or `dotnet fail` is useful, but IMHO the difference is insignificant; the errors are usually at the end of the log and fairly clear anyway.

Still, this was more of a philosophical argument. In this particular case, I can live with separate restore/build/test steps.

> It would make me feel better if we kept `NuGet.config`.

I understand. I won't force the issue (particularly since this is your time on the table), but I'd prefer if we either tried without `nuget.config`, or tried getting to the bottom of the strange behavior. I have left some comments below.

TysonMN:
> the best parts of the CI config, like with all code, are the parts you never have to write.

I don't agree with that.  Such a believe leads to concluding that "implicit is better than explicit".  However, as clearly stated in the Python community, I agree with the statement that "[explicit is better than implicit](https://www.python.org/dev/peps/pep-0020/#the-zen-of-python:~:text=Explicit%20is%20better%20than%20implicit.)".

> In this particular case, I can live with separate restore/build/test steps.

Thanks.  I feel rather strongly that separating these steps is better.

> I'd prefer if we either tried without `nuget.config` [...]

Ok, let's try it.  What I actually want is consistent builds.  Maybe this file will help achieve that and maybe it won't.

I force pushed it commit without it.  And [restore was successful again](https://github.com/TysonMN/Elmish.WPF/pull/4/checks?check_run_id=1953255299) :)

cmeeren:
> However, as clearly stated in the Python community, I agree with the statement that "[explicit is better than implicit](https://www.python.org/dev/peps/pep-0020/#the-zen-of-python:~:text=Explicit%20is%20better%20than%20implicit.)".

I agree with that. (Or I'd have problems with F#!) But I also think that as long as tooling/languages/etc. have sane and stable defaults, it's useful to lean into that. For example, you'd never manually implement `IEquatable` or `IComparable` in the already built-in way for all your F# records just because "explicit is better than implicit". In any case, I think explicit/implicit is a red herring in this particular case; the goal here, as you say, is to get consistent builds. I'm fine with separate steps, so the point is moot.

Regarding the nuget.config: The only useful thing I have done with it for my own OSS projects without special requirements, is to make sure the list of packages is stable across environments that may have non-standard machine-wide package sources configured. [This is one example.](https://github.com/cmeeren/Feliz.MaterialUI/blob/master/NuGet.Config) It just ensures that there is only one source being used, and it's the default one. Note that this is irrelevant for CI builds; it is only useful for contributors who have other package sources set up on their machine. If you want to add that configuration here, that's fine, though I have never had problems myself without it.

cmeeren (approved):

TysonMN:
Another useful thing achieved via a `NuGet.config` file is specificing a local package source: i.e. a folder in the repository containing checked-in/commited NuGet packages.  [Mark Seemann has done this in the past](https://blog.ploeh.dk/2014/01/29/nuget-package-restore-considered-harmful/).

I prefer committing self-produced NuGet packages into a folder marked as a local package source instead of fighting Azure DevOps to get the security set up correctly.

TysonMN:
FYI: [the build triggered by the merge into `master`](https://github.com/elmish/Elmish.WPF/actions/runs/589724317) passed

TysonMN:
FYI: By pushing [this "release" commit](https://github.com/elmish/Elmish.WPF/commit/4bec32041339b3dcd904447a8b13f645e50351da), I triggered [another build](https://github.com/elmish/Elmish.WPF/runs/1953946233?check_suite_focus=true), which [published the NuGet package](https://github.com/elmish/Elmish.WPF/runs/1953946233?check_suite_focus=true#step:8:50) and [added the correct tag](https://github.com/elmish/Elmish.WPF/runs/1953946233?check_suite_focus=true#step:8:56).

TysonMN:
FYI: I deactivated the webhook to AppVeyor (instead of deleting in case you want any of the information it contains).

cmeeren:
> Another useful thing achieved via a `NuGet.config` file is specificing a local package source: i.e. a folder in the repository containing checked-in/commited NuGet packages. [Mark Seemann has done this in the past](https://blog.ploeh.dk/2014/01/29/nuget-package-restore-considered-harmful/).

Indeed. I am doing just that in Facil for [reasons described in this section](https://github.com/cmeeren/Facil/blob/master/.github/CONTRIBUTING.md#development), since the nupkg contains build tools that are part of the tests.

> FYI: By pushing [this "release" commit](https://github.com/elmish/Elmish.WPF/commit/4bec32041339b3dcd904447a8b13f645e50351da), I triggered [another build](https://github.com/elmish/Elmish.WPF/runs/1953946233?check_suite_focus=true), which [published the NuGet package](https://github.com/elmish/Elmish.WPF/runs/1953946233?check_suite_focus=true#step:8:50) and [added the correct tag](https://github.com/elmish/Elmish.WPF/runs/1953946233?check_suite_focus=true#step:8:56).

Happy to hear it works!

> FYI: I deactivated the webhook to AppVeyor (instead of deleting in case you want any of the information it contains).

I don't think I need that. Feel free to delete it if you want.

TysonMN:
> > FYI: I deactivated the webhook to AppVeyor (instead of deleting in case you want any of the information it contains).
> 
> I don't think I need that. Feel free to delete it if you want.

Deleted

PR #349, opened 2021-02-19
Simplify FileDialog save/load

TysonMN:
I don't think the `save` or `load` functions need to be executed on the main thread.  This can be tested by adding the line
```fsharp
do! Async.SwitchToThreadPool ()
```
to the beginning of these CEs on this branch.  Everything still works.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,7 @@
 * Changed CurrentModel and UpdateModel on ViewModel<_,_> from public to internal
 * Added support for composable binding validation
 * `runWindow` now shows the given window after settings its `DataContext`.  This removes the need to have `Visibility` values default to `Collapsed`.
+* Changed minimum Elmish version from 3.0.3 to 3.1.0 (which is currently the latest).  Commands created with `OfAsync` are now executed on a threadpool thread.  For example, it is now easier to show file dialogs without blocking the Elmish dispatch loop.  See [this diff](https://github.com/elmish/Elmish.WPF/commit/d1ec823ccd7f377a860b76bc2358706dc6a70c84).
 
 #### 4.0.0-beta-2
 * Added logging when INotifyDataErrorInfo.HasErrors is called
--- a/src/Samples/FileDialogs.Core/Program.fs
+++ b/src/Samples/FileDialogs.Core/Program.fs
@@ -1,9 +1,7 @@
-﻿module Elmish.WPF.Samples.FileDialogs.Program
+module Elmish.WPF.Samples.FileDialogs.Program
 
 open System
 open System.IO
-open System.Threading
-open System.Windows
 open Serilog
 open Serilog.Extensions.Logging
 open Elmish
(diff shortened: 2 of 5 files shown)


cmeeren:
Does this impact [this part of the readme](https://github.com/elmish/Elmish.WPF#how-can-i-use-save-file--open-file-dialogs)?

TysonMN:
Oh, I was slightly confused.  I was confusing blocking the Elmish dispatch loop (or not) with must be executed on the main thread (or not).  (See how I named this PR?)

The [first commit](https://github.com/elmish/Elmish.WPF/pull/349/commits/61c039a8194a059e0abd2f5f6f16cc0675283b95) _does_ block the Elmish dispatch loop.  The [second commit](https://github.com/elmish/Elmish.WPF/pull/349/commits/61ec1fa2bdf07529c2c21773523a41c4cd195232) (that I just pushed) adds
```fsharp
do! Async.SwitchToThreadPool ()
```
which recovers the original behavior that doesn't block the Elmshing dispatch loop.

So this PR is just to simply the code used to avoid blocking the Elmish dispatch loop.

And actually, there is a slightly simpler way to do this.  One more commit coming soon.

TysonMN:
Now this [third commit](https://github.com/elmish/Elmish.WPF/pull/349/commits/cc358d73f548b36a8ef5f0afc31b9da1f8898ade) that I just pushed simplifies the code further.  It uses version 3.1.0 of Emlish instead of version 3.0.3 that is required by Elmish.WPF.  [Version 3.0.5 of Elmish changed the `ofAsync` functions to execute on the thread pool](https://github.com/elmish/elmish/commit/18fdbd926d26d92776c8b8dcb2315eadb4d65cbc#diff-68d24fa81558aae3d8c59e2aa57a4fa719ea3b04d7fa14beff45f16f00858f50R3).
https://github.com/elmish/Elmish.WPF/blob/db1ee9cff774cf3290caff2da6bac216c1f01391/src/Elmish.WPF/Elmish.WPF.fsproj#L47

> Does this impact [this part of the readme](https://github.com/elmish/Elmish.WPF#how-can-i-use-save-file--open-file-dialogs)?

Definitely.

What do you think of the approach now in this branch of executing commands in the thread pool by using a sufficiently new version of Elmish?

Maybe the easiest thing to do is change Elmish.WPF to depend on the latest version of Elmish (or at least version 3.0.5).  Then we might not even need to mention this potential issue about blocking the Elmish dispatch loop.

cmeeren:
I think we should require a newer Elmish, get rid of the section in the readme, and update this PR accordingly (at least remove [this line](https://github.com/elmish/Elmish.WPF/pull/349/files#diff-938113000b9d86d84668040869b9f0d049c4e40328ea499b6c9366504d466027R14)).

I have no strong opinions on which Elmish version we should require (3.0.5 vs. 3.1.0). Elmish.WPF is a framework used in apps, not a library referenced by other libraries, so we can require pretty much whatever we want. Generally I see no reason to not require the latest version. It's often safer, in the sense that even minor updates to libraries such as Elmish may potentially contain binary breaking changes that cause stuff to fail at runtime if Elmish.WPF is compiled for a lower version than the user uses.

But first, could you verify that this PR works as expected (doesn't crash / require the UI thread, doesn't block any loops) on all our supported target frameworks, i.e. .NET Core 3 and .NET Framework in addition to .NET 5?

TysonMN:
Yes, I agree with depending on the lastest Elmish release.

I force pushed the branch.  Just two commits now.  No crashes or loop blocking for all three targets.

I think this is ready to merge.

TysonMN:
(Force pushed again to add a line to the release notes)

TysonMN:
*sigh*

Another force push to made the same simplification to the `FileDialogsCmdMsg` sample.

RELEASE_NOTES.md
```diff
@@ -3,6 +3,7 @@
 * Improved logging
 * Changed CurrentModel and UpdateModel on ViewModel<_,_> from public to internal
 * Added support for composable binding validation
+* Chagned minimum Elmish version from 3.0.3 to 3.1.0 (which is currently the latest)
```
cmeeren, line 6: Typo. Also, this point should perhaps mention something to the effect that juggling around synchronization context for the relevant dialogs are no longer needed.

src/Samples/FileDialogs.Core/Program.fs
```diff
@@ -1,9 +1,7 @@
-﻿module Elmish.WPF.Samples.FileDialogs.Program
+module Elmish.WPF.Samples.FileDialogs.Program
```
cmeeren, line 1: I am very confused about why GH sometimes show these diffs on the first line. Do you know why? I would guess the file has been saved for the first time in UTF-8 with a BOM (Byte Order Mark) at the start; could that be it?

You don't have to revert it, I'm just curious.

RELEASE_NOTES.md
TysonMN: Just force pushed a rebase onto `master` that also fixes that typo and mentions the execution happening on the threadpool.

src/Samples/FileDialogs.Core/Program.fs
TysonMN: I had to look at a HEX editor to see the difference.  It sounds like your guess is correct.  I don't know what those bytes mean though.

## File in `master`
2021-02-19_22-20-54_524_chrome

## File in PR's branch
2021-02-19_22-21-00_525_chrome

cmeeren: https://en.wikipedia.org/wiki/Byte_order_mark

RELEASE_NOTES.md
cmeeren: Thanks. That is useful. I'm thinking however that we should also explicitly mention that code related to save/load dialogs can be simplified. Otherwise people will have no idea they can simplify this unless they diff the readme and see that the FAQ entry is removed. We don't have to provide a lot of detail though; we can link to the `Program.fs` diff in the FileDialogs sample. Something like this:

> This means that e.g. Save File / Open File dialogs can be simplified. See [this diff for details](https://github.com/elmish/Elmish.WPF/pull/349/files#diff-48689f4b515bf41b1bdad0aee209185e63ee109980cff7c8a4bd2996bc318aa2).

But replace the link with the actual commit diff after merging. (Not idea if this can be done before merging.)

TysonMN: Sure.  This can be done by splitting the commits, ordering the commits, pushing the first two, and then including the link to the second commit diff in the third commit.  I just force pushed those commits.

cmeeren: Yes, but now the link points to the commit in the context of the PR. After merging, I suggest updating the link to a "raw" diff view of the commit, to reduce noise. As far as this PR goes, though, this is fine.

cmeeren (approved):
Looks good. Thanks a lot! 👍

RELEASE_NOTES.md
TysonMN: Oh, ok.  I didn't think carefully about that link.  It contained the correct git commit hash but not the right GitHub URL.  I replaced it with the correct URL, which won't work until I complete this PR...because the branch in this PR is in my fork instead of in this repo.

TysonMN: I merged the PR.  [This is the link in the release notes](https://github.com/elmish/Elmish.WPF/commit/d1ec823ccd7f377a860b76bc2358706dc6a70c84), which is working now.

PR #350, opened 2021-02-19
Set DataContext before showing window

TysonMN:
We are showing the window too early.  We should only show the window after we assign our view model to its `DataContext`.

Here is how this can go wrong.  Suppose the binding initialization code takes a long time.  In the following example, I added a two-second delay to the `SubModelOpt` sample.  I also removed the `Visibility` fallbacks to `Collapsed`.  I also changed to entry point to be an explicit `Main` method.  Then this is what happens: both forms are visible during that two-second loading time.

After making the change in this PR, which is to start the Elmish dispatch loop first and then show the window, the behavior is good.  Specifically, the window does not open during that two second delay.  When it does open, no forms are visible, which is the correct behavior.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,7 @@
 * Improved logging
 * Changed CurrentModel and UpdateModel on ViewModel<_,_> from public to internal
 * Added support for composable binding validation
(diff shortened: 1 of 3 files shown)


cmeeren (approved):
Looks good!

Issue #330, opened 2021-02-03
TaskCanceledException on app close when onCloseRequest of Binding.subModelWin is used


BentTranberg:
Reproduce:

1. Modify this binding in sample NewWindow. Add CloseWin1 or HideWin1 at end, like this.

```
    "Win1" |> Binding.subModelWin(
      (fun m -> m.Win1State), fst, id,
      window1Bindings,
      createWindow1, CloseWin1)

```

2. Run program.
3. Show Window 1
4. Close main window.
5. VS debugger will halt on a TaskCanceledException.

If Window 1 is never shown, the exception will not occur.
If the parameter is not used, the exception will never occur.
If the parameter is used, Window 1 is opened, and Window 1 is closed in any way, then the exception will not occur.

So it looks as if the problem is with having Elmish.WPF close Window 1 during shutdown. (edit: It's late. My analysis may not be 100%, but it's close enough for now.)

A recent change to the sample, resulting from issue #328, is likely relevant here.

BentTranberg:
Is the TaskCanceledException actually a bug, or is it just a kind of "FYI"? I would say it's annoying in any case, because it stops debugging as default in VS.

Is a possible solution simply not to do the magic of that extra parameter if the window and/or application is in some sort of shutdown phase? I mean, everything is fine without it.

TysonMN:
I was able to reproduce.  Seems to happen every time.  (Though, it might not have happened the first time I tried...or I did something wrong).

I agree with you that I behavior is fundamentally correct but not as optimized as it could be.  This is both an issue for developers when running from within Visual Studio (as you pointed out) as well as a pollution of the `FirstChanceException` event.

> Is a possible solution simply not to do the magic of that extra parameter if the window and/or application is in some sort of shutdown phase?

Seems like it should be possible, but I can't find a solution.

- The [`Window.Closing`](https://docs.microsoft.com/en-us/dotnet/api/system.windows.window.closing) on supplies the [`CancelEventArgs`](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.canceleventargs), which only allows you to cancel the closing ([unless the application is shutting down](https://docs.microsoft.com/en-us/dotnet/api/system.windows.window.closing?view=net-5.0#remarks:~:text=If%20Shutdown%20is%20called%2C%20the%20Closing,Closing%20is%20canceled%2C%20cancellation%20is%20ignored.)).
[...]

PR #355, opened 2021-03-01
stop subscriptions on window close, which is before dispatcher is disposed

### LINKED ISSUE #353: TaskCanceledException on program end in timerTick in FileDialogsCmdMsg.Core

image

TysonMN:
Fixes #353

The only changes are in the samples.  The new code stops the infinite loops inside the subscriptions that would otherwise dispatch a message to a disposed `Dispatcher`.

Issue #353 becomes perfectly repeatable for me (in my experience) after decreasing the sleep time `1000` to `10`.  Do that as well as comment out the subscription to `Window.OnClosing` to confirm that the issue also becomes perfectly repeatable for you.  Then comment that line back in to verify that this fix works.

@BentTranberg, I think this fixes #353.  Do you agree?

### DIFF
--- a/src/Samples/FileDialogs.Core/Program.fs
+++ b/src/Samples/FileDialogs.Core/Program.fs
@@ -1,6 +1,9 @@
 module Elmish.WPF.Samples.FileDialogs.Program
 
 open System
+open System.Threading
+open System.Threading.Tasks
+open System.Windows
 open System.IO
 open Serilog
 open Serilog.Extensions.Logging
@@ -84,13 +87,7 @@ let bindings () : Binding<Model, Msg> list = [
 let designVm = ViewModel.designInstance (init () |> fst) (bindings ())
 
 
-let timerTick dispatch =
-  let timer = new Timers.Timer(1000.)
-  timer.Elapsed.Add (fun _ -> dispatch (SetTime DateTimeOffset.Now))
-  timer.Start()
-
-
-let main window =
+let main (window: Window) =
 
   let logger =
     LoggerConfiguration()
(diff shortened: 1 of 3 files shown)


cmeeren:
Is this for merging, or just for confirming that #353 can be avoided in user code?

If it's for merging:
* Is it possible to fix this in Elmish.WPF code instead of user code?
* If not, I have some nitpicks here, but I won't spend time on that until I know this PR should be merged.

BentTranberg:
Wow, there's lots of lessons for me in that code! It's the kind of magic that is beyond me unless the wizards care to share. Thanks a lot. Hasn't occurred to me that some "code behind" can be done on the F# side.

It is quite a mouthful for what seems to be such a simple issue, so if "possible to fix this in Elmish.WPF code instead" is potentially a much simpler solution in user code, that's certainly very interesting.

TysonMN:
This is for merging.  I would also prefer a fix in Elmish.WPF instead.  However, as @BentTranberg pointed out in https://github.com/elmish/Elmish.WPF/issues/353#issue-818153279, a fix for Elmish.WPF for #353 would probably also be a fix for #330, but as I said in https://github.com/elmish/Elmish.WPF/issues/330#issuecomment-773346512, I don't have any more ideas for what a fix in Elmish.WPF could be.

TysonMN:
> Wow, there's lots of lessons for me in that code! It's the kind of magic that is beyond me unless the wizards care to share.

Indeed.  There are a lot of ideas coming together here.

1. We want to stop the (otherwise) infinite loop of dispatches, so I switched from `Timer` to an `Async<_>` because...
2. Stopping some flow of control on other thread should be done with a `CancellationToken`, which can be obtained from a `CancellationTokenSource`.  In particular, a nice property of `CancellationTokenSource.Cancel` is that it is [idempotent](https://en.wikipedia.org/wiki/Idempotence), which means that calling it twice is equivalent to calling it once.  It makes it easier to reason about impure code.  We all know that the first "rule" of functional programming is to maximize the amount of pure code.  The second "rule" could be, among the impure code, to maximize the amount of idempotent code.  (Blog post in the queue on that topic.)
3. Although the `while` loop looks infinite, the actual behavior is that the `CancellationToken` is checked as part of the predicate.  I discovered this behavior in the process of writing [this blog post](https://tysonwilliams.coding.blog/2020-12-01_csharp_task_to_fsharp_async#loop).
4. It is not enough to cancel the asynchronous computation.  We also have to wait for the asynchronous computation to stop executing.  Therefore, cancelling the `CancellationToken` is followed by waiting for the `Task<_>` that comes from a `TaskCancellationSource<_>`, which is set by the asynchronous computation when it is cancelled.  It suffices to use `SetResult` there instead of `TrySetResult`, but I prefer `TrySetResult` because it is idempotent while `SetResult` is not (a second call to `SetResult` causes an exception to be thrown).

Did I miss anything?  Is there any other parts that are unclear?

BentTranberg:
> magic that is beyond me

I realize now that it was an ambiguous statement - true in any case - and thanks a lot again. My original intention was meant to indicate I wouldn't have come up with that kind of solution, though CancellationToken was on my mind.  I would not have thought of async, and I would have messed around in the C# project too. Now that I see it, I understand it, but your explanation is on a higher level - not something I could have expressed.

cmeeren:
Sorry for taking so long. This (specifically, #353) is fixable in Elmish.WPF. See #358.

Issue #300, opened 2020-11-22
Attempting to emulate C# CollectionViewSource among three views of the same list.


awaynemd:
How to emulate the C# CollectionView functionality.

I have attempted (unsuccesfully) to do the following:

1. download a list of names (i.e., ViewName) from a WCF service. This works! (Although I do not know if my use of Async is correct not to block the UI). But for now, this does work correctly to download about 20,000 names.

2.  Create only ONE read of the database.

3.  The read database names are then to be provided to three separate user controls. Each user control is to receive user input from a textbox and will be sorted in a different order. Each user control will then provide this user input into a shared filter. Shared in the sense that all three usercontrols will use the same filter on the same list of objects.  The filters are:
           1. All ViewNames with LastName starting with the text characters received from the LastName usercontrol.
           2. All ViewNames with FirstName starting with the text character received from the FirstName usercontrol.
           3. All ViewNames with BirthDate equaling the date typed into the BirthDate usercontrol.
The final filter which ALL LIST VIEWS will use in the boolean combination of the three above conditions.

(The sorted viewables will be:
1. For the LastName, sort is by LastName, FirstName, and Birthdate (all ascending),
2. For the FirstName, sort is by FirstName, LastName, and Birthdate (all ascedning),
3. For the BirthDate, sort is by BirthDate, LastName, FirstName  (all ascending).

Everything initially worked individually, but I was unable to use the filter from one usercontrol in another.  That is, my thinking is to have one source list/array and somehow pass the conditions of each usercontrol to the other usercontrols --or top module. Hence, althought the sort orders for the list will be different, the source list and the filter will be identical between all user controls.

I tried passing a common record, "names", with the Msg component--but failed miserably.

If anybody could give any thought to this, this newbie would sure appreciate it!

(As an aside, I never could figure out how to verify a text string as a valid date in F# to be used in a birth date filter :( )

TIA

Full (frustrating) code found at: https://github.com/awaynemd/PatientFinder

TysonMN:
It might be two weeks before I have time to look into this.

> I never could figure out how to verify a text string as a valid date in F#...

Do you know how to do this in C#?

awaynemd:
@TysonMN Yes. 

I'm updating the repo with most of it working. Each usercontrol correctly works to display its specific filter to the xaml's.  

The obvious errors I see are:
1.  Each submodel has its own copy of the original Suggestion list -- no changes by filtering are mirrored to each submodel.
[...]

PR #358, opened 2021-03-05
Dispatch only if dispatcher is active

### LINKED ISSUE #330: TaskCanceledException on app close when onCloseRequest of Binding.subModelWin is used

### LINKED ISSUE #353: TaskCanceledException on program end in timerTick in FileDialogsCmdMsg.Core

cmeeren:
Closes #355

Fixes #353 
Fixes #330 (as of https://github.com/elmish/Elmish.WPF/pull/358/commits/a0f75c1ef9bc300f5a98e2c2a2a00b0018490600)

From https://github.com/elmish/Elmish.WPF/pull/355#issuecomment-787956342:

> I would also prefer a fix in Elmish.WPF instead. However, as @BentTranberg pointed out in [#353 (comment)](https://github.com/elmish/Elmish.WPF/issues/353#issue-818153279), a fix in Elmish.WPF for #353 would probably also be a fix for #330, but as I said in [#330 (comment)](https://github.com/elmish/Elmish.WPF/issues/330#issuecomment-773346512), I don't have any more ideas for what a fix in Elmish.WPF could be.

I may be missing something, but I just dotted into `Dispatcher` and saw it had an `HasShutdownStarted` property, so I figured I could try guarding `Invoke` behind a check to that property. Seems to work like a charm.

The changes in `ViewModel.fs` are not needed to address #330 and #353, but for good measure I added this everywhere we call `Dispatch`.

~Haven't looked into why this doesn't fix #330. Feel free to investigate.~

Waiting for review and/or further discussion before merging.

### DIFF
--- a/src/Elmish.WPF/AutoOpen.fs
+++ b/src/Elmish.WPF/AutoOpen.fs
@@ -3,6 +3,7 @@ module internal AutoOpen
 
 open System
 open System.Collections.Generic
+open System.Threading.Tasks
 open System.Windows.Threading
 
 
@@ -17,5 +18,6 @@ let (|Kvp|) (kvp: KeyValuePair<_,_>) =
 
 type Dispatcher with
   member this.InvokeIfActive(callback: Action) =
-    if not this.HasShutdownStarted then
+    try
       this.Invoke(callback)
+    with :? TaskCanceledException -> ()
--- a/src/Elmish.WPF/AutoOpen.fs
+++ b/src/Elmish.WPF/AutoOpen.fs
@@ -1,7 +1,9 @@
 ﻿[<AutoOpen>]
 module internal AutoOpen
 
+open System
 open System.Collections.Generic
+open System.Windows.Threading
 
 
 let flip f b a = f a b
@@ -11,3 +13,9 @@ let ignore2 _ _ = ()
 /// Deconstructs a KeyValuePair into a tuple.
 let (|Kvp|) (kvp: KeyValuePair<_,_>) =
   Kvp (kvp.Key, kvp.Value)
+
+
+type Dispatcher with
+  member this.InvokeIfActive(callback: Action) =
+    if not this.HasShutdownStarted then
+      this.Invoke(callback)
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -154,7 +154,7 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
       initialVisibility =
     let win = getWindow currentModel dispatch
     winRef.SetTarget win
-    win.Dispatcher.Invoke(fun () ->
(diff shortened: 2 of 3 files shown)


cmeeren:
This now also fixes #330. `DispatchIfActive` now calls `Dispatch` and just catches `TaskCanceledException` to avoid a race condition where `HasShutdownStarted` changes between checking and invoking.

I can't immediately think of any good reasons not to do this, but let me know if this is a bad idea.

TysonMN:
Oh, we were working on fixes at the same time!  I just created PR #359 to fix #330.

I don't have time to look at your PR now, but I am eager to look at it later!

TysonMN:
> I may be missing something, but I just dotted into `Dispatcher` and saw it had an `HasShutdownStarted` property, so I figured I could try guarding `Invoke` behind a check to that property. Seems to work like a charm.
> [...]
> I changed `DispatchIfActive` to always call `Dispatch` and just catch `TaskCanceledException`, instead of first checking `HasShutdownStarted`. This avoids a race condition where `HasShutdownStarted` changes between checking and invoking.

I also considered guarding with `HasShutdownStarted` (as is listed in https://github.com/elmish/Elmish.WPF/issues/330#issuecomment-773346512) but reached the same conclusion as you.

> I can't immediately think of any good reasons not to do this, but let me know if this is a bad idea.

I don't like it for two reasons.
1. It pollutes the `FirstChanceException` event.  I prefer the look-before-you-leap style over the ask-for-forgiveness-not permission style.  I prefer for a thrown exception to mean that the code contains a bug.
2. A thrown exception might cause the debugger to break (depending on the exception settings), which interrupts whatever the actual goal is.  This can happen when people are exploring the samples or we test them but not when people depend on Elmish.WPF as a NuGet package.

I really like the fix I found for #300 that is in PR #359.  Maybe the idea there of calling `BeginInvoke` can also be used to fix #353.

cmeeren:
> 1. I prefer the look-before-you-leap style over the ask-for-forgiveness-not permission style

I agree, as long as there is a robust alternative. `HasShutdownStarted` isn't robust enough. Eager to see what you find out regarding `BeginInvoke`.

PR #359, opened 2021-03-05
Close window and dispatch messages via Dispatcher.InovkeAsync

### LINKED ISSUE #330: TaskCanceledException on app close when onCloseRequest of Binding.subModelWin is used

### LINKED ISSUE #353: TaskCanceledException on program end in timerTick in FileDialogsCmdMsg.Core

TysonMN:
Fixes #330
Fixes #353
Closes #355
Closes #358

Calling `Window.Close` is not allowed when the `Window` in question is in the process of closing.  In case that is the situation, we can avoid this below slightly delaying our call to `Window.Close` by using `Dispatcher.BeginInvoke`.

### DIFF
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -6,6 +6,7 @@ open System.Collections.Generic
 open System.Collections.ObjectModel
 open System.ComponentModel
 open System.Windows
+open System.Windows.Threading
 open Microsoft.Extensions.Logging
 
 open Elmish
@@ -368,7 +369,13 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
             | true, w ->
                 log.LogTrace("[{BindingNameChain}] Closing window", winPropChain)
                 b.WinRef.SetTarget null
-                w.Dispatcher.Invoke(fun () -> w.Close ())
+                (*
+                 * The Window might be in the process of closing,
+                 * so instead of immediately exeucting Window.Close via Dispatcher.Invoke,
+                 * queue a call to Window.Close via Dispatcher.BeginInvoke.
+                 * https://github.com/elmish/Elmish.WPF/issues/330
+                 *)
+                w.Dispatcher.BeginInvoke(DispatcherPriority.Normal, Action(w.Close)) |> ignore
             b.WinRef.SetTarget null
 
           let hide () =
--- a/src/Elmish.WPF/ViewModel.fs
+++ b/src/Elmish.WPF/ViewModel.fs
@@ -6,7 +6,6 @@ open System.Collections.Generic
 open System.Collections.ObjectModel
 open System.ComponentModel
 open System.Windows
-open System.Windows.Threading
 open Microsoft.Extensions.Logging
 
 open Elmish
@@ -372,10 +371,10 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
                 (*
                  * The Window might be in the process of closing,
                  * so instead of immediately exeucting Window.Close via Dispatcher.Invoke,
-                 * queue a call to Window.Close via Dispatcher.BeginInvoke.
+                 * queue a call to Window.Close via Dispatcher.InvokeAsync.
                  * https://github.com/elmish/Elmish.WPF/issues/330
                  *)
-                w.Dispatcher.BeginInvoke(DispatcherPriority.Normal, Action(w.Close)) |> ignore
+                w.Dispatcher.InvokeAsync(w.Close) |> ignore
             b.WinRef.SetTarget null
 
           let hide () =
--- a/src/Elmish.WPF/WpfProgram.fs
+++ b/src/Elmish.WPF/WpfProgram.fs
@@ -1,8 +1,6 @@
 ﻿namespace Elmish.WPF
 
-open System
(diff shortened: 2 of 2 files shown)


TysonMN:
Quoting https://github.com/elmish/Elmish.WPF/pull/358#issuecomment-791873420
> I really like the fix I found for #300 that is in PR #359. Maybe the idea there of calling `BeginInvoke` can also be used to fix #353. I will look into that this weekend.

Indeed.  Using `Dispatcher.BeginInvoke` also fixes #353.  I added a second commit with that fix.

Unlike @cmeeren's approach in PR #358, I decided not to apply this change to all calls to `Dispatcher.Invoke`.

src/Elmish.WPF/ViewModel.fs
```diff
@@ -368,7 +369,13 @@ and [<AllowNullLiteral>] internal ViewModel<'model, 'msg>
...
                 log.LogTrace("[{BindingNameChain}] Closing window", winPropChain)
                 b.WinRef.SetTarget null
-                w.Dispatcher.Invoke(fun () -> w.Close ())
+                (*
+                 * The Window might be in the process of closing,
+                 * so instead of immediately exeucting Window.Close via Dispatcher.Invoke,
+                 * queue a call to Window.Close via Dispatcher.BeginInvoke.
+                 * https://github.com/elmish/Elmish.WPF/issues/330
+                 *)
+                w.Dispatcher.BeginInvoke(DispatcherPriority.Normal, Action(w.Close)) |> ignore
```
cmeeren, line 378: If you want, I think you can remove the parentheses around the `Action` argument. I usually remove parentheses if there's only a single argument, but do what you want.

```suggestion
                w.Dispatcher.BeginInvoke(DispatcherPriority.Normal, Action w.Close) |> ignore
```

src/Elmish.WPF/WpfProgram.fs
```diff
@@ -67,7 +69,9 @@ module WpfProgram =
           vm.UpdateModel model
 
     let uiDispatch (innerDispatch: Dispatch<'msg>) : Dispatch<'msg> =
-      fun msg -> element.Dispatcher.Invoke(fun () -> innerDispatch msg)
+      fun msg ->
+        let f () = innerDispatch msg
+        element.Dispatcher.BeginInvoke(DispatcherPriority.Normal, Action(f)) |> ignore
```
cmeeren, line 74: See previous comment.

```suggestion
        element.Dispatcher.BeginInvoke(DispatcherPriority.Normal, Action f) |> ignore
```

cmeeren (commented):
Couple of questions:
* Have you considered `InvokeAsync` instead of `BeginInvoke`? (I have no idea if that would work or what would be best; just a thought.)
* Any reason not to do this everywhere we currently call `Invoke`? Are you sufficiently certain that these places won't blow up in the same manner under certain (possibly nondeterministic and hard-to-guess/debug) circumstances?

TysonMN:
> * Have you considered `InvokeAsync` instead of `BeginInvoke`? (I have no idea if that would work or what would be best; just a thought.)

Great suggestion.

The documentation for [`BeginInvoke`](https://docs.microsoft.com/en-us/dotnet/api/system.windows.threading.dispatcher.begininvoke) and [`InvokeAsync`](https://docs.microsoft.com/en-us/dotnet/api/system.windows.threading.dispatcher.invokeasync) is nearly the same.  I like `InvokeAsync` better because I think the name is more idiomatic with the current approach to asynchronous programming (even though this method doesn't return `Task<_>` or `Async<_>`) and because it has overloads that are easier to call.

And of course I tested that using `InvokeAsync` still fixes both #330 and #353.

I just pushed a commit that changes both calls to use `InvokeAsync`.

> * Any reason not to do this everywhere we currently call `Invoke`? Are you sufficiently certain that these places won't blow up in the same manner under certain (possibly nondeterministic and hard-to-guess/debug) circumstances?

Actually, I think the remaining three calls to `Invoke` are not needed.  In each case, I think the executing thread is already the main thread.  I would prefer to merge this PR and separately consider the other uses of `Invoke`.  I will make a new issue about it.

src/Elmish.WPF/ViewModel.fs
TysonMN: No longer relevant

src/Elmish.WPF/WpfProgram.fs
TysonMN: No longer relevant

cmeeren:
Thanks!

PR #365, opened 2021-03-08
removed UseWpf in samples

TysonMN:
Replaces #282

This PR removes the line
```xml
<UseWpf>true</UseWpf>
```
from all of `Core` sample projects.

Removing this line from these projects was not possible when PR #282 was created because we were still targeting .NET Core 3 (`netcoreapp3.0` or `netcoreapp3.1`).  Now we are targeting .NET 5 (`net5.0-windows`).  When this change was made in 77635638db374328b1d091d75ea38ca2d39e5989, the SDK was simplified.  I have now realized that we could have also removed this line as well.

@cmeeren, since you previously said in https://github.com/elmish/Elmish.WPF/pull/282#issuecomment-705178872 that you were indifferent to this kind of change, I will merge when the build successfully completes.


TysonMN:
(That commit reference accidently went to the parent of the commit I wanted.  I corrected it now.)

cmeeren:
Thanks!

PR #368, opened 2021-03-09
removed async CE in window closing event

TysonMN:
This PR deletes three lines and changes the indentation of one line, so it helps to view the diff while ignoring whitespace changes.

I think the `async` computational expression is not needed in the callback given to the `Window.Closing` event.  It might have been needed before.  Now though, invoking `dispatch` eventually leads to this asynchronous `Dispatcher` execution.
https://github.com/elmish/Elmish.WPF/blob/cdf421f9ec7c3ec4d6d3e3923cd20786004ed098/src/Elmish.WPF/WpfProgram.fs#L70


cmeeren (approved):
As long as everything still works, this is fine 👍

PR #369, opened 2021-03-09
removed workaround for DataContext bug in .NET 3

TysonMN:
The `DataContext` binding bug at design-time seems to be fixed now.  This PR removes the workaround we were using.  I verified that the binding is working correctly at design-time in all of those cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,25 +258,6 @@ Then use the following attributes wherever you need a design-time VM:
 
 When targeting legacy .NET Framework, “Project code” must be enabled in the XAML designer for this to work.
 
-##### .NET Core 3 workaround
-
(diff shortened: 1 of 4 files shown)


cmeeren:
Could you update the readme, too?

TysonMN:
I considered that, but I think the README could remain as is.  It mentions the workaround for .NET Core 3.  Do you want to remove this information?

TysonMN:
Oh, the bug is fixed even when targeting .NET Core 3.  I will remove the workaround from the README.

TysonMN:
Things are more complicated that I thought.  Although I verified that the design-time bindings are working in our samples and in [my reproduction of that bug](https://developercommunity.visualstudio.com/t/design-time-data-in-datacontext-binding-not-displa/1133390), the design-time bindings are not working in [someone else's reproduction of that bug](https://developercommunity.visualstudio.com/t/wpf-datacontext/1112239).

PR #398, opened 2021-07-16
Proofreading

the-not-mad-psychologist:


TysonMN:
Thanks for the contribution @the-not-mad-psychologist :)

PR #405, opened 2021-08-11
Corrected SubModelSeq namespace in XAML

the-not-mad-psychologist:


TysonMN:
Same problem with GitHub here.  Oh well.

Thanks again! :)

Issue #188, opened 2019-11-23
Can't get multiselect ListBox to work


BentTranberg:
I am trying to implement use of ListBox with SelectionMode="Extended". I have implemented something I would have thought should work, but there are obvious problems with it. Perhaps somebody can have a look. It's a very simple example.

The repo is here: https://github.com/BentTranberg/ExploreElmishWpf

After start, this particular example is in the **Multiselect** pane.

Some of the items in the listbox should be "checked" after start, but they're all "not checked". Furthermore, when clicking on an item, this exception is raised : `System.InvalidOperationException: Collection was modified; enumeration operation may not execute.`

The source is so short I think I'll post a compressed version here.

What I am trying to achieve is simply to bind ListBoxItem.IsSelected to Marked in the submodel.

```
        <ListBox ItemsSource="{Binding Lines}" SelectionMode="Extended">
            <ListBox.ItemTemplate>
                <DataTemplate>
                    <Label Content="{Binding DisplayText}"/>
                </DataTemplate>
            </ListBox.ItemTemplate>
            <ListBox.ItemContainerStyle>
                <Style TargetType="ListBoxItem">
                    <Setter Property="IsSelected" Value="{Binding Marked, Mode=OneWayToSource}"/>
                </Style>
            </ListBox.ItemContainerStyle>
        </ListBox>
```

```
    type Line = { Id: Guid; Marked: bool; DisplayText: string }

    type Model = { Lines: Line list }

    type Msg =
        | SetMarked of Guid * bool

    let init () =
        {
            Lines = [
                { Id = Guid.NewGuid(); Marked = false; DisplayText = "Zero" }
                { Id = Guid.NewGuid(); Marked = true; DisplayText = "One" }
                { Id = Guid.NewGuid(); Marked = false; DisplayText = "Two" }
                { Id = Guid.NewGuid(); Marked = true; DisplayText = "Three" }
            ]
        }, Cmd.none

    let update msg m =
        match msg with
        | SetMarked (id, marked) ->
            let lines = m.Lines |> List.map (fun line ->
                if line.Id = id then { line with Marked = marked } else line)
            { m with Lines = lines }, Cmd.none

    let bindings () : Binding<Model, Msg> list = [
        "Lines" |> Binding.subModelSeq((fun m -> m.Lines), (fun line -> line), fun () -> [
                "Id" |> Binding.oneWay (fun (_, line) -> line.Id)
[...]

PR #409, opened 2021-08-14
Add working example for multi-selection

Evangelink:
Relates to #188 

I wanted to first see what code was looking like when doing multi-selection in a `ListBox` to try and implement the suggested `Binding.subModelSelectedItems` and realized that the property `SelectedItems` is readonly and that the recommended ways to implement the selection are either to bind the `IsSelected` property of the `ListBoxItem` or to use the `SelectionChanged` event (not documented in the current PR but happy to edit or submit another PR) so I am not sure how you were considering to use this `subModelSelectedItems`.

I have made some tests and it seems that switching over to `SelectionMode = Single` correctly updates the model to contain only 1 selected item. Obviously, as described in the linked ticket, the following approach is limited when we want to ensure only one item is selected because the model represents the multi-selection.

The current example is with a `ListBox` but I could provide examples with `ComboBox` and maybe `ListView`.


src/Elmish.WPF.sln
```diff
@@ -68,9 +68,13 @@ Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "Sticky.Core", "Samples\Stic
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Sticky", "Samples\Sticky\Sticky.csproj", "{668E3AAD-B2AA-4934-804E-234C4F502038}"
 EndProject
-Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Capabilities.Core", "Samples\Capabilities.Core\Capabilities.Core.fsproj", "{B3BD88DC-7362-46F8-8040-6574C7B6B759}"
+Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "Capabilities.Core", "Samples\Capabilities.Core\Capabilities.Core.fsproj", "{B3BD88DC-7362-46F8-8040-6574C7B6B759}"
```
Evangelink, line 71: Auto-changed by my VS, I can reset if you want.

src/Samples/Multiselect/MainWindow.xaml
```diff
@@ -0,0 +1,44 @@
...
+        d:DataContext="{x:Static vm:Program.designVm}">
+  <Window.Resources>
+    <ObjectDataProvider x:Key="dataFromEnum" MethodName="GetValues"
+                            ObjectType="{x:Type sys:Enum}">
+      <ObjectDataProvider.MethodParameters>
+        <x:Type TypeName="SelectionMode"/>
+      </ObjectDataProvider.MethodParameters>
+    </ObjectDataProvider>
+  </Window.Resources>
+  <StackPanel Margin="0,15,0,0">
```
Evangelink, line 22: Design is not awesome but I have copied what's done in the other examples.

src/Samples/Multiselect/MainWindow.xaml
```diff
@@ -0,0 +1,44 @@
...
+        xmlns:vm="clr-namespace:Elmish.WPF.Samples.Multiselect;assembly=Multiselect.Core"
+        xmlns:sys="clr-namespace:System;assembly=mscorlib"
+        WindowStartupLocation="CenterScreen"
+        Title="SubModelSelectedItem"
+        Height="350"
+        Width="500"
+        mc:Ignorable="d"
+        d:DataContext="{x:Static vm:Program.designVm}">
+  <Window.Resources>
+    <ObjectDataProvider x:Key="dataFromEnum" MethodName="GetValues"
```
Evangelink, line 15: This could come from the "view model" if you prefer.

src/Samples/Multiselect/MainWindow.xaml
```diff
@@ -0,0 +1,44 @@
...
+        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+        xmlns:vm="clr-namespace:Elmish.WPF.Samples.Multiselect;assembly=Multiselect.Core"
+        xmlns:sys="clr-namespace:System;assembly=mscorlib"
+        WindowStartupLocation="CenterScreen"
+        Title="SubModelSelectedItem"
+        Height="350"
+        Width="500"
+        mc:Ignorable="d"
+        d:DataContext="{x:Static vm:Program.designVm}">
```
Evangelink, line 13: I have some problem in my VS with this declaration but as far as I can tell, it's exactly defined as in the other projects were I don't have an issue.

src/Elmish.WPF.sln
TysonMN: This change is correct.  Not sure what I did to get the legacy GUID.

I never remember which one is which either.  I have to search for [this issue](https://github.com/dotnet/project-system/issues/2863) every time.

src/Samples/Multiselect.Core/Program.fs
```diff
@@ -54,4 +54,4 @@ let main window =
 
   WpfProgram.mkSimple init update bindings
   |> WpfProgram.withLogger (new SerilogLoggerFactory(logger))
-  |> WpfProgram.startElmishLoop window
+  |> WpfProgram.startElmishLoop window
```
Evangelink, line 57: FYI the empty line at the end of the files was added as a convenience for git diffs.

TysonMN (approved):
Interesting.  I was expecting a change to the core code.  Instead, you "merely" added a sample.  Seems like an obvious strict improvement, so I will merge it in.

I will consider refactoring the single- and multi-selection samples now that have a concrete proposal.  My plan is to continue spending my time elsewhere (such as dispatch wrapping), but I will prioritize anything related to you (such as reviewing PRs, having discussions in issues, or making code changes for you...such as giving you an example to follow or making some change easier for you).

> I have made some tests and it seems that switching over to `SelectionMode = Single` correctly updates the model to contain only 1 selected item. Obviously, as described in the linked ticket, the following approach is limited when we want to ensure only one item is selected because the model represents the multi-selection.

I am curious about this.  For single selection, I typically prefer to store an ID or index of the selected item in the parent.  That ensures that there really is at most one item selected (but also allows the case that the model suggests there is something selected, but there isn't an item with that ID/index).  For multiselection, it seems better to have each item know if it is selected.  If the parent stores this separately, then the same error is possible, but I see no downside to each child storing its own selection state.

src/Samples/Multiselect.Core/Program.fs
TysonMN: Sure.  I used to add final blank line until I started using GitKraken.  Unlike every other git diff visualization I have seen (such as by Source Tree or GitHub), I think their diff visualization is better when there is no ending blank line.

I don't feel strongly about this though.

PR #410, opened 2021-08-14
Fix a few typos in the comments

Evangelink:
Fix a few typos in the code comments.

### DIFF
--- a/src/Elmish.WPF/Binding.fs
+++ b/src/Elmish.WPF/Binding.fs
@@ -800,10 +800,10 @@ module Binding =
 
   /// <summary>
   ///   Accepts a dispatch wrapping function.
-  ///   This can be used to debouce, throttle, or limit this binding.
+  ///   This can be used to debounce, throttle, or limit this binding.
   ///   If more than one dispatching wrapping is added to a single binding,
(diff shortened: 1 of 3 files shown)


src/Elmish.WPF/Binding.fs
```diff
@@ -21,7 +21,7 @@ type internal OneWaySeqLazyData<'model, 'a, 'b, 'id when 'id : equality> =
     Equals: 'a -> 'a -> bool
     GetId: 'b -> 'id
     ItemEquals: 'b -> 'b -> bool }
-
```
Evangelink, line 24: Sorry, I will revert all of these whitespace changes (I forgot again to disable the VS extension removing them on save).

TysonMN:
Thanks!

src/Elmish.WPF/Binding.fs
TysonMN: I prefer to have no trailing spaces at the end of a line, so your PR removes some of them, then feel free to leave them in (as long as the diff is easy to understand).

Evangelink: I will do a separate PR with trailing space removal.

PR #420, opened 2021-09-01
Restore the wrapDispatch parameter

### LINKED ISSUE #413: Avoid breaking change in v4 because of wrapDispatch
PR #256 introduced what would be a breaking change for v4.  But now because of the progress in https://github.com/elmish/Elmish.WPF/issues/271#issuecomment-902369178, that PR can essentially be reverted.  Let's not introduce a breaking change in v4 because of this `wrapDispatch` parameter.

However, we are not keeping these methods forever.  Instead, any method that accepts a `wrapDispatch` parameter will also be given the `Obsolete` attribute.  The text should say that this method will be removed in v5 and that users should prepare for that by replacing this method call with its implementation.  Maybe the implementation (aka the suggested replacement) could also be given in the documentation so that the user can see it in the tooltip when hovering over this method.

One thing that is a bit annoying about this is that there are probably some methods that we want keep long term _after_ removing the optional `wrapDispatch` parameter.  In those cases, I think the correct change is to make two copies of this method: one without the parameter and one with the parameter.  Then only the one with the parameter would be marked as `Obsolete`.

Would anyone like to do this work?  (Maybe @Evangelink?)

This work can either begin after I push the new wrap dispatch code or before by adding a placeholder for the needed function API call.

Evangelink:
This restored function is marked as obsolete with a suggestion to use the newly added alternate overload without the parameter.

Fixes #413

### DIFF
--- a/src/Elmish.WPF/Binding.fs
+++ b/src/Elmish.WPF/Binding.fs
@@ -2155,6 +2155,26 @@ type Binding private () =
   /// </summary>
   /// <param name="exec">Returns the message to dispatch.</param>
   /// <param name="canExec">Indicates whether the command can execute.</param>
+  /// <param name="wrapDispatch">
+  ///   Wraps the dispatch function with additional behavior, such as
+  ///   throttling, debouncing, or limiting.
+  /// </param>
+  [<System.Obsolete("In version 5, this method will be removed.  Use the overload without the \"wrapDispatch\" parameter followed by a call to \"Binding.alterMsgStream\".  For an example, see how this method is implemented.")>]
+  static member cmdParamIf
+      (exec: obj -> 'model -> 'msg,
+       canExec: obj -> 'model -> bool,
+       wrapDispatch: Dispatch<'msg> -> Dispatch<'msg>)
+      : string -> Binding<'model, 'msg> =
+    Binding.cmdParamIf (exec, canExec)
+    >> Binding.alterMsgStream wrapDispatch
+
+  /// <summary>
+  ///   Creates a <c>Command</c> binding that depends on the
+  ///   <c>CommandParameter</c>
+  ///   and can execute if <paramref name="canExec" /> returns <c>true</c>.
+  /// </summary>
+  /// <param name="exec">Returns the message to dispatch.</param>
+  /// <param name="canExec">Indicates whether the command can execute.</param>
   /// <param name="uiBoundCmdParam">
   ///   If <c>true</c>, <c>CanExecuteChanged</c> will trigger every time WPF's
   ///   <c>CommandManager</c>
@@ -2171,11 +2191,11 @@ type Binding private () =
   static member cmdParamIf
       (exec: obj -> 'model -> 'msg,
        canExec: obj -> 'model -> bool,
-       ?uiBoundCmdParam: bool,
-       ?wrapDispatch: Dispatch<'msg> -> Dispatch<'msg>)
+       uiBoundCmdParam: bool,
+       wrapDispatch: Dispatch<'msg> -> Dispatch<'msg>)
       : string -> Binding<'model, 'msg> =
-    Binding.cmdParamIf (exec, canExec, defaultArg uiBoundCmdParam false)
-    >> (wrapDispatch |> Option.map Binding.alterMsgStream |> Option.defaultValue id)
+    Binding.cmdParamIf (exec, canExec, uiBoundCmdParam)
+    >> Binding.alterMsgStream wrapDispatch
 
 
   /// <summary>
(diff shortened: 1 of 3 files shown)


src/Elmish.WPF/Binding.fs
```diff
@@ -1263,6 +1263,22 @@ type Binding private () =
...
 
+  /// <summary>Creates a two-way binding.</summary>
+  /// <param name="get">Gets the value from the model.</param>
+  /// <param name="set">Returns the message to dispatch.</param>
+  /// <param name="wrapDispatch">
+  ///   Wraps the dispatch function with additional behavior, such as
+  ///   throttling, debouncing, or limiting.
+  /// </param>
+  [<System.Obsolete("In version 5, this method will be removed.  Use the overload without the \"wrapDispatch\" parameter followed by a call to \"Binding.alterMsgStream\".  For an example, see how this method is implemented.")>]
+  static member twoWay
```
Evangelink, line 1274: Add an obsolete overload with the `wrapDispatch` parameter and call the non-obsolete overload followed by the call to `Binding.alterMsgStream wrapDispatch`.

Evangelink:
@TysonMN As suggested in the ticket, I am creating a draft PR with the parameter restored only for the `twoWay` function. I will wait for a go/no go before moving forward.

TysonMN:
Looks great.  Feel free to either do more partial work and get additional feedback or just do everything.  Whatever you want.

src/Elmish.WPF/Binding.fs
Evangelink: I totally missed this edge case. I have tried a few things and I can't find any good way to solve this so yeah +1 on saying let's have a potential breaking change edge case.

src/Elmish.WPF/Binding.fs
```diff
@@ -1263,6 +1263,22 @@ type Binding private () =
     >> Binding.mapModel get
     >> Binding.mapMsgWithModel set
 
+  /// <summary>Creates a two-way binding.</summary>
```
Evangelink, line 1266: FYI I went with 1 empty line between the obsolete overload and the non-obsolete one and kept the 2 lines between methods.

src/Elmish.WPF/Binding.fs
```diff
@@ -1684,6 +2148,35 @@ type Binding private () =
...
+  /// </param>
+  [<System.Obsolete("In version 5, this method will be removed.  Use the overload without the \"wrapDispatch\" parameter followed by a call to \"Binding.alterMsgStream\".  For an example, see how this method is implemented.")>]
+  static member cmdParamIf
+      (exec: obj -> 'model -> 'msg,
+       canExec: obj -> 'model -> bool,
+       ?uiBoundCmdParam: bool,
+       ?wrapDispatch: Dispatch<'msg> -> Dispatch<'msg>)
+      : string -> Binding<'model, 'msg> =
+    Binding.cmdParamIf (exec, canExec, defaultArg uiBoundCmdParam false)
+    >> Binding.alterMsgStream (defaultArg wrapDispatch id)
```
Evangelink, line 2178: This is causing `alterMsgStream` to be called everytime (sometimes with `id`), it might be better to check if the arg is defined and only call the alter then.

Evangelink: It could be something like

```suggestion
    >> match wrapDispatch with | Some wrapDispatch -> Binding.alterMsgStream wrapDispatch | None -> id
```

Or we could add some helper like `mapBoth` or whatever name you prefer.

Evangelink:
Moving back to draft, I totally forgot to check the tests.

src/Elmish.WPF/Binding.fs
TysonMN: Actually, we could also add an overload that accepts `Option<>`.  Maybe this idea will help in the more complicated cases below.

src/Elmish.WPF/Binding.fs
TysonMN: It can be made more terse:
```fsharp
>> (wrapDispatch |> Option.map Binding.alterMsgStream |> Option.defaultValue id)
```

src/Elmish.WPF/Binding.fs
TysonMN: > I think the only edge case is if the user explicitly passed in an `Option<>`. I think that is possible in `v3` (thought I could be wrong),

I just tested; I am wrong.  Passing in `Option<>` for an optional parameter is not possible.

src/Elmish.WPF/Binding.fs
```diff
@@ -1684,6 +2148,35 @@ type Binding private () =
...
+  ///   ability to execute. This will likely lead to many more triggers than
+  ///   necessary, but is needed if you have bound the <c>CommandParameter</c>
+  ///   to another UI property.
+  /// </param>
+  /// <param name="wrapDispatch">
+  ///   Wraps the dispatch function with additional behavior, such as
+  ///   throttling, debouncing, or limiting.
+  /// </param>
+  [<System.Obsolete("In version 5, this method will be removed.  Use the overload without the \"wrapDispatch\" parameter followed by a call to \"Binding.alterMsgStream\".  For an example, see how this method is implemented.")>]
+  static member cmdParamIf
```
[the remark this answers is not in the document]
TysonMN: For this one method with two optional parameters, how about creating these three overloads?
1. optional `uiBoundCmdParam` and no `wrapDispatch`
2. no `uiBoundCmdParam` and required `wrapDispatch` with `Obsolete` attribute
3. required `uiBoundCmdParam` and required `wrapDispatch` with `Obsolete` attribute

Two optional parameters means there is four ways to pass in arguments.  These three overloads cover those four cases because the first overload covers two cases and the other two overloads each cover one case.

Does this sound correct to you?

Evangelink:
(sorry for the multiple extra commits, I am on a computer without the requiered dev tools).

TysonMN:
Doesn't matter.  You can create as many commits as you want.  We can always squash later if we want.

Evangelink:
@TysonMN No rush for the review, I just wanted to let you know that PR was ready.

Evangelink:
@TysonMN The last few commits sucks a bit but that's not a big deal if you are find with the structure :)

TysonMN:
Excellent work, and thanks again!

PR #429, opened 2021-09-12
Replace uses of ! and := with .Value

### LINKED ISSUE #414: Replace uses of ! and := with .Value
The operators `!` and `:=` will eventually be deprecated.  Replace these with `.Value` as described in
https://github.com/fsharp/fslang-design/blob/main/preview/FS-1111-refcell-op-information-messages.md

I think all of our uses are currently in the `ViewModel` file.

(Hat tip @dharmaturtle)

Evangelink:
Fixes #414


TysonMN:
Perfect.  Thanks :)

PR #430, opened 2021-09-12
Omit <summary> if no other XML tags

### LINKED ISSUE #416: Omit <summary> if no other XML tags
> If a `///` comment does not start with a `<`, then the entire comment text is taken as the summary documentation for the code construct that immediately follows. Use this method when you want to write only a brief summary for each construct.

https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/xml-documentation#comments-without-xml-tags

Evangelink:
Fixes #416

### DIFF
--- a/src/Elmish.WPF/Binding.fs
+++ b/src/Elmish.WPF/Binding.fs
@@ -796,53 +796,53 @@ module Binding =
 
 
   module OneWay =
-    ///   Elemental instance of a one-way binding.
+    /// Elemental instance of a one-way binding.
     let id<'a, 'msg> : string -> Binding<'a, 'msg> =
       { Get = box }
       |> OneWayData
       |> BaseBindingData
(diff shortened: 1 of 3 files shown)


src/Elmish.WPF/Binding.fs
```diff
@@ -796,89 +796,71 @@ module Binding =
 
 
   module OneWay =
-    /// <summary>
     ///   Elemental instance of a one-way binding.
```
Evangelink, line 799: I haven't changed the spacing indentation for the content. Let me know if you prefer them to be updated and I will push a change.

TysonMN: Yes, let's simply that to just
```fsharp
/// Elemental ...
```

TysonMN:
Thanks :)

PR #432, opened 2021-09-12
Add withErrorHandler and withMsgErrorHandler

### LINKED ISSUE #426: Add WpfProgram wrapper for Program.withErrorHandler
I would like to display a message to the user when there's an exception in the Elmish loop. Based on the name and the signature, `Program.onError` seems to fit the bill, though the description says "For library authors only", so I'm not 100% sure it's what I'm after.

In any case, I can't use `Program.onError` (or anything else in Elmish's `Program` module) because the `WpfProgram` fields are internal, and the `WpfProgram` module doesn't expose any function that wraps `onError` or allows mapping `Program`. (Allowing users to map `Program` could be a more general solution, though it would be possible to shoot oneself in the foot.)

cmeeren:
Fixes #426

I also added `withMsgErrorHandler`, where you supply a handler that returns `'msg list` instead of `unit`, and the returned messages are then dispatched by Elmish.WPF. As I said in https://github.com/elmish/Elmish.WPF/issues/426#issuecomment-916045586, this functionality can be implemented in user code, but it requires more boilerplate impure code (events/subscriptions) and was so trivial to implement in Elmish.WPF that I see no compelling reason not to support it.

### DIFF
--- a/src/Elmish.WPF/WpfProgram.fs
+++ b/src/Elmish.WPF/WpfProgram.fs
@@ -11,7 +11,7 @@ type WpfProgram<'model, 'msg> =
     ElmishProgram: Program<unit, 'model, 'msg, unit>
     Bindings: Binding<'model, 'msg> list
     LoggerFactory: ILoggerFactory
-    ErrorHandler: string * exn -> 'msg list
+    ErrorHandler: string -> exn -> 'msg list
     /// Only log calls that take at least this many milliseconds. Default 1.
     PerformanceLogThreshold: int
   }
@@ -25,7 +25,7 @@ module WpfProgram =
     { ElmishProgram = program
       Bindings = getBindings ()
       LoggerFactory = NullLoggerFactory.Instance
-      ErrorHandler = fun _ -> []
+      ErrorHandler = fun _ _ -> []
       PerformanceLogThreshold = 1 }
 
 
@@ -96,7 +96,7 @@ module WpfProgram =
 
     let errorHandler (msg: string, ex: exn) =
       updateLogger.LogError(ex, msg)
-      program.ErrorHandler (msg, ex) |> List.iter dispatch
+      program.ErrorHandler msg ex |> List.iter dispatch
 
     program.ElmishProgram
     |> if updateLogger.IsEnabled LogLevel.Trace then Program.withTrace logMsgAndModel else id
--- a/src/Elmish.WPF/WpfProgram.fs
+++ b/src/Elmish.WPF/WpfProgram.fs
@@ -164,7 +164,7 @@ module WpfProgram =
   /// message case.
   ///
   /// The update handler for the dispatched messages MUST NOT cause exceptions themselves,
-  /// or the app will enter into an infinite loop.
+  /// or the app may enter into an infinite loop.
   let withErrorHandler onError program =
     { program with ErrorHandler = onError }
 
--- a/src/Elmish.WPF/WpfProgram.fs
+++ b/src/Elmish.WPF/WpfProgram.fs
@@ -160,10 +160,14 @@ module WpfProgram =
 
   /// Calls the specified function for unhandled exceptions in the Elmish
   /// dispatch loop (e.g. in commands or the update function). This essentially
-  /// delegates to Elmish's Program.withErrorHandler. The first (string)
-  /// argument of onError is a message from Elmish describing the context of the
-  /// exception. Note that this may contain a rendered message case with all
-  /// data ("%A" formatting).
+  /// delegates to Elmish's Program.withErrorHandler.
+  ///
+  /// The first (string) argument of onError is a message from Elmish describing
+  /// the context of the exception. Note that this may contain a rendered
+  /// message case with all data ("%A" formatting).
+  ///
+  /// Note that exceptions passed to onError are also logged to the logger
+  /// specified using WpfProgram.withLogger.
   let withElmishErrorHandler onError program =
     { program with ErrorHandler = onError }
 
--- a/src/Elmish.WPF/WpfProgram.fs
+++ b/src/Elmish.WPF/WpfProgram.fs
@@ -11,8 +11,7 @@ type WpfProgram<'model, 'msg> =
     ElmishProgram: Program<unit, 'model, 'msg, unit>
     Bindings: Binding<'model, 'msg> list
     LoggerFactory: ILoggerFactory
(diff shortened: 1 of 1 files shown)


src/Elmish.WPF/WpfProgram.fs
```diff
@@ -11,6 +11,8 @@ type WpfProgram<'model, 'msg> =
     ElmishProgram: Program<unit, 'model, 'msg, unit>
     Bindings: Binding<'model, 'msg> list
     LoggerFactory: ILoggerFactory
+    ErrorHandler: string * exn -> unit
+    MsgErrorHandler: string * exn -> 'msg list
```
TysonMN, line 15: About just `ErrorHandler: string * exn -> 'msg list` since the variant returning `unit` can be implemented by calling this and returning `[]`?

cmeeren: Excellent suggestion. Should we do the same for the two public functions? Just note in the docs that the function may be impure and return an empty list? Or do you prefer keeping them separate? (I may lean very slightly toward keeping them separate, but I don't really have a strong opinion on this.)

TysonMN: Let's just have one function.

This introduces a point of fragility, which is that an exception causes this handler to execute, which dispatches a message, which causes an exception, and so on until the application crashes.

You said you had it return `'msg list` because you didn't see a compelling reason not to add it.  Is this reason compelling enough not to add it?  Do you actually want the ability to dispatch a message on an exception?  If not, then I would prefer to avoid point of fragility.

cmeeren: > This introduces a point of fragility, which is that an exception causes this handler to execute, which dispatches a message, which causes an exception, and so on until the application crashes.

Good catch. I just verified that the app does indeed enter into an infinite loop.

Note that the effect is the same if we don't allow dispatching messages from the error handler, and the user instead uses e.g. events and subscriptions to dispatch messages on exceptions. That would be the go-to method of getting messages into the loop. There may be valid reasons for handling exceptions in-loop, e.g. modifying the user interface as with any other message (instead of just showing a separate dialog).

I think we should support this, but make it clear that the update handler for the dispatched messages must never throw exceptions, or the app will enter into an infinite loop.

I pushed a commit that is my suggestion. Let me know what you think.

src/Elmish.WPF/WpfProgram.fs
```diff
@@ -155,6 +158,17 @@ module WpfProgram =
...
 
 
+  /// Uses the specified handler for dispatch loop exceptions, dispatching the returned
+  /// messages when an exception occurs. The first (string) argument of onError is message
+  /// from Elmish describing the context of the exception, and may contain a rendered
+  /// message case.
+  ///
+  /// The update handler for the dispatched messages MUST NOT cause exceptions themselves,
+  /// or the app will enter into an infinite loop.
+  let withErrorHandler onError program =
```
cmeeren, line 168: `onError` here has the signature `string * exn -> 'msg list`. Personally I would favor `string -> exn -> 'msg list`. The former is consistent with Elmish's own `withErrorHandler`, but that is not really relevant for Elmish.WPF users, since they don't use the Elmish.Program module directly. I suggest we change the signature of `onError` to `string -> exn -> 'msg list`.

TysonMN: Yes, let use the curried version

src/Elmish.WPF/WpfProgram.fs
TysonMN: If you only intend to use this feature via a function that returns an empty just, then I prefer that the return type be `unit`.  I understand that the user could still get a message dispatched by jumping through more hoops, but this additional fraction is intentional in order to discourage this design.

src/Elmish.WPF/WpfProgram.fs
```diff
@@ -155,6 +158,17 @@ module WpfProgram =
     { program with LoggerFactory = loggerFactory }
 
 
+  /// Uses the specified handler for dispatch loop exceptions, dispatching the returned
+  /// messages when an exception occurs. The first (string) argument of onError is message
+  /// from Elmish describing the context of the exception, and may contain a rendered
+  /// message case.
+  ///
```
TysonMN, line 165: The `ILogger` logs many different kinds of errors.  This handler you are adding only logs some of those errors.  I would like this name to indicate which subset of errors that it handles.

TysonMN (changes requested):
I added a comment on a line, and FastHub is requiring me to enter a message for the review as a whole.

src/Elmish.WPF/WpfProgram.fs
cmeeren: Sounds good to me. YAGNI and all that. Just pushed this change.

src/Elmish.WPF/WpfProgram.fs
cmeeren: I'm not sure exactly what you're after, but I updated the doc comment (before reading your comment here). Does that resolve this?

TysonMN: The name that comes to my mind is `withElmishErrorHandler`.  How does that sound to you?

cmeeren: Sure. I also updated the doc comment. How do you like it?

src/Elmish.WPF/WpfProgram.fs
```diff
@@ -155,6 +158,16 @@ module WpfProgram =
     { program with LoggerFactory = loggerFactory }
 
 
+  /// Calls the specified function for unhandled exceptions in the Elmish
+  /// dispatch loop (e.g. in commands or the update function). This essentially
+  /// delegates to Elmish's Program.withErrorHandler. The first (string)
+  /// argument of onError is a message from Elmish describing the context of the
+  /// exception. Note that this may contain a rendered message case with all
+  /// data ("%A" formatting).
```
TysonMN, line 166: One last thing.

Let's have this documentation say something about the relationship between this function and `WpfProgram.withLogger`.  Something about how the Emlish errors are also send to the logger, so providing both a logger and an Elmish error hander will result in being informed of this occurrence twice.

cmeeren: Done

TysonMN:
Ok, done.  Thanks :)

PR #455, opened 2022-03-30
Update TUTORIAL.md

min44:
Correction of a typo.


TysonMN:
Thanks for contributing a PR.

However,  `Increment` and `Decrement` are indeed get-only properties.  WPF gets the `ICommand`s from those bindings and never tries to set any `ICommand` instances to those bindings.

PR #458, opened 2022-04-04
Reverted match case rearrangement that disabled stickiness

### LINKED ISSUE #456: "Sticky" sample is not sticky
**High-level description**
The Sticky sample is supposed to update only when a new even number is given, but it updates with every increment, even or odd.

**Steps to Reproduce**

1. Run the Sticky sample.
2. Click on the "+" button three times.

**Expected or desired behavior**
The counter now reads "2", the most recent even number.

**Actual behavior**
The counter now reads "3", which is not even.

**Additional Information**
Elmish.WPF version: 4.0.0-beta-43
Target framework version: .NET 5.0

From `BindingInternal.BindingData.addSticky`:

```F#
let addSticky (predicate: 'model -> bool) (binding: BindingData<'model, 'msg>) =
  let mutable stickyModel = None
  let f newModel =
    match predicate newModel, stickyModel with
    | _, None ->
        newModel
    | true, _ ->
        stickyModel <- Some newModel
        newModel
    | false, Some sm ->
        sm
  binding |> mapModel f
```

`stickyModel` seems always to be set to `None`, so the `match` expression always returns `newModel` from the first branch. I hypothesize that `stickyModel` resets on every update.

How can we resolve this? Should we add another case to `BindingData`?

the-not-mad-psychologist:
I found the commit that bugged sticky bindings. I think it was meant to be a refactor, but it changed behavior. The Sticky sample now works.

Resolves #456.


the-not-mad-psychologist:
I believe that, in the current code, `stickyModel` is never set.

TysonMN:
Excellent!

I want to include a test as well...something that I didn't originally do (my bad).

Do you think you could add such a test?  If not, I can add it.

TysonMN:
Note to self:

Consider this implementation:

```fsharp
let f newModel =
     if predicate newModel then
          stickyModel <- Some newModel
          newModel
      else
          stickyModel |> Option.defaultValue newModel
```

the-not-mad-psychologist:
> Do you think you could add such a test? If not, I can add it.

Done.

> ```fsharp
> let f newModel =
>      if predicate newModel then
>           stickyModel <- Some newModel
>           newModel
>       else
>           stickyModel |> Option.defaultValue newModel
> ```

I think this much more readable than the current implementation. Shall I add it to this PR since we're working with this code anyway?

TysonMN:
> > Do you think you could add such a test? If not, I can add it.
> 
> Done.

Wonderful test!  Great work :D

> I think this is much more readable than the current implementation. Shall I add it to this PR since we're working with this code anyway? It passes the test.

I went ahead and added it since I am ready to complete this PR.

I also force pushed to your branch to change the order of your commit to one that I prefer.  Specifically, I like to be able to checkout the the commit that adds the test and [see it fail for the right reason](https://blog.ploeh.dk/2019/10/21/a-red-green-refactor-checklist/).

PR #477, opened 2022-06-10
Remove lazy effect and model mapping from OneWaySeq binding

TysonMN:


TysonMN:
@[marner2](https://github.com/marner2), I tried to add you as a reviewer, but GitHub won't allow it.

src/Elmish.WPF/Binding.fs
```diff
@@ -498,7 +498,9 @@ type Binding private () =
        itemEquals: 'b -> 'b -> bool,
        getId: 'b -> 'id)
       : string -> Binding<'model, 'msg> =
-    BindingData.OneWaySeqLazy.create get equals map itemEquals getId
```
TysonMN, line 501: I should a add this (beta-breaking) change to the release notes

src/Elmish.WPF/BindingInternal.fs
```diff
@@ -407,60 +401,46 @@ module internal BindingData =
         (mSet "set")
 
 
-  module OneWaySeqLazy =
+  module OneWaySeq =
 
     let mapMinorTypes
         (outMapA: 'a -> 'a0)
```
the-not-mad-psychologist, line 407: Shouldn't `outMapA` and `inMapA` be renamed to `outMap` and `inMap` since they are no longer distinguished from other map functions?

the-not-mad-psychologist (approved):

src/Elmish.WPF/BindingInternal.fs
marner2: `outMapId` needs to be distinguished

src/Elmish.WPF/Binding.fs
marner2: Was this the only place this was used? I thought there were more...

src/Elmish.WPF/BindingInternal.fs
```diff
@@ -168,11 +164,9 @@ module internal BindingData =
       | OneWayToSourceData d -> OneWayToSourceData {
           Set = binaryHelper d.Set
         }
-      | OneWaySeqLazyData d -> OneWaySeqLazyData {
+      | OneWaySeqData d -> OneWaySeqData {
           Get = f >> d.Get
-          Map = d.Map
```
marner2, line 173: I thought we were going to have issues here, but yeah I do see now how this is fully redundant with `Get`

marner2 (approved):
I'm pretty sure this will conflict a bit with my existing stuff. I removed #471 since it conflicts heavily, but I'm also adding types here in #473

src/Elmish.WPF/Binding.fs
TysonMN: Oh, that is not part of the public API.  Instead, I should expose something like this to the public API.

src/Elmish.WPF.Tests/ViewModelTests.fs
```diff
@@ -526,7 +526,7 @@ module OneWaySeqLazy =
     }
 
 
-  [<Fact>]
+  [<Fact>] // TODO: this test should pass, but it doesn't. Need to merge PR 478 first.
```
TysonMN, line 529: Like the comment says, PR #478 needs to merge first so this test will pass

marner2: This makes sense. So basically, by adding the lazy effect, it was calling "Get" once for each side of the equality check. Afer PR #478, it has the getter right there to call (which doesn't affect this call).

marner2 (approved):
Approved for when the other gets merged in and everything passes.

TysonMN:
Nice!  I can "Update Branch" (i.e. rebase the branch) from the GitHub app.  I had done a merge commit, so I will keep doing that.

src/Elmish.WPF.Tests/ViewModelTests.fs
TysonMN: > [...] by adding the lazy effect, it was calling "Get" once for each side of the equality check. Afer PR #478, it has the getter right there to call (which doesn't affect this call).

Yes, the `Get` was composed into both "sides" of the equality.  That is where two calls came from. The third call was in the code that prepares to call merge.  This is because the model mapping in the Lazy case mapped both the equals function and the nested binding.

Now this third call won't happen because the model that the merge preparation code needs is now passed down to it from the Lazy case.

marner2:
@TysonMN You waiting on something to merge this?

TysonMN:
Just the time in the day to click the button

PR #478, opened 2022-06-11
Optimize lazy effect by reducing calls to get

TysonMN:
Each recursive binding case should either do model/msg mapping on its own functions or its inner binding but not both.

Somewhat annoyingly, this increases duplicate code (which is why I didn't write it this way in the first place).  I wonder if there is a good way to dedup this.

This is also a good reminder that I should write more (view model) tests.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,6 @@
 #### 4.0.0-beta-44
 * Fixed sticky effect that was broken in `4.0.0-beta-42`
+* Improved performance of Lazy effect by reducing calls to later model mappings
 
 #### 4.0.0-beta-43
 * Added `WpfProgram.withElmishErrorHandler`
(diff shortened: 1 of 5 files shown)


TysonMN:
@marner2 and @the-not-mad-psychologist, want to review?

marner2:
An intuitive understanding for why this is necessary here but not other places is because `lazy` must call equals. Is that fair?

marner2 (approved):
Looks good to me.

TysonMN:
> An intuitive understanding for why this is necessary here but not other places is because `lazy` must call equals. Is that fair?

Yes.  I sort of knew about this inefficiency, but I did it anyway because it was easier than this change.  The alter message stream case is already doing "this".  The change in this PR is to essentially copy what I had done in that case.

It was very hard for me to get the alter message stream case to work.  The essential difficulty was getting recursion with different type parameters to work in F#.

Looking back now, I am sure that I didn't know how to solve that problem when I created the Lazy effect.  Now that I understand this, I should check if any of the other recursive cases also include this inefficiency.

marner2:
Yeah I think it's only the case where the recursive cases actually have to look at something about the model. `Validation` doesn't have to, and `Caching` surprisingly doesn't have to either (since you invalidate it blindly).

PR #483, opened 2022-06-13
Add cache tests

### LINKED ISSUE #185: Write test for behavior of CachedBInding's trySetMember
In the `Cached` case of `trySetMember`, the assignment of `None` (as opposed to `Some value`) to the field `Cache` is correct.  As pointed out in https://github.com/elmish/Elmish.WPF/pull/183#discussion_r344018327, assigning `None` allows format-as-you-type behavior.  We should create a test that depends on that behavior.
 https://github.com/elmish/Elmish.WPF/blob/7feeb9bc23e38334ea41bb36c2fbf334b7b04ba0/src/Elmish.WPF/ViewModel.fs#L698-L702

TysonMN:
Resolves #185

### DIFF
--- a/src/Elmish.WPF.Tests/DynamicViewModelTests.fs
+++ b/src/Elmish.WPF.Tests/DynamicViewModelTests.fs
@@ -1410,6 +1410,75 @@ module SubModelSelectedItem =
 
 
 
+module CacheEffect =
+
+  [<Fact>]
+  let ``model mapping called exactly once when Get called twice`` () =
+
+    Property.check <| property {
+      let! name = GenX.auto<string>
+      let! model = GenX.auto<int>
+      let! bindingComponsitionOrder = Gen.bool
+    
+      let mapping = InvokeTester id
+      let cachingAndMapping =
+        if bindingComponsitionOrder
+        then Binding.mapModel mapping.Fn >> Binding.addCaching
+        else Binding.addCaching >> Binding.mapModel mapping.Fn
+      let binding =
+        name
+        |> Binding.OneWay.id
+        |> cachingAndMapping
+      let vm = TestVm(model, binding)
+      
+      vm.Get name |> ignore // populate cache
+      vm.Get name |> ignore
+      
+      test <@ 1 = mapping.Count @>
+    }
+
+
+  [<Fact>]
+  let ``second Get returns new model after first Get and then Update`` () =
+    let name = ""
+    let model = 0
+    let newModel = 1
+    let binding =
+      name
+      |> Binding.OneWay.id
+      |> Binding.addCaching
+    let vm = TestVm(model, binding)
+
+    vm.Get name |> ignore   // populate cache
+    vm.UpdateModel newModel // clear cache
+    let actual = vm.Get name |> unbox
+
+    test <@ newModel = actual @>
+
+
+  [<Fact>]
+  let ``cache not cleared on Set`` () =
+    let name = ""
+    let initialModel = 0
+    let newModel = 1
+    let mapping = InvokeTester (fun x -> x)
+    let binding =
+      name
+      |> Binding.TwoWay.id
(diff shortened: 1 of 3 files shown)


src/Elmish.WPF.Tests/DynamicViewModelTests.fs
```diff
@@ -1410,6 +1410,75 @@ module SubModelSelectedItem =
...
+    let binding =
+      name
+      |> Binding.TwoWay.id
+      |> Binding.mapModel mapping.Fn
+      |> Binding.addCaching
+    let vm = TestVm(initialModel, binding)
+
+    vm.Get name |> ignore // populate cache
+    vm.Set name newModel
+    mapping.Reset()       // Set calls mapping function, so reset count
```
TysonMN, line 1474: @marner2, I don't think this caching effect can prevent the model mapping function from being called on `Set` when the cache is populated.  `Set` needs the model but the cache contains a different type.

Do you agree?

src/Elmish.WPF.Tests/DynamicViewModelTests.fs
```diff
@@ -1410,6 +1410,75 @@ module SubModelSelectedItem =
 
 
 
+module CacheEffect =
+
+  [<Fact>]
+  let ``model mapping called excactly once when Get called twice`` () =
```
the-not-mad-psychologist, line 1416: *exactly

src/Elmish.WPF/BindingVmHelpers.fs
```diff
@@ -699,10 +699,9 @@ type Set(value: obj) =
     match binding with
     | BaseVmBinding b -> this.Base(model, b)
     | Cached b ->
-        let successful = this.Recursive(model, b.Binding)
-        if successful then
-          b.Cache.Value <- None  // TODO #185: write test
-        successful
+        // Set only dispatches a message and while UpdateModel changes the model,
+        // so don't clear the cache here
```
the-not-mad-psychologist, line 703: Grammatically unclear. "While" would work, but "and while" doesn't when used like that. Maybe also add a period at the end.

the-not-mad-psychologist: Or do you mean `while UpdateModel`? I'm assuming you're not using a `while`-loop, though.

TysonMN: Thanks for the review.

I reworded. Is that clear now?

src/Elmish.WPF.Tests/DynamicViewModelTests.fs
marner2: I'm fighting very hard against this, but I think you're right. "Set" is really "DispatchMessage." The model isn't affected until you actually run "Update."

I agree.

TysonMN: Can you elaborate on what you mean by

> I'm fighting very hard against this

?

marner2: It just seems wrong to not bust the cache when you set it. But I think I'm pretty convinced that the "actual" set happens on `Update`.

src/Elmish.WPF/BindingVmHelpers.fs
the-not-mad-psychologist: Much better, thanks!

the-not-mad-psychologist (approved):

marner2:
Maybe as part of another PR. I think it’s clear enough here.

src/Elmish.WPF.Tests/DynamicViewModelTests.fs
marner2: Maybe as part of another PR. I think it’s clear enough here.

PR #501, opened 2022-07-21
Rename TryGetMember and TrySetMember to avoid confusion with system methods

LyndonGingerich:
`TryGetMember` -> `TryGet`
`TrySetMember` -> `TrySet`

This is in response to confusion while debugging `DynamicViewModel.TryGetMember`.

PR #502, opened 2022-07-21
Repaired missing log argument

LyndonGingerich:
Also added a test to make sure it won't happen again. This can cause some quite mysterious exceptions.

### DIFF
--- a/src/Elmish.WPF.Tests/DynamicViewModelTests.fs
+++ b/src/Elmish.WPF.Tests/DynamicViewModelTests.fs
@@ -1442,7 +1442,7 @@ module CacheEffect =
       let! name = GenX.auto<string>
       let! model = GenX.auto<int>
       let! bindingComponsitionOrder = Gen.bool
-    
+
       let mapping = InvokeTester id
       let cachingAndMapping =
         if bindingComponsitionOrder
@@ -1453,10 +1453,10 @@ module CacheEffect =
         |> Binding.OneWay.id
         |> cachingAndMapping
       let vm = TestVm(model, binding)
-      
+
       vm.Get name |> ignore // populate cache
       vm.Get name |> ignore
-      
+
       test <@ 1 = mapping.Count @>
     }
 
--- a/src/Elmish.WPF.Tests/DynamicViewModelTests.fs
+++ b/src/Elmish.WPF.Tests/DynamicViewModelTests.fs
@@ -1431,6 +1431,29 @@ module SubModelSelectedItem =
       test <@ vm.Dispatches = [ set (selectedSubModel |> ValueOption.map getId) m ] @>
     }
 
+  [<Fact>]
+  let ``attempting to select a nonexistent item throws RuntimeBinderException`` () =
+    let selectedItemName = "Foo"
+    let subModelSeqName = "Bar"
+    let bindings =
+      [ selectedItemName |> Binding.subModelSelectedItem (subModelSeqName, Some, ignore)
+        subModelSeqName |> Binding.subModelSeq ((fun _ -> []), ignore, (fun () -> [])) ]
+    let mutable error : string option = None
+    let loggingArgs =
+      { LoggingViewModelArgs.none
+        with
+          log =
+            { new Microsoft.Extensions.Logging.ILogger
+              with
+                member _.BeginScope _ = { new IDisposable with member _.Dispose() = () }
+                member _.IsEnabled _ = true
+                member _.Log (_, _, state, ex, formatter) = error <- formatter.Invoke(state, ex) |> Some } }
+    let viewModelArgs = ViewModelArgs.create 0.0 ignore "main" loggingArgs
+    let vm = DynamicViewModel(viewModelArgs, bindings)
+
+    raises<Microsoft.CSharp.RuntimeBinder.RuntimeBinderException> <@ vm.Get selectedItemName @>
+    test <@ error.Value.Contains "TryGetMember FAILED: Failed to find an element" @>
+
 
 
 module CacheEffect =
--- a/src/Elmish.WPF/DynamicViewModel.fs
+++ b/src/Elmish.WPF/DynamicViewModel.fs
@@ -118,7 +118,7 @@ type [<AllowNullLiteral>] internal DynamicViewModel<'model, 'msg>
           | Error e ->
               match e with
               | GetError.OneWayToSource -> log.LogError("[{BindingNameChain}] TryGetMember FAILED: Binding {BindingName} is read-only", nameChain, binder.Name)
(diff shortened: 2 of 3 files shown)


src/Elmish.WPF.Tests/DynamicViewModelTests.fs
```diff
@@ -1431,6 +1431,35 @@ module SubModelSelectedItem =
...
+        { LoggingViewModelArgs.none
+          with
+            log =
+              { new Microsoft.Extensions.Logging.ILogger
+                with
+                  member _.BeginScope _ = { new IDisposable with member _.Dispose() = () }
+                  member _.IsEnabled _ = true
+                  member _.Log (_logLevel, _eventId, state, excn, formatter) =
+                    let s = formatter.Invoke(state, excn)
+                    if excn = null then error <- Some s } }
```
marner2, line 1454: I don't think we actually need this `if` statement.

marner2 (approved):
There might be a bunch more errors that we should be testing with the logger. I anticipate extracting out the logger into some kind of helper then writing a whole series of tests for the logger. In any case, this looks good for now.

src/Elmish.WPF.Tests/DynamicViewModelTests.fs
LyndonGingerich: You're right.

src/Elmish.WPF.Tests/DynamicViewModelTests.fs
```diff
@@ -1431,6 +1431,35 @@ module SubModelSelectedItem =
...
+        { LoggingViewModelArgs.none
+          with
+            log =
+              { new Microsoft.Extensions.Logging.ILogger
+                with
+                  member _.BeginScope _ = { new IDisposable with member _.Dispose() = () }
+                  member _.IsEnabled _ = true
+                  member _.Log (_logLevel, _eventId, state, excn, formatter) =
+                    let s = formatter.Invoke(state, excn)
+                    error <- Some s } }
```
marner2, line 1454:
```suggestion
                    error <- formatter.Invoke(state, excn) |> Some } }
```
Does this work?

src/Elmish.WPF/DynamicViewModel.fs
```diff
@@ -118,7 +118,7 @@ type [<AllowNullLiteral>] internal DynamicViewModel<'model, 'msg>
           | Error e ->
               match e with
               | GetError.OneWayToSource -> log.LogError("[{BindingNameChain}] TryGetMember FAILED: Binding {BindingName} is read-only", nameChain, binder.Name)
-              | GetError.SubModelSelectedItem d -> log.LogError("[{BindingNameChain}] TryGetMember FAILED: Failed to find an element of the SubModelSeq binding {SubModelSeqBindingName} with ID {ID} in the getter for the binding {SubModelSelectedItemName}", d.NameChain, d.SubModelSeqBindingName, d.Id)
+              | GetError.SubModelSelectedItem d -> log.LogError("[{BindingNameChain}] TryGetMember FAILED: Failed to find an element of the SubModelSeq binding {SubModelSeqBindingName} with ID {ID} in the getter for the binding {BindingName}", d.NameChain, d.SubModelSeqBindingName, d.Id, binder.Name)
```
TysonMN, line 121: > - ...{SubModelSelectedItemName}...
> + ...{BindingName}...

I had picked `SubModelSelectedItemName` since there are two bindings involved, and I didn't want any confusion.  Although, the other binding is called `SubModelSeqBindingName`, so by process of elimination, using `BindingName` here is still unambiguous.

Do you prefer `BindingName` because that is the name used in every other log event?

src/Elmish.WPF.Tests/DynamicViewModelTests.fs
```diff
@@ -1431,6 +1431,34 @@ module SubModelSelectedItem =
...
+                with
+                  member _.BeginScope _ = { new IDisposable with member _.Dispose() = () }
+                  member _.IsEnabled _ = true
+                  member _.Log (_logLevel, _eventId, state, excn, formatter) =
+                    error <- formatter.Invoke(state, excn) |> Some } }
+      let viewModelArgs = ViewModelArgs.create 0.0 ignore "main" loggingArgs
+
+      let vm = DynamicViewModel(viewModelArgs, bindings)
+      raises<Microsoft.CSharp.RuntimeBinder.RuntimeBinderException> <@ vm.Get selectedItemName @>
+      let error = error
```
TysonMN, line 1458: This workaround isn't needed any more
https://github.com/dotnet/fsharp/issues/9291

src/Elmish.WPF/DynamicViewModel.fs
LyndonGingerich: While the other interpolations are directly named after their fields on `d`, it was difficult, as one unfamiliar with the code, to discover what argument was supposed to go with `SubModelSelectedItemName`. I thought that the name `BindingName` more clearly indicated that it was supposed to take `binder.Name`. But then, code isn't necessarily optimized for those new to the project. What think you?

LyndonGingerich: Also, I thought that, since this `match` branch has to do mainly with `subModelSelectedItem` bindings, the type of the matched binding is probably obvious.

marner2: > Do you prefer BindingName because that is the name used in every other log event?

Yes. If you use a logging system like seq, and you want to search for all log events for a given binding name, you would have to include this one explicitly for this log case in your query expression. It's *much better* that they be named similarly for this purpose.

TysonMN:
If I click `Update Branch`, then a merge commit will be created. Instead, I want to rebase this on `master`.

Is there any way to do this on Android?  I tried several apps this morning, but couldn't find anything.

marner2:
@TysonMN I’m not able to do anything with this branch because @LyndonGingerich created it. But if you use the web browser view and zoom out and request the desktop site, it gives you the dropdown:
CA5B6FCC-45DB-4358-A930-95FCF6AC948B

marner2:
@TysonMN Actually it works at the normal mobile web viewer and the normal zoom level. I think the answer is just to use the browser instead of the app.

TysonMN:
Ok. I clicked the button to see if it would give me the choice to rebase. No dice. I will try the website on the other PR.

TysonMN:
That worked (of course).  Now I am curious to try again in the app just to see if it uses the new default choice for me from the website.  Probably the same behavior though.

marner2:
@TysonMN I think if you merge #501 it will give you the option to rebase again, which will get rid of the merge commit.

PR #514, opened 2022-09-13
Comment out some uncalled Samples code.

rfreytag:
For clarity, commented-out some uncalled code in src\Samples\NewWindow.Core\AutoOpen.fs

### DIFF
--- a/src/Samples/NewWindow.Core/AutoOpen.fs
+++ b/src/Samples/NewWindow.Core/AutoOpen.fs
@@ -31,6 +31,7 @@ module InOutModule =
       | InOut.In  msg -> msg |> f
       | InOut.Out msg -> msg |> g
 
-    let map f g = cata (f >> InOut.In) (g >> InOut.Out)
-    let mapIn  f = map f id
-    let mapOut f = map id f
+    // Unused code
+    //let map f g = cata (f >> InOut.In) (g >> InOut.Out)
+    //let mapIn  f = map f id
+    //let mapOut f = map id f
--- a/src/Samples/NewWindow.Core/AutoOpen.fs
+++ b/src/Samples/NewWindow.Core/AutoOpen.fs
@@ -30,8 +30,3 @@ module InOutModule =
     let cata f g = function
       | InOut.In  msg -> msg |> f
       | InOut.Out msg -> msg |> g
-
-    // Unused code
-    //let map f g = cata (f >> InOut.In) (g >> InOut.Out)
-    //let mapIn  f = map f id
-    //let mapOut f = map id f
--- a/src/Samples/NewWindow.Core/NewWindow.Core.fsproj
+++ b/src/Samples/NewWindow.Core/NewWindow.Core.fsproj
@@ -17,13 +17,17 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Serilog" Version="2.11.0" />
+    <PackageReference Include="Serilog" Version="2.12.0" />
     <PackageReference Include="Serilog.Extensions.Logging" Version="3.1.0" />
-    <PackageReference Include="Serilog.Sinks.Console" Version="4.0.1" />
+    <PackageReference Include="Serilog.Sinks.Console" Version="4.1.0" />
   </ItemGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\..\Elmish.WPF\Elmish.WPF.fsproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Update="FSharp.Core" Version="6.0.6" />
   </ItemGroup>
 
 </Project>
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -30,7 +30,7 @@ jobs:
       run: dotnet test --no-build --configuration Release src/Elmish.WPF.Tests/Elmish.WPF.Tests.fsproj
 
     - name: Publish NuGet
-      uses: alirezanet/publish-nuget@v3.0.4
+      uses: alirezanet/publish-nuget@v3.0.4
       with:
         PROJECT_FILE_PATH: "src/Elmish.WPF/Elmish.WPF.fsproj"
         PACKAGE_NAME: "Elmish.WPF"


LyndonGingerich:
Is the codebase better without this code available? If so, shouldn't we remove it rather than commenting it out?

rfreytag:
That is certainly possible - code removal.  But my belief is the developer's intent is clearer with the commented code in place.  My intention is tell a novice "this is acceptable style but not used in this particular example."

LyndonGingerich:
Ah, I understand your intention now. I don't yet understand the motive. Is it important that the reader know immediately that this code is unused?

rfreytag:
As a novice, I find myself struggling with the F#-coding idioms used in the samples.  Generally, I want to comment-out unused code and add comments to help others understand what matters and why in a way that would have helped me.  This is probably going to be the nature of my contributions as I learn.

LyndonGingerich:
Documenting the code is an excellent idea.

By the way, if you find anywhere [the tutorial](https://github.com/elmish/Elmish.WPF/blob/master/TUTORIAL.md) or other documentation is unclear or outdated, feel free to open an issue for it. You'll get the information you need and help contribute to the project at the same time.

Help me understand your perspective. How would the code have been easier to understand if the unused bits had been commented out?

rfreytag:
These are not criticisms so much as "perspectives of a novice".  I _appreciate_ F# as challenging and productive and Elmish.WPF for its utility.  The probably best answer to this learning curve is not code differently but instead a "Guide to F# Idioms for the Novice Seeking to Read This Code;" or something like that.  That said, overloading: get, set, map, cata, MapIn..., MapOut..., without:  universal use of 'qualified access' or probably-overly-wordy-but-descriptive names and no comments has been a challenge.  Here is a novice question.  I come from a OO background in both dynamicly- and staticly-typed languages that require or good practice demands functions have full parameter lists.  F# infers types and parameters if not explicit  This is cool and great for eliminating errors at compile-time but relying on it seems to make the code inscrutable.  Is there a reason mostly not specify the parameter list of a function in production F# code?  As it is, I am tracing long dependencies through the code figuring out what functions are expecting to understand what variables mean and how they got there.  It feels like extensive regions of the code are tightly coupled which hinders readability for this novice.  Hopefully it all gets easier with more F# exposure if this is the norm with solid reasons for these practice.  Sorry, don't want to sound like I am complaining - hope this is somehow useful - but I think I need is a reality check - thank you.

TysonMN:
> Is the codebase better without this code available?

Yes. Please modify this PR to delete this unused code.

TysonMN:
> Is there a reason mostly not specify the parameter list of a function in production F# code?

I am not completely sure I know what you mean.  If I have misunderstood, then please give an example.

My guess is that you are asking:
> Why are [functions curried](https://en.wikipedia.org/wiki/Currying) in F#?

The short answer is that curried functions are more composable, which makes them more useful.

rfreytag:
Thanks TysonMN - Can one have both composability and explicit parameter lists?

TysonMN:
To make sure we are talking about the same things, do you want to give an example of one function with an explicit parameter list and the corresponding function with an implicit parameter list?

Maybe you asking
> Why do programmers in F# use [higher-order functions](https://en.wikipedia.org/wiki/Higher-order_function)?

(Notice that every curried function is a higher-ordered function.)

Higher-ordered functions are an [abstraction that eliminates "irrelevant" function parameters and amplifies the essential ones](https://www.goodreads.com/quotes/8806618-abstraction-is-the-elimination-of-the-irrelevant-and-the-amplification).  It is rare to see higher-ordered functions in OOP languages like C# and Java because of all the [ceremony](https://blog.ploeh.dk/2019/12/16/zone-of-ceremony/).

LyndonGingerich:
@rfreytag I'm not taking it as complaining. Your concerns are valid. Code is best written to be read easily. Documentation comments can often be useful; so can `RequireQualifiedAccess`. PRs adding them will be given due consideration.

> Thanks TysonMN - Can one have both composability and explicit parameter lists?

Are you talking about eta-expansion? For example:
```fsharp
let add x y = x + y
let add2 y = add 2 y
```
instead of
```fsharp
let add x y = x + y
let add2 = add 2
```

> probably-overly-wordy-but-descriptive names

The question is whether the name adds helpful information. For example, we could write `let flip f a b = f b a` as `let flip functionToFlip a b = functionToFlip b a`. Some might argue that such names would be more descriptive. But the identifier `f` usually refers to a function, and it's being passed into a function called `flip`, so we already know that `functionToFlip` is a function that gets flipped. We might as well just call it `f` and take less time reading it--and short F# functions like this, when phrased well, are very fast to read and figure out. But there are other cases in which names should be longer, and if you have suggestions either way (or questions about the reasons for the current wording), feel free to raise them.

> That said, overloading: get, set, map, cata, MapIn..., MapOut...

If it's ever unclear which of any of these we're referring to, the code needs changed.

> Hopefully it all gets easier with more F# exposure if this is the norm with solid reasons for these practice. Sorry, don't want to sound like I am complaining - hope this is somehow useful - but I think what I need is a reality check - thank you.

You're the norm. Everyone is. Ask questions and make suggestions. The length of this discussion isn't intended as criticism. You found dead code. Good job. You're also suggesting adding documentation. I think this project could use more documentation. I look forward to hearing your suggestions.

rfreytag:
@TysonMN - great links; especially the idea of 'ceremony' in languages.  Not sure how I got so far without previously encountering: manifest or nominal typing.   I must try what I am proposing on Sample code to see for myself.

@LyndonGingerich - very much appreciate your thoughtful welcoming advice.  Yes(!) it appears I am finding _eta-expansion and partial application_ to be a readability challenge.  Your example of ...

`let flip f a b = f b a`

... actually seemed clear but its usage on the very next line in `../src/Samples/NewWindow.Core/AutoOpen.fs` ...

`let map get set f a =  a |> get |> f |> flip set a`

  ... seemed opaque to this novice as I mentally juggled how the last parameter for `flip` was supplied.  Any advice on how to _read, perhaps vocally read, that line would be a big help._  I know Perl is not admired amongst language aficionados but the designer and author of the Perl Reference (O'Reilly) is has an linguistics background and it shows in the mnemonics and [literate programming](https://thenewstack.io/why-literate-programming-might-help-you-write-better-code/) in the documentation.  For example, one might verbalize `>>` as _composes-with_.  This verbalization mentally orders the expressions so they are easily understood for me and maybe others.  
 
Very open to strategies _you_ use as you read F# code to grasp what is being done.  

As I wrote previously, the only answer might be a list of _idioms_ used in the Elmish.WPF Sample code.  I may try my hand at that.  At least it will serve as a punch-down list for my progress through this codebase.

LyndonGingerich:
> Very open to strategies you use as you read F# code to grasp what is being done.

Mostly, I try not to write cryptic F#. If I find F# that is hard to read, I look for bits I can simplify to make it easier to read. Even if my changes aren't worth making, the act of fiddling with it increases my enlightenment.

In my experience, `flip` usually makes code harder to read. @TysonMN, what if we rewrote `map` as `let map get set f a = a |> get |> f |> set <| a`?

LyndonGingerich:
> For example, one might verbalize >> as composes-with. This verbalization mentally orders the expressions so they are easily understood for me and maybe others.

Good point. Maybe difficulty in verbalization could be considered a code smell.

rfreytag:
`a |> set <| b` requires one read with the _operator precedence_ list close-to-hand.  Also Don Syme's presentation [F# Code I Love](https://www.youtube.com/watch?v=1AZA1zoP-II) recommends against `<|` 

I get what is going on with `flip` and `map`.  Just using that as an example of novice challenges you have put far behind you.  :-)

src/Samples/NewWindow.Core/NewWindow.Core.fsproj
```diff
@@ -17,13 +17,17 @@
   </ItemGroup>
```
TysonMN, line 17: Please remove the changes in this file

src/Samples/NewWindow.Core/AutoOpen.fs
```diff
@@ -29,8 +29,4 @@ module InOutModule =
 
     let cata f g = function
       | InOut.In  msg -> msg |> f
-      | InOut.Out msg -> msg |> g
-
-    let map f g = cata (f >> InOut.In) (g >> InOut.Out)
-    let mapIn  f = map f id
-    let mapOut f = map id f
\ No newline at end of file
+      | InOut.Out msg -> msg |> g
```
TysonMN, line 32: Sent does this diff include an added line?

TysonMN (changes requested):

TysonMN:
> In my experience, `flip` usually makes code harder to read. @TysonMN, what if we rewrote `map` as `let map get set f a = a |> get |> f |> set <| a`?

Both `flip` and backwards pipe make code harder to read.  They should be avoided in committed code.  (They are extremely helpful when refactoring though.)

I think `flip` is simpler in this code than using a backwards pipe.  That might typically be true, but I am not sure.

The reason I committed this use of `flip` is that this line is very common in code I write.  It is often repeated for each field of each record.  Each of these is a [monomorphic functor](https://blog.ploeh.dk/2020/10/19/monomorphic-functors/).  If F# had [higher-kinded types](https://en.wikipedia.org/wiki/Kind_%28type_theory%29), then this line of code wouldn't need to be duplicated.

LyndonGingerich:
I often write functions such as `map` that need an argument passed in both at the beginning and the end, like this one. I can think of a few approaches:

```fsharp
let map get set f a = a |> get |> f |> flip set a
```
```fsharp
let map get set f a = a |> get |> f |> set <| a
```
```fsharp
let map get set f a = (a |> get |> f |> set) a
```
```fsharp
let map get set f a =
  let map = a |> get |> f |> set
  map a
```
```fsharp
let map get set f a = (get >> f >> set) a a
```

In practice, I usually use `<|` at the very end of expressions. I'm open to arguments for other approaches.

TysonMN:
Oh! I really like

```fsharp
let map get set f a = (get >> f >> set) a a
```

That means it can (arguably) be simplified to

```fsharp
let map get set f = get >> f >> set |> Func.flatten
```

rfreytag:
`let map get set f a = (get >> f >> set) a a`

The use of parentheses is widely thought good practice and likely familiar to those new to F#/Haskell idioms.

This discussion has helped me a great deal.  Thank you!

LyndonGingerich:
Something in me says that some will prefer `let map get set f a = a |> (get >> f >> set) a`. I still prefer `<|` even if Mr. Syme doesn't, though.

rfreytag:
@LyndonGingerich Would this be clearer?  

Use `(a |> set) <| b` 
or
use `a |> (set <| b)`

?

LyndonGingerich:
> Depending on the behavior desired. Using parentheses even when they are redundant is ok, I think, if they clarify.

If they clarify, yes, but we read left-to-right, so I think it's already clear. The one we read first goes in first.

Or `Func.flatten` works as well, but that takes a little accustomization.

I like `<|` because it seems to me the most simple, intuitive, and lightweight solution. But we shouldn't change the current standard without a lot more discussion.

rfreytag:
@LyndonGingerich I messaged my email address to you over LinkedIn.  I am looking for some help on proprietary Elmish.WPF code on which I am stuck.  If you have a moment I'd really appreciate it.

PR #521, opened 2022-10-03
Split REFERENCE.md from TUTORIAL.md

LyndonGingerich:


TysonMN:
The tutorial is referenced from `README.md`.  How about adding a link to this new reference file as well?

LyndonGingerich:
> The tutorial is referenced from README.md. How about adding a link to this new reference file as well?

Good idea. Done.

PR #550, opened 2022-12-20
Do not throw NullReferenceException when ValueOption.toNull is called on ValueNone of reference type

LyndonGingerich:
`unbox null` throws a `NullReferenceException`.

Thanks to @marner2 for figuring out the solution.

### DIFF
--- a/src/Elmish.WPF/InternalUtils.fs
+++ b/src/Elmish.WPF/InternalUtils.fs
@@ -80,8 +80,9 @@ module ValueOption =
   let toNull<'a> = function
     | ValueSome x -> Ok x
     | ValueNone ->
-      if box Unchecked.defaultof<'a> = null then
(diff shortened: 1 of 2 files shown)


marner2 (approved):

src/Elmish.WPF/InternalUtils.fs
```diff
@@ -81,7 +81,7 @@ module ValueOption =
     | ValueSome x -> Ok x
     | ValueNone ->
       if box Unchecked.defaultof<'a> = null then
```
TysonMN, line 83: How about extracting `Unchecked.defaultof<'a>`?

TysonMN (changes requested):

src/Elmish.WPF/InternalUtils.fs
LyndonGingerich: Done.

PR #551, opened 2023-01-27
Updated to Elmish v4

JordanMarr:
I just upgraded Avalonia.FuncUI to Elmish 4, so I thought I would try to update it here as well while it was fresh on my mind.
(My apologies if you have already discussed this somewhere, but I didn't see anything about it.)

### DIFF
--- a/src/Elmish.WPF/Elmish.WPF.fsproj
+++ b/src/Elmish.WPF/Elmish.WPF.fsproj
@@ -54,12 +54,12 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Elmish" Version="[3.1.0, 3.99]" />
+    <PackageReference Include="Elmish" Version="4.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="[6.0.1, 99]" />
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Update="FSharp.Core" Version="[6.0.5, 99]" />
+    <PackageReference Update="FSharp.Core" Version="7.0.0" />
   </ItemGroup>
 
 </Project>
--- a/src/Elmish.WPF/WpfProgram.fs
+++ b/src/Elmish.WPF/WpfProgram.fs
@@ -99,7 +99,7 @@ module WpfProgram =
        *)
       fun msg -> element.Dispatcher.InvokeAsync(fun () -> innerDispatch msg) |> ignore
 
-    let logMsgAndModel (msg: 'msg) (model: 'model) =
+    let logMsgAndModel (msg: 'msg) (model: 'model) _ =
       updateLogger.LogTrace("New message: {Message}\nUpdated state:\n{Model}", msg, model)
 
     let errorHandler (msg: string, ex: exn) =
@@ -110,8 +110,7 @@ module WpfProgram =
     |> if updateLogger.IsEnabled LogLevel.Trace then Program.withTrace logMsgAndModel else id
     |> Program.withErrorHandler errorHandler
     |> Program.withSetState setState
-    |> Program.withSyncDispatch cmdDispatch
-    |> Program.run
+    |> Program.runWithDispatch cmdDispatch ()
 
 
   /// Instantiates Application and sets its MainWindow if it is not already
--- a/src/Samples/FileDialogs.Core/Program.fs
+++ b/src/Samples/FileDialogs.Core/Program.fs
@@ -84,11 +84,16 @@ let bindings () : Binding<Model, Msg> list = [
 let designVm = ViewModel.designInstance (init () |> fst) (bindings ())
 
 
-let timerTick dispatch =
-  let timer = new Timers.Timer(1000.)
-  timer.Elapsed.Add (fun _ -> dispatch (SetTime DateTimeOffset.Now))
-  timer.Start()
+let subscriptions (model: Model) : Sub<Msg> =
+  let timerTickSub dispatch =
+    let timer = new Timers.Timer(1000.)
+    let disp = timer.Elapsed.Subscribe(fun _ -> dispatch (SetTime DateTimeOffset.Now))
+    timer.Start()
+    disp
 
+  [
+    [ nameof timerTickSub ], timerTickSub
+  ]
 
 let main window =
 
@@ -100,7 +105,8 @@ let main window =
       .WriteTo.Console()
       .CreateLogger()
 
+
   WpfProgram.mkProgram init update bindings
-  |> WpfProgram.withSubscription (fun _ -> Cmd.ofSub timerTick)
+  |> WpfProgram.withSubscription subscriptions
   |> WpfProgram.withLogger (new SerilogLoggerFactory(logger))
   |> WpfProgram.startElmishLoop window
--- a/src/Samples/FileDialogsCmdMsg.Core/Program.fs
+++ b/src/Samples/FileDialogsCmdMsg.Core/Program.fs
@@ -108,11 +108,16 @@ open Platform
 let designVm = ViewModel.designInstance (init () |> fst) (bindings ())
 
 
-let timerTick dispatch =
-  let timer = new Timers.Timer(1000.)
-  timer.Elapsed.Add (fun _ -> dispatch (SetTime DateTimeOffset.Now))
-  timer.Start()
+let subscriptions (model: Model) : Sub<Msg> =
+  let timerTickSub dispatch =
+    let timer = new Timers.Timer(1000.)
+    let disp = timer.Elapsed.Subscribe(fun _ -> dispatch (SetTime DateTimeOffset.Now))
+    timer.Start()
+    disp
 
+  [
+    [ nameof timerTickSub ], timerTickSub
+  ]
 
 let main window =
 
@@ -125,6 +130,6 @@ let main window =
       .CreateLogger()
 
   WpfProgram.mkProgramWithCmdMsg init update bindings toCmd
-  |> WpfProgram.withSubscription (fun _ -> Cmd.ofSub timerTick)
+  |> WpfProgram.withSubscription subscriptions
   |> WpfProgram.withLogger (new SerilogLoggerFactory(logger))
   |> WpfProgram.startElmishLoop window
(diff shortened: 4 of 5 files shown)


src/Elmish.WPF/Elmish.WPF.fsproj
```diff
@@ -54,12 +54,12 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Elmish" Version="[3.1.0, 3.99]" />
+    <PackageReference Include="Elmish" Version="4.0.0" />
```
TysonMN, line 57: Please change to

```
[4.0.0, 4.99]
```

TysonMN (changes requested):

TysonMN:
Where is the code for Elmish v4?  Is it the head of `master`?  Can you link to it?

src/Elmish.WPF/Elmish.WPF.fsproj
JordanMarr: Updated.

JordanMarr:
> Where is the code for Elmish v4? Is it the head of master? Can you link to it?

All my changes were included in the `elmish-v4` branch. What are you looking for?

TysonMN:
Sorry for being unclear.

I am asking for a link to the Elmish v4 code in the Elmish GitHub repository.

JordanMarr:
Ah ok. Looks like it has its own branch:
https://github.com/elmish/elmish/tree/v4.x/src

src/Elmish.WPF/Elmish.WPF.fsproj
```diff
@@ -54,12 +54,12 @@
...
 
   <ItemGroup>
-    <PackageReference Include="Elmish" Version="[3.1.0, 3.99]" />
+    <PackageReference Include="Elmish" Version="[4.0.0, 4.99]" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="[6.0.1, 99]" />
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Update="FSharp.Core" Version="[6.0.5, 99]" />
+    <PackageReference Update="FSharp.Core" Version="6.0.7" />
```
TysonMN, line 62: I don't see any reason to change this line.  Elmish v4 is using version `4.7.*`.

https://github.com/elmish/elmish/blob/8b812950f98b7784406b87cb63a1219bb4c5b795/netstandard/Elmish.fsproj#L16

JordanMarr: Interesting. I went by the final calculated dependency listed on NuGet, which is ">= 6.0.7"
https://www.nuget.org/packages/Elmish#dependencies-body-tab

TysonMN: That is strange.

What does the package manager in Visual Studio say?  Can you share a screenshot?

JordanMarr: From VS 2022 package manager:
image

JordanMarr: I see the problem:
The url you sent was from a different branch:
image


Here is the v4 branch:
image

JordanMarr: I notice that it also includes the "Fable.Core", so we must be looking at the wrong thing.

JordanMarr: Oh, that's because it is "Fable.Elmish.fsproj"!

JordanMarr: Some new information here:
https://github.com/elmish/elmish/issues/267#issuecomment-1418199030

JordanMarr:
I have added `<DisableImplicitFSharpCoreReference>True</DisableImplicitFSharpCoreReference>` per [this discussion](https://github.com/elmish/elmish/issues/267) which allows us to use the lower FSharp.Core version.

JordanMarr:
[PR](https://github.com/elmish/elmish/pull/268) has been submitted to the Elmish repository.
The Elmish.WPF version has been reverted back to the previous value.

TysonMN:
Thanks for all of your hard work :)

I want to take a close look but am busy at the moment.

JordanMarr:
@TysonMN, 

I totally understand and have been there many times myself. 
There is no rush, and I appreciate the heads up. 
Thanks for your work maintaining this fantastic library. 👍

marner2:
@JordanMarr Can you resolve the conflicts again? Thanks!

JordanMarr:
Merged!

marner2:
@TysonMN @JordanMarr would it make sense to add a `WpfProgram.withEffect` function for easy migration? Just a drop-in so that if someone doesn't want to use the new subscription feature they could just replace it with this other function?

The other option would be to do what [this comment from the original issue](https://github.com/elmish/elmish/issues/183#issuecomment-1224105679) suggests, which is keep the current behavior of `WpfProgram.withSubscription` and make a new one called `WpfProgram.withLiveSubscription`.

It's just one less thing to break.

marner2:
@JordanMarr [Here is a branch on my fork](https://github.com/JordanMarr/Elmish.WPF/compare/elmish-v4...marner2:Elmish.WPF:elmish-v4) of what I was thinking.

JordanMarr:
@marner2 ,
I tend to lean towards not adding the helper function, but it's not my decision to make, and I don't really have a strong opinion either way. If you do move forward with creating a helper function, I think `withEffect` sounds better than `withLiveSubscription`.

**Pros:**
* Creating `withEffect` helper function would make it slightly easier for users to upgrade in the near term. 

**Cons:**
* Your example hardcodes the same subscription name string so it will break if more than one subscription is used. Also, it doesn't dispose anything, so the user will be allowed to _not_ consider disposal, which the v4 API forces them to do (which I think is probably a good thing).
* It would add more code for this library to maintain and support going forward, potentially causing bloat, and multiple ways of doing things that would require explanation for users over time.
* Elmish is used by many other F# libraries, so most users will become familiar with the new v4 subscription syntax one way or the other. Why not just rip off the bandaid?

marner2:
Just a point with one of the cons you listed:
> * Your example hardcodes the same subscription name string so it will break if more than one subscription is used. Also, it doesn't dispose anything, so the user will be allowed to not consider disposal, which the v4 API forces them to do (which I think is probably a good thing).

You can't actually call `Program.withSubscription` twice. The one axes over the other, so there will be no existing Elmish.WPF programs that this will break by hardcoding the name string (see [elmish Program.fs:55](https://github.com/elmish/elmish/blob/v4.x/src/program.fs#L55)). Though that does bring up the point that we should probably surface `mapSubscription` as well in this migration.

As to the other point, I think maybe if we addressed it as a deprecated helper function rather than a full static function on `WpfProgram` it would better split the difference between the other concerns you brought up and an easy migration path. Either that or maybe we can come up with a more clever mapping strategy between `Effect<'msg>`s and `Subscribe<'msg>`s that would make sense to keep maintained (for example, indicating that they have full lifetime)

Honestly for me personally my eyes start going crosseyed looking at the difference between `Dispatch<'msg>`, `Effect<'msg>`,  `Cmd<'msg>`, `Subscribe<'msg>`, `Sub<'msg>`, etc especially since the main project I work in makes use of `'CmdMsg`'s to separate the impure effects from the pure update loop (not to mention WPF's `ICommand` interface). A part of me wants to rename `Cmd<'msg>` to `Eff<'msg>` so the parallel structure there is more apparent (then we can rename `'CmdMsg` to `'EffMsg`).

marner2:
@JordanMarr I updated the branch to make an obsolete helper function that takes in the prefix externally. See [my v4 branch compared to yours](https://github.com/JordanMarr/Elmish.WPF/compare/elmish-v4...marner2:Elmish.WPF:elmish-v4). Let me know if you think that's ok.

I left `Subscribe.ofEffect` as a helper function just because it seems like it could be useful (and descriptive *enough* to not be too confusing). Though I'm not too attached to it if you think I should just make it a private helper.

I also added `WpfProgram.mapSubscription` just to maintain parity.

marner2:
@JordanMarr I just realized that I misled you about the `withSubscription` behavior. Apparently that changed to overwrite everything in v4.0 and wasn't actually pre-existing behavior.

Either way, the correct way is to support `mapSubscription` and force the user to give it a new prefix. So the helper I created still is valid, they just need to know not to call `withSubscription` when they mean `mapSubscription`.

marner2:
@TysonMN do we need to surface the new v4 `termination` function in `Elmish.WPF`?

JordanMarr:
@marner, I think that all looks great.

marner2:
@JordanMarr can you pull those changes into this branch?

I think we should also discuss (maybe @TysonMN could weigh in) about only surfacing an additive version of subscriptions (no replacing everything that was there) as well as for termination (which is more relevant for subscriptions, as it actually calls all of the dispose methods).

TysonMN:
To terminate the Elmish loop?

TysonMN:
Doesn't seem necessary, but also seems fine to expose?

TysonMN:
Actually, I am surprised there is a function.  Instead, I expect the function call that starts the Elmish loop would return an instance of `IDisposable`.

marner2:
@TysonMN It seems to be a way to stop the elmish loop from a message in the message loop, which I don't think would be possible if we just return an `IDisposable` from `startElmishLoop`.

Is there currently a way to shut down the app from inside the elmish loop? Currently I think we just rely on the main thread to exit the main dispatcher, which would cause subsequent `dispatch` calls to fail (because we would be calling `MainThread.Dispatcher.InvokeAsync(...)` on a shutdown thread). But I don't think there is currently any way to exit from an elmish MVU message. You'd have to invoke a command that has a reference to the window or application that shuts it down cleanly.

TysonMN:
I don't think so.

Fair enough then.

marner2:
@JordanMarr I think it's settled then. I pushed my changes to the main repository (I can't access yours), can you pull from there into your branch and push to your branch so it shows up in this pull request?

Then we can release a new beta version and talk about closing out a proper v4 release.

marner2:
@JordanMarr Can you do a hard reset of your branch on the commit I mentioned? This looks like just a merge commit.

We've been having the practice of keeping the git history mostly clean of complicated merges and stuff like that, and rebasing all branches on the tip before merging.

marner2:
I figured out how to do it I think.

PR #556, opened 2023-02-28
Name elements of the SubModelOpt sample Form2 consistently

### LINKED ISSUE #439: Inconsistent naming pattern for setter messages
When looking at the SubModelOpt.Core/Program.fs sample file, it took a while before I realized that the messages TextInput, TextInput1 and TextInput2 were simply setter messages. I was further confused by Form1.Model having field Text, while Form2.Model has fields Input1 and Input2.

I suggest the names be changed so that fields are named Text, Text1 and Text2, and likewise their bindings, while their corresponding setter messages be named SetText, SetText1 and SetText2. I assume this naming pattern is the convention we have come to expect, and it is much easier to instantly recognize the meaning of the code when we follow it. I haven't checked if other samples have code that does not follow this pattern.

I know this is a ridiculously simple change that I could have submitted rather than create an issue, but the thing is that I'm not used to using Git and GitHub for that kind of thing, so I'd spend much more time rather than less.

LyndonGingerich:
resolves #439

### DIFF
--- a/src/Samples/SubModelOpt.Core/Program.fs
+++ b/src/Samples/SubModelOpt.Core/Program.fs
@@ -31,8 +31,8 @@ module Form1 =
 module Form2 =
 
   type Model =
-    { Input1: string
-      Input2: string }
+    { Text1: string
+      Text2: string }
 
   type Msg =
     | Text1Input of string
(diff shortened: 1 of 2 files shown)


TysonMN:
Thanks @LyndonGingerich! :)

PR #564, opened 2023-05-08
Combine Seq.map and Seq.iter

LyndonGingerich:
Cleanup after #563


TysonMN (changes requested):
The change is fine, but I think the commit message is misleading.  I think something like this would be better:

 Call updateModel more directly in SubModelT.seq

(wrote this from my phone so double check that the names are all correct).

LyndonGingerich:
That commit message doesn't make as much sense to me, but I changed it.

TysonMN:
I thought the previous commit message was referring to the implementations of those functions.  Of course I now see that isn't possible (since we didn't implement those functions).

Another commit message that is not similar to your first one that I think is clear would be

> Inline Seq.map argument into Seq.iter argument

LyndonGingerich:
Better?

PR #567, opened 2023-05-19
Adding fully specified parameter lists to NewWindow.Core and FileDialogsCmdMsg.Core

rfreytag:
I found Sample/NewWindow.Core/App.fs hard to read. I have a version in [my fork](https://github.com/rfreytag/Elmish.WPF/tree/Adding_fully-specified_parameter_lists), branch: Adding_fully-specified_parameter_lists that I find more readable with the following changes:

    Every function has fully-specified parameter lists and return values. Readability is improved by not having to hover over function names to see what the function manipulates.

    module App, module Window1 and module Window2 have been renamed module App_Window1 and module App_Window2 so their usage in App.bindings is not confused with the actual references to Window2 in App.bindings.

    Clarified module App, module Window1 and module Window2: get, set, and map so they specify the type on which they operate for readability. They are now respectively:
        module App, module Window1: get_Window1State1, set_Window1State1, and map_Window1State1 and
        module App, module Window2: get_Window2, set_Window2, map_Window2.

    Replaced all usage of function with explicit match <parameter name> with to support with the explicit parameter lists.

I added fully-specified parameter lists to the remaining Sample/NewWindow.Core F# files while I was at it.  And then for good measure I did the same to FileDialogsCmdMsg.Core.

Doing this helped me understand what was going on. I hope it will help others adopt Elmish.WPF. I will issue a pull request if you like where I'm going.

### DIFF
--- a/src/Samples/NewWindow.Core/App.fs
+++ b/src/Samples/NewWindow.Core/App.fs
@@ -21,34 +21,42 @@ type AppMsg =
   | Window2Close
   | Window2Msg of Window2Msg
 
-
 module App =
-  module Window1 =
-    let get m = m.Window1
-    let set v m = { m with Window1 = v }
-    let map = map get set
-  module Window2 =
-    let get m = m.Window2
-    let set v m = { m with Window2 = v }
-    let map = map get set
-    let mapOutMsg = function
+  module App_Window1 =
+    let get_Window1State (m: App) : WindowState<string> =
+      m.Window1
+    let set_Window1State (v: WindowState<string>) (m: App) : App =
+      { m with Window1 = v }
+    let map_Window1State : ((WindowState<string> -> WindowState<string>) -> App -> App) =
+      map get_Window1State set_Window1State
+  module App_Window2 =
+    let get_Window2 (m: App) : Window2 option =
+      m.Window2
+    let set_Window2 (v: Window2 option) (m: App) : App =
+      { m with Window2 = v }
+    let map_Window2 : ((Window2 option -> Window2 option) -> App -> App) =
+      map get_Window2 set_Window2
+    let mapOutMsg (msg: Window2OutMsg) : AppMsg =
+      match msg with
       | Window2OutMsg.Close -> Window2Close
-    let mapInOutMsg = InOut.cata Window2Msg mapOutMsg
+    let mapInOutMsg : InOut<Window2Msg,Window2OutMsg> -> AppMsg =
+      InOut.cata Window2Msg mapOutMsg
 
   let init =
     { Window1 = WindowState.Closed
       Window2 = None }
 
-  let update = function
-    | Window1Show -> "" |> WindowState.toVisible |> Window1.map
-    | Window1Hide -> "" |> WindowState.toHidden  |> Window1.map
-    | Window1Close -> WindowState.Closed |> Window1.set
-    | Window1SetInput s -> s |> WindowState.set |> Window1.map
-    | Window2Show -> Window2.init |> Some |> Window2.set
-    | Window2Close -> None |> Window2.set
-    | Window2Msg msg -> msg |> Window2.update |> Option.map |> Window2.map
+  let update (msg: AppMsg) : (App -> App) =
+    match msg with
+    | Window1Show -> "" |> WindowState.toVisible |> App_Window1.map_Window1State
+    | Window1Hide -> "" |> WindowState.toHidden  |> App_Window1.map_Window1State
+    | Window1Close -> WindowState.Closed |> App_Window1.set_Window1State
+    | Window1SetInput s -> s |> WindowState.set |> App_Window1.map_Window1State
+    | Window2Show -> Window2.init |> Some |> App_Window2.set_Window2
+    | Window2Close -> None |> App_Window2.set_Window2
+    | Window2Msg msg -> msg |> Window2.update |> Option.map |> App_Window2.map_Window2
 
-  let bindings (createWindow1: unit -> #Window) (createWindow2: unit -> #Window) () = [
+  let bindings (createWindow1: unit -> #Window) (createWindow2: unit -> #Window) unit : Binding<App,AppMsg> list = [
     "Window1Show" |> Binding.cmd Window1Show
     "Window1Hide" |> Binding.cmd Window1Hide
     "Window1Close" |> Binding.cmd Window1Close
@@ -60,13 +68,13 @@ module App =
       Window1.bindings >> Bindings.mapMsg Window1SetInput,
       createWindow1)
     "Window2" |> Binding.subModelWin(
-      Window2.get >> WindowState.ofOption,
+      App_Window2.get_Window2 >> WindowState.ofOption,
       snd,
-      Window2.mapInOutMsg,
+      App_Window2.mapInOutMsg,
       Window2.bindings,
       createWindow2,
       isModal = true)
   ]
 
 let private fail _ = failwith "never called"
-let designVm = ViewModel.designInstance App.init (App.bindings fail fail ())
+let designVm : obj = ViewModel.designInstance App.init (App.bindings fail fail ())
--- a/src/Samples/NewWindow.Core/Program.fs
+++ b/src/Samples/NewWindow.Core/Program.fs
@@ -11,7 +11,7 @@ open Elmish.WPF
 open AppModule
 
 
-let main mainWindow (createWindow1: Func<#Window>) (createWindow2: Func<#Window>) =
+let main (mainWindow: Window) (createWindow1: Func<#Window>) (createWindow2: Func<#Window>) : unit =
   let logger =
     LoggerConfiguration()
       .MinimumLevel.Override("Elmish.WPF.Update", Events.LogEventLevel.Verbose)
@@ -19,14 +19,14 @@ let main mainWindow (createWindow1: Func<#Window>) (createWindow2: Func<#Window>
       .MinimumLevel.Override("Elmish.WPF.Performance", Events.LogEventLevel.Verbose)
       .WriteTo.Console()
       .CreateLogger()
-  let createWindow1 () = createWindow1.Invoke()
-  let createWindow2 () =
+  let createWindow1 unit : #Window = createWindow1.Invoke()
+  let createWindow2 unit : #Window =
     let window = createWindow2.Invoke()
     window.Owner <- mainWindow
     window
 
-  let init () = App.init
-  let bindings = App.bindings createWindow1 createWindow2
+  let init unit : App = App.init
+  let bindings : (unit -> Binding<App,AppMsg> list) = App.bindings createWindow1 createWindow2
   WpfProgram.mkSimple init App.update bindings
   |> WpfProgram.withLogger (new SerilogLoggerFactory(logger))
   |> WpfProgram.startElmishLoop mainWindow
--- a/src/Samples/NewWindow.Core/Window2.fs
+++ b/src/Samples/NewWindow.Core/Window2.fs
@@ -28,37 +28,38 @@ type Window2OutMsg =
 
 module Window2 =
   module Input =
-    let get m = m.Input
-    let set v m = { m with Input = v }
+    let get (m: Window2) : string = m.Input
+    let set (v: string) (m: Window2) : Window2 = { m with Input = v }
   module IsChecked =
-    let get m = m.IsChecked
-    let set v m = { m with IsChecked = v }
+    let get (m: Window2) : bool = m.IsChecked
+    let set (v: bool) (m:Window2) : Window2 = { m with IsChecked = v }
   module ConfirmState =
-    let set v m = { m with ConfirmState = v }
+    let set (v: ConfirmState option) (m: Window2) : Window2 = { m with ConfirmState = v }
 
-  let init =
+  let init : Window2 =
     { Input = ""
       IsChecked = false
       ConfirmState = None }
 
-  let update = function
+  let update (msg: Window2Msg) : (Window2 -> Window2) =
+    match msg with
     | SetInput s -> s |> Input.set
     | SetChecked b -> b |> IsChecked.set
     | Submit -> ConfirmState.Submit |> Some |> ConfirmState.set
     | Cancel -> ConfirmState.Cancel |> Some |> ConfirmState.set
     | Close  -> ConfirmState.Close  |> Some |> ConfirmState.set
 
-  let private confirmStateVisibilityBinding confirmState =
+  let private confirmStateVisibilityBinding (confirmState: ConfirmState) : (string -> Binding<Window2,'a>) =
     fun m -> m.ConfirmState = Some confirmState
     >> Bool.toVisibilityCollapsed
     |> Binding.oneWay
 
-  let private confirmStateToMsg confirmState msg m =
+  let private confirmStateToMsg (confirmState: ConfirmState) (msg: 'a) (m: Window2) : InOut<'a,Window2OutMsg> =
     if m.ConfirmState = Some confirmState
     then InOut.Out Window2OutMsg.Close
     else InOut.In msg
 
-  let bindings () =
+  let bindings unit : Binding<Window2,InOut<Window2Msg,Window2OutMsg>,obj> list =
     let inBindings =
       [ "Input" |> Binding.twoWay (Input.get, SetInput)
         "IsChecked" |> Binding.twoWay (IsChecked.get, SetChecked)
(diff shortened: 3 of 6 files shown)


marner2:
@LyndonGingerich @TysonMN do you think we should we should do something similar with the rest of the examples?

rfreytag:
@marner2 Yes, I do think these would help with the other examples.  I am making this change to the Capabilties.Core now.  Could you merge this request so I can branch from 'master' again?

marner2:
Ok sounds good. Please avoid any samples touched by #551 for now to avoid conflicts (until it gets merged in).

rfreytag:
FileDialogsCmdMsg.Core changes were merged in with the commit; sorry.  But I can stay away from: 

- FileDialogs.Core
- SubModel.Core
- SubModelStatic.Core

I plan on next working on Capabilities.Core and perhaps merging in FileDialogsCmdMsg.Core.

src/Samples/NewWindow.Core/App.fs
```diff
@@ -21,34 +21,42 @@ type AppMsg =
...
+    let set_Window1State (v: WindowState<string>) (m: App) : App =
+      { m with Window1 = v }
+    let map_Window1State : ((WindowState<string> -> WindowState<string>) -> App -> App) =
+      map get_Window1State set_Window1State
+  module App_Window2 =
+    let get_Window2 (m: App) : Window2 option =
+      m.Window2
+    let set_Window2 (v: Window2 option) (m: App) : App =
+      { m with Window2 = v }
+    let map_Window2 : ((Window2 option -> Window2 option) -> App -> App) =
```
TysonMN, line 37: This function names are worse.  Please revert these.

src/Samples/NewWindow.Core/Program.fs
```diff
@@ -11,22 +11,22 @@ open Elmish.WPF
...
   let logger =
     LoggerConfiguration()
       .MinimumLevel.Override("Elmish.WPF.Update", Events.LogEventLevel.Verbose)
       .MinimumLevel.Override("Elmish.WPF.Bindings", Events.LogEventLevel.Verbose)
       .MinimumLevel.Override("Elmish.WPF.Performance", Events.LogEventLevel.Verbose)
       .WriteTo.Console()
       .CreateLogger()
-  let createWindow1 () = createWindow1.Invoke()
-  let createWindow2 () =
+  let createWindow1 unit : #Window = createWindow1.Invoke()
```
TysonMN, line 22: I don't think this is a refactor.  (I can't check because I don't have my personal computer with me now.)

Before, the argument had to be `()`, which is the only value of type `Unit`.  Instead, by giving this parameter a name, you are allowing F# to infer the type.  But this parameter is unused, so I think F# infers its type is `obj`.

Same problem in other places.

src/Samples/NewWindow.Core/App.fs
```diff
@@ -21,34 +21,42 @@ type AppMsg =
...
-  module Window1 =
-    let get m = m.Window1
-    let set v m = { m with Window1 = v }
-    let map = map get set
-  module Window2 =
-    let get m = m.Window2
-    let set v m = { m with Window2 = v }
-    let map = map get set
-    let mapOutMsg = function
+  module App_Window1 =
```
TysonMN, line 25: I don't see how this module rename helps.

What do you find confusing about the original naming?

TysonMN: Oh, you said in your OP:

> ...so their usage in `App.bindings` is not confused with the actual references to `Window2` in `App.bindings`.

src/Samples/NewWindow.Core/Program.fs
rfreytag: I had thought I'd read in documentation that `unit` and `()` were synonymous.  I could very well be mistaken.  

See https://web.archive.org/web/20221225214301/https://fsharpforfunandprofit.com/posts/how-types-work-with-functions/
... a little less than 1/2 down Scott W. says...
"
What is this “unit”?

Well, even if a function returns no output, it still needs a range. There are no “void” functions in mathematics-land. Every function must have some output, because a function is a mapping, and a mapping has to have something to map to!

So in F#, functions like this return a special range called “unit”. This range has exactly one value in it, called “()”. You can think of unit and () as somewhat like “void” (the type) and “null” (the value) in C#. But unlike void/null, unit is a real type and () is a real value.
"

As I said I'm stuck with what I read but you know F# better than I.  

I do know that when I read `()` I found it less clearly distinct from all the other uses of paretheses than the reserved word `unit` so where possible I used `unit`.

src/Samples/NewWindow.Core/App.fs
rfreytag: My problem was encountering `Window1` and `Window2` as actual window labels and as modules.  That overload required keeping context in mind but with so much else going on in NewWindow.Core my "mental buffer" filled up.  It helped me a lot to make sure each label meant only one thing without relying on context.  :-)

Your code looks elegant and compact.  I just struggled when I couldn't be sure one-name-meant-one-object/function.

src/Samples/NewWindow.Core/App.fs
```diff
@@ -60,13 +68,13 @@ module App =
       Window1.bindings >> Bindings.mapMsg Window1SetInput,
       createWindow1)
     "Window2" |> Binding.subModelWin(
-      Window2.get >> WindowState.ofOption,
+      App_Window2.get_Window2 >> WindowState.ofOption,
```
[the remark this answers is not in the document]
rfreytag: I agree.  Is there a way to refer to the `module App_Window2` as `App.Window2`?  I could not figure out a way when I tried it yesterday.

src/Samples/NewWindow.Core/Window2.fs
```diff
@@ -28,37 +28,38 @@ type Window2OutMsg =
...
     |> Binding.oneWay
 
-  let private confirmStateToMsg confirmState msg m =
+  let private confirmStateToMsg (confirmState: ConfirmState) (msg: 'a) (m: Window2) : InOut<'a,Window2OutMsg> =
     if m.ConfirmState = Some confirmState
     then InOut.Out Window2OutMsg.Close
     else InOut.In msg
 
-  let bindings () =
+  let bindings unit : Binding<Window2,InOut<Window2Msg,Window2OutMsg>,obj> list =
```
LyndonGingerich, line 62: I strongly object to naming `unit` parameters. They are meant to appear similar to parameterless function calls in other languages; if we name them, we lose the resemblance.

src/Samples/NewWindow.Core/App.fs
LyndonGingerich: I don't think renaming the functions is a good idea. If the names of get/set/map functions aren't clear enough, we should replace them with lambdas (which I think is what many other F# users do). Adding parent module names to function names just repeats information; it doesn't add any.

We could add `RequireQualifiedAccess` to the get/set/map modules if that would help.

src/Samples/NewWindow.Core/Program.fs
LyndonGingerich: Tested on my machine.

```fsharp
let foo = unit // The value or constructor 'unit' is not defined.
```

`unit` is not a keyword in F#. You just named an argument of unspecified type `unit`.

Repeating because it's relevant here: `unit` looks the way it does so that parameterless function calls in F# can resemble parameterless function calls in other languages. Parameterless function `python_func` in Python will be called as `python_func()`, and parameterless function `fSharpFunc` in F# will be called as `fSharpFunc ()`. The creators of F# intended the language to be used this way and everyone (of whom I am aware) uses it this way.

src/Samples/NewWindow.Core/App.fs
TysonMN: The ability to use the same name in the same scope for different things is an advantage of F#.  [Naming is (impossibly) hard](http://tysonwilliams.coding.blog/2020-07-15_naming_is_impossibly_hard), so being able to avoid some of it by reusing existing names can improve things.

That said, I am symothtic to naming these modules differently.  Instead of repeating the parent module name though, I think `Window1Module` (using [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation)) would be better.

TysonMN:
@rfreytag, since this PR was already merged, can you create another PR that address my concerns?

src/Samples/NewWindow.Core/App.fs
marner2: I think adding `RequireQualifiedAccess` is a good compromise here.

src/Samples/NewWindow.Core/Program.fs
LyndonGingerich: Looks as though it's generic (`'a`), so you could pass, say, an integer in if you wanted, which would mean someone was doing something wrong. Further, it's detected as unused now, which annotating its type does not resolve (and `unit: unit` looks ugly anyway).

LyndonGingerich:
I don't think this change was thought through with a lot of nuance. I see lines such as `let init : string = ""`, which doesn't need a type annotation.

Unless I misunderstand, the samples attempt to communicate simply and clearly with those who are used to idiomatic F#. We do not wish to require readers to learn a new set of idioms specific to Elmish.WPF; many probably need only one or two samples to browse. Since the reader is already trying to learn a new framework, he should not have to learn a new way of writing F# at the same time. For this reason, I propose that we do not introduce new idioms that F# developers cannot immediately understand and effortlessly remember.

Changes this PR makes:
- Separating sections of the code with one newline rather than two. Newline style varies by codebase; so long as we use newlines consistently and separate sections of the code, I don't think we should worry about their numbers. This change should not have been made.
- Adding type annotations. Type annotations are situationally useful, but they also add more text to read. We should not annotate the types of identifiers which themselves specify their types (e.g. `cmdMsg: CmdMsg`). This changed should have been considered individually. I don't know when type annotations help others, but I probably would not have added most of these myself. We have better ways of increasing clarity.
- Eta-expansion of `function` to `match`. I have no strong opinions on this change, but I will suggest that type unclarity on a `function` might be better resolved by qualifying the discriminated union cases (and probably adding `RequireQualifiedAccess` to the DU).

On the other hand, we do use idioms in this sample that most F# developers probably wouldn't recognize:
- `init` for initial values. But this is an Elmish idiom, so I think we should keep it.
- `Model` for the top-level record type. Developers will probably understand this.
- `msg` to mean "message". This is a simple abbreviation and probably clear.
- `m` to mean "model". I think this abbreviation is likely unclear. I move to expand it to `model`. Where the type in question is not called `model`, we should rename it to reflect the type in question.
- `flip`. `flip` decreases readability by increasing semantic code complexity. Let's replace it with lambdas.
- `AutoOpen.map`. This function is difficult to read and comprehend, but replacing it might increase the code length significantly. Could we at least rename it to the more familiar (and less overloaded) `lens`?
- `InOut`. This DU is simple and I don't see how we could replace it with anything simpler anyway. We could use `Choice`, but `InOut` has more meaning. (If we decide that the effort of learning the in-message/out-message concept outweighs the organization, though, then we should replace `InOut` with `Choice`.)
- `cata`. I know what `cata` means, and Mark Seeman knows what `cata` means, but I don't think most developers would recognize it. Dictionary.com [defines](https://www.dictionary.com/browse/cata-) it as "'down', 'against', 'back'", which would not make its meaning clear even to one familiar with the prefix. Let's inline it. `match` is quickly and easily readable.
- `get`/`set`/`map`. I see most people using lambdas. Let's replace `get` and `set` with lambdas and thus reduce mental overhead. We can keep `map` where it's useful, but let's rename it to `lens`.
- `mapOutMsg`/`mapInOutMsg`: These don't need to be so far away from their call sites. They are trivially short and used only once each. Let's inline them.

LyndonGingerich:
> LyndonGingerich TysonMN do you think we should we should do something similar with the rest of the examples?

@marner2 Above post intends to answer your question.

src/Samples/NewWindow.Core/App.fs
LyndonGingerich: Isn't the use of a module of the same name with a record (paralleling the functionality of a class) commonly used in F#, though? I would (humbly) vote against adding "Module" to the ends of module names.

marner2:
Yeah I think those suggestions in the last half of your comment make sense @LyndonGingerich, and we definitely should be more considered and deliberate with future sweeping updates of the samples.

I was ok with the update as-is because it's always valuable to get empirical feedback when someone unfamiliar with the project details their experience (and it was just a single file that was updated).

As far as explicitly specified types taking longer to read, while that may be true of production code, samples have a slightly different target. Which is again why I'm ok with liberal eta-expansion (especially of `function` to `match`) showing up in the samples. It just helps guide people who unfamiliar with the idioms. The other thing about these sample projects is that they are intentionally short, and generally in very small implementations you should err more on the side of verbosity. When you have a very large codebase where things get repeated a bunch, that's where the shortcuts start to kick in and pay off.

src/Samples/NewWindow.Core/App.fs
TysonMN: Yes, requiring qualified access is good. Qualified access was my intention.

src/Samples/NewWindow.Core/App.fs
TysonMN: I also prefer to reuse the same name, but I am open to using a different name.

TysonMN:
Adding more type annotations is typically fine. There were some in this PR that don't improve readability, and those can be removed.

LyndonGingerich:
Here's my first effort at a low-complexity version of the sample program. I'm not making a PR because I'm intending to diff on the original, not the current version with this PR merged.

https://github.com/LyndonGingerich/Elmish.WPF/compare/070d311428042676f7f2b92fdd5f40c504c6e4d1...LyndonGingerich:Elmish.WPF:lyndon_suggested_simplifications_05-19-2023

rfreytag:
@TysonMN I will remove the changes to which you objected and reissue the PR.  In my defense, I did offer my branch for review _before_ issuing a pull request.

TysonMN:
Yes, no worries.  You didn't do anything wrong.  It is easy to change code (back).

src/Samples/NewWindow.Core/App.fs
rfreytag: I reverted to the original `get`, `set`, and `map`.

src/Samples/NewWindow.Core/App.fs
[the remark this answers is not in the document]
rfreytag: @LyndonGingerich having multiple meanings for `Window1` and `Window2` diminished `App.fs` readability.  To allow `App.Window1` and `App.Window2` usage I brought `init`, `binding` and `update` to the top level in `App.fs`.  This lead to `Program.fs` requiring explicit references to `AppModule.*`.

src/Samples/NewWindow.Core/Program.fs
rfreytag: I _think_ I removed the place where `unit : unit` appeared.

src/Samples/NewWindow.Core/App.fs
rfreytag: @TysonMN   I bring ye tidings from the computer stone-age where single-letter variables where normal.  I now overcompensate using verbose labels.  :-)

src/Samples/NewWindow.Core/Window2.fs
rfreytag: @LyndonGingerich I get your objection but offer that an exception might be made for consistent parameterization of the `Samples`.

rfreytag:
As you have time would you review my fork, branch `Adding_fully-specified_parameter_lists` latest commit (6de8aae7)?

The commit note reads...
"
- Reverted `get`, `set`, and `map` per @TysonMN preference.  
- Reverted `Window1.init`, `Window2.bindings` per @LyndonGingerich preference.  
- Moved `init`, `update`, and `bindings` out of `App` to allow `App.Window1` and `App.Window2` for _readability_ in both `bindings` and `update`.  
- `Program.fs` now references `AppModule.init`, `AppModule.bindings` and `AppModule.update`.  
- Removed/replaced two usages of `unit` per @LyndonGingerich but fully-specified parameters seemed to require the remaining `unit` references.  
- Reverted all line-spacing edits (i.e. in `NewWindow.Core` and `FileDialogsCmdMsg.Core`).
"

If this doesn't suit I accept it might be best if I stay in my fork.

rfreytag:
@LyndonGingerich I have been rushing to revise/undo my prior PR.  Will read your [thoughts](https://github.com/LyndonGingerich/Elmish.WPF/compare/070d311428042676f7f2b92fdd5f40c504c6e4d1...LyndonGingerich:Elmish.WPF:lyndon_suggested_simplifications_05-19-2023) with interest as time allows.

TysonMN:
Please first create a PR that only addresses the concerns I raised on this PR.  After that is merged, we can look at your other improvement suggestions.

rfreytag:
@TysonMN this [branch in my fork](https://github.com/rfreytag/Elmish.WPF/tree/Master_with_RCHF_contributions_removed) has all @rfreytag contributions removed and all committed changes to today.

PR #569, opened 2023-05-23
All @rfreytag contributions removed while retaining all merged commits by others up to 20230523

rfreytag:
All @rfreytag contributions removed while retaining all merged commits by others up to 20230523

### DIFF
--- a/src/Samples/FileDialogsCmdMsg.Core/Program.fs
+++ b/src/Samples/FileDialogsCmdMsg.Core/Program.fs
@@ -6,18 +6,22 @@ open Serilog.Extensions.Logging
 open Elmish
 open Elmish.WPF
 
+
 module Core =
+
 
   type Model =
     { CurrentTime: DateTimeOffset
       Text: string
       StatusMsg: string }
 
+
   type CmdMsg =
     | Save of string
     | Load
 
-  let init (): Model * CmdMsg list =
+
+  let init () =
     { CurrentTime = DateTimeOffset.Now
       Text = ""
       StatusMsg = "" },
@@ -35,7 +39,8 @@ module Core =
     | SaveFailed of exn
     | LoadFailed of exn
 
-  let update (msg: Msg) (m: Model): Model * CmdMsg list =
+
+  let update msg m =
     match msg with
     | SetTime t -> { m with CurrentTime = t }, []
     | SetText s -> { m with Text = s}, []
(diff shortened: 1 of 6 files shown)


TysonMN:
I see two commits in `master` by you.  Is the intention of this PR to revert both of those commits?

rfreytag:
@TysonMN As I understand it you wanted all the changes removed.  Right?

TysonMN:
Ok.  I created a PR #570 instead of this one.  The proper way to undo changes is to revert a commit in git.  Since this PR of your didn't do that, I was confused about that you intention was with this PR.  My PR reverted the merge commit of your previous PR.  There were a couple small conflicts and I manually resolved.

One reason some objectionable changes made it past PR review in that your PR was large and made many different changes.  I recommend readings [10 tips for better Pull Requests by Mark Seemann](https://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests/), especially the first two items.  As a first-order approximation, each PR review gets the same amount of attention.  Thus, if a PR is small, then each changed line gets a lot of attention, but if the PR is large, then each changed line gets very little attention, which causes things to get missed.

Once you have split up a large PR into smaller ones, the next thing to consider is the order in which you will submit them.  The correct order is from least contentious to most contentious.  That can be easier said than done because it is not always clear what the reviewer will consider contentious.  As you gain more programming experience, you will get a general sense of what is typically considered contentious.  As you repeatedly work with specific people, you will learn what they consider to be contentious.

From memory, I can think of four different changes your previous PR made:
1. Adding type annotations
2. Removing blank lines
3. Renaming modules
4. Renaming functions

You should make separate PRs for each of these changes.

I tried to write to order them from what I think is generally considered least to most contentious.  However, I don't I will approve any of your changes to remove blank lines.  An F# convention is that the number of blank indicates how related the functions or modules are ([source](https://learn.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#:~:text=Extra%20blank%20lines%20may%20be%20used%20(sparingly)%20to%20separate%20groups%20of%20related%20functions.%20Blank%20lines%20may%20be%20omitted%20between%20a%20bunch%20of%20related%20one%2Dliners%20(for%20example%2C%20a%20set%20of%20dummy%20implementations).%20Use%20blank%20lines%20in%20functions%2C%20sparingly%2C%20to%20indicate%20logical%20sections.)):

> Extra blank lines may be used (sparingly) to separate groups of related functions. Blank lines may be omitted between a bunch of related one-liners (for example, a set of dummy implementations). Use blank lines in functions, sparingly, to indicate logical sections.

PR #607, opened 2024-06-06
[Samples] from dynamic bindings to static VM ("SubModelOpt" sample)

YkTru:
- I propose to convert all actual samples using dynamic bindings to static bindings, or make 2 folders and put appropriate samples in each: "Dynamic bindings", "Static bindings"

- I've run this new sample and everything works as expected, but I don't have enough knowledge to know if I've overlooked anything (e.g., not using the right helpers for better performance, missing relevant “>> Binding.x” (e.g., boxT, addLazy, etc.).

- I've converted 2-space indentation to 4-space idiomatic indentation (in fact, why do all samples use 2-space indentation? it's not practical to refactor everything to 4-space if you need to explore/reuse/etc. the sample).

- I suggest using (not idiomatic I know) “_Binding” for the binding helpers and “_ViewModel” for the static VM to improve readability and intellisense experience (i.e. writing “_V” filters automatically to get all static VMs; I do the same in WPF for "_Behaviors", "_Converter", "_ResourceDict" etc. which helps a lot IMO).

- I suggest skipping 2 lines between the module and the associated ViewModel, and 3 between the previous ViewModel and a new module; I suppose this too is not idiomatic, but IME it helps a lot with readability/navigation.

- I can do this kind of conversion from dyamic to static VM for all samples if you think it's relevant, it helps me learn at the same time. (I also intend to provide samples for what I think a “basic abstract project structure” that can scale well might look like (mostly based on previous discussions) if you think it's relevant too (?); this could at least be a starting point for possible samples of this kind, since I'm not an expert. Would another folder in the sample repo named “Project Structures” be relevant then?)

---

- should: 
```fs
let form2Visible_Binding =
    Binding.OneWayT.id
    >> Binding.mapModel (fun (m: App.Model) ->
        match m.Dialog with
        | Some (App.Dialog.Form2 _) -> true
        | _ -> false)

```
be replaced by:
```
    let form2Visible_Binding =
        Binding.oneWay (fun (m: App.Model) ->
            match m.Dialog with
            | Some (App.Dialog.Form2 _) -> true
            | _ -> false)

``` 
and same for all other binding helpers?

---

- Problem: `DataContext="{Binding Form1}"` (and Form2) is the only binding unable to resolve at compile time (though at runtime it works fine):
```fs
<local:Form1
    d:DataContext="{Binding DataContext.Form1, RelativeSource={RelativeSource AncestorType=StackPanel}}"
    DataContext="{Binding Form1}"
    Visibility="{Binding DataContext.Form1Visible, RelativeSource={RelativeSource AncestorType=StackPanel}, Converter={StaticResource VisibilityConverter}}" />
```
codelens says: `(Form1_ViewModel).Path=`, is there a way to access `(App_ViewModel).Path=` here?

### DIFF
--- a/src/Samples/SubModelOpt.Core/Program.fs
+++ b/src/Samples/SubModelOpt.Core/Program.fs
@@ -1,6 +1,5 @@
 ﻿namespace Program
 
-
 open Serilog
 open Serilog.Extensions.Logging
 open Elmish.WPF
--- a/src/Samples/SubModelOpt.Core/Program.fs
+++ b/src/Samples/SubModelOpt.Core/Program.fs
@@ -1,4 +1,5 @@
-﻿module Elmish.WPF.Samples.SubModelOpt.Program
+﻿namespace Program
+
 
 open Serilog
 open Serilog.Extensions.Logging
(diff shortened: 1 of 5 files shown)


YkTru:
(is there any interest?)

YkTru:
@marner2 ?

YkTru:
@TysonMN ? (sorry to tag you but I don't know who to ask else)

marner2:
I think you're on to something here, but we don't want to replace the old versions yet. This all looks like a correct transform though at my first go through. I'll have to download the source to see what's going on with that binding issue. My guess is that it's something to do with that `d:DataContext` being different from the `DataContext`.

TysonMN:
Sounds like you are doing too much in one PR.  If you want to change from 2 to 4 spaces for intention, then do that in a PR of its own.

If you want people to review your work, then make your work easy to review.

YkTru:
@marner2 Great, when you say :

> I think you're on to something here

1-Which part exactly do you think should be kept/revised/removed?

--

> My guess is that it's something to do with that d:DataContext being different from the DataContext.

2-What do you mean "being different" since the code is: 
```
d:DataContext="{Binding DataContext.Form1, RelativeSource={RelativeSource AncestorType=StackPanel}}"
    DataContext="{Binding Form1}"
```
?

Thank you

YkTru:
> Sounds like you are doing too much in one PR. If you want to change from 2 to 4 spaces for intention, then do that in a PR of its own.
> 
> If you want people to review your work, then make your work easy to review.

@TysonMN You're absolutely right, changing the "spaces" add too many changes, would you advise me to make another PR with only the modified code, but keeping the 2-spaces indentation of the original sample? (sorry, I'm new to PR-ing)

TysonMN:
Yes

